PR #4, opened 2025-11-02
[compliance tests] initial server scenarios

ihrpr:
This is a very simplified conformance tests suit with:
- requirements for servers to implement (naming, what features should be doing)
- typescript everything server implementation (using register<> API, not we are moving away from tool, prompt resourse to register<> methods)
- tests (some are very simple)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -59,7 +59,6 @@
             "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
             "dev": true,
             "license": "MIT",
-            "peer": true,
             "dependencies": {
                 "@babel/code-frame": "^7.27.1",
                 "@babel/generator": "^7.28.5",
@@ -1790,7 +1789,7 @@
         },
         "node_modules/@types/body-parser": {
             "version": "1.19.6",
-            "resolved": "https://artifactory.infra.ant.dev:443/artifactory/api/npm/npm-all/@types/body-parser/-/body-parser-1.19.6.tgz",
+            "resolved": "https://registry.npmjs.org/@types/body-parser/-/body-parser-1.19.6.tgz",
             "integrity": "sha512-HLFeCYgz89uk22N5Qg3dvGvsv46B8GLvKKo1zKG4NybA8U2DiEO3w9lqGg29t/tfLRJpJ6iQxnVw4OnB7MoM9g==",
             "dev": true,
             "license": "MIT",
@@ -1801,7 +1800,7 @@
         },
         "node_modules/@types/connect": {
             "version": "3.4.38",
-            "resolved": "https://artifactory.infra.ant.dev:443/artifactory/api/npm/npm-all/@types/connect/-/connect-3.4.38.tgz",
+            "resolved": "https://registry.npmjs.org/@types/connect/-/connect-3.4.38.tgz",
             "integrity": "sha512-K6uROf1LD88uDQqJCktA4yzL1YYAK6NgfsI0v/mTgyPKWsX1CnJ0XPSDhViejru1GcRkLWb8RlzFYJRqGUbaug==",
             "dev": true,
             "license": "MIT",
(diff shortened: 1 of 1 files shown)


src/scenarios/server/client-helper.ts
```diff
@@ -0,0 +1,89 @@
...
+        }
+    );
+
+    const transport = new StreamableHTTPClientTransport(new URL(serverUrl));
+
+    await client.connect(transport);
+
+    return {
+        client,
+        close: async () => {
```
pcarleton, line 38: why not call close directly?

src/scenarios/index.ts
```diff
@@ -3,13 +3,92 @@ import { InitializeScenario } from './client/initialize.js';
...
+
 export const scenarios = new Map<string, Scenario>([
     ['initialize', new InitializeScenario()],
     ['tools-call', new ToolsCallScenario()]
 ]);
 
 export const clientScenarios = new Map<string, ClientScenario>([
-    ['initialize', new ServerInitializeClientScenario()]
+    // Legacy
+    ['initialize', new ServerInitializeClientScenario()],
```
pcarleton: fine to delete this 😄

src/scenarios/server/lifecycle.ts
```diff
@@ -0,0 +1,102 @@
...
+    name = 'server-initialize';
+    description = 'Test basic server initialization handshake';
+
+    async run(serverUrl: string): Promise<ConformanceCheck[]> {
+        const checks: ConformanceCheck[] = [];
+
+        try {
+            const connection = await connectToServer(serverUrl);
+
+            // The connection process already does initialization
```
pcarleton, line 18: this is an example of where if we use the SDK, we lose some visibility. this one seems fine as long as the error the SDK throws is easy to interpret in the check output, but just flagging the pattern where we may want to break out of the sdk

src/scenarios/server/lifecycle.ts
```diff
@@ -0,0 +1,102 @@
...
+                    url: 'https://modelcontextprotocol.io/specification/2025-06-18/basic/lifecycle#initialization'
+                }]
+            });
+        }
+
+        return checks;
+    }
+}
+
+export class ServerInitializeCapabilitiesScenario implements ClientScenario {
```
pcarleton: I think delete this one. Seems better to test the actual capability being used, e.g. checking this in Elicitation, Sampling and Roots tests

src/scenarios/server/prompts.ts
```diff
@@ -0,0 +1,333 @@
...
+                    url: 'https://modelcontextprotocol.io/specification/2025-06-18/server/prompts#listing-prompts'
+                }]
+            });
+        }
+
+        return checks;
+    }
+}
+
+export class PromptsGetSimpleScenario implements ClientScenario {
```
pcarleton, line 76: I'm still getting a feel for the cost of an additional scenario, but I'd be inclined to do all these in 1 vs. separate scenarios for each.

SERVER_REQUIREMENTS.md
```diff
@@ -0,0 +1,874 @@
+# MCP Server Conformance Requirements
```
pcarleton, line 1: I was imagining this being a "suite" that is called something like "Basic", and it's a list of the tests to run, where each test describes the functionality it needs. 

I like that this file describes the current SDK goal in a "1-stop" way though, so let's keep this, and we can explore "suites" or profiles later.  Like for SDK's, we want this full set of features, but for servers that don't want some, we can have suites that don't fail on missing optional features.

src/scenarios/server/resources.ts
```diff
@@ -0,0 +1,365 @@
+/**
+ * Resources test scenarios for MCP servers
+ */
+
+import { ClientScenario, ConformanceCheck } from '../../types.js';
+import { connectToServer } from './client-helper.js';
+
+export class ResourcesListScenario implements ClientScenario {
```
pcarleton, line 8: similar here, my current take is these would be better as 1 scenario w/ multiple checks

src/scenarios/server/tools.ts
```diff
@@ -0,0 +1,773 @@
...
+    ListToolsResultSchema,
+    CallToolResultSchema,
+    LoggingMessageNotificationSchema,
+    ProgressNotificationSchema,
+    CreateMessageRequestSchema,
+    ElicitRequestSchema,
+    Progress
+} from '@modelcontextprotocol/sdk/types.js';
+
+export class ToolsListScenario implements ClientScenario {
```
pcarleton, line 9: these ones seem good as separate, but I might group it like:
* list + call simple
* call content types (image, audio, mixed, resource)
* (rest as individual)

the rough heuristic I'm building up is "would it make sense for someone to implement a server that just does this scenario?" (open to iterating on that heuristic, I think we'll want it to be easy for folks to reason through when writing scenarios and using the scenarios).  Right now, we're proposing folks have an everything-server that does all the things, but I don't want it to be a requirement that it's all 1 server since some scenarios will be mutually exclusive (i.e. an SDK should be able to support writing a server that _doesn't_ support certain stateful features if that unlocks other transport things).

Some of this might also be fine to "start out everything super small, and then group things up later based on what's annoying or not as we use it more".

src/server-runner.ts
```diff
@@ -56,40 +57,97 @@ async function main(): Promise<void> {
...
+            const { listClientScenarios } = await import('./scenarios/index.js');
+            const scenarios = listClientScenarios();
+
+            console.log(`Running ${scenarios.length} scenarios against ${serverUrl}\n`);
+
+            for (const scenarioName of scenarios) {
+                console.log(`\n=== Running scenario: ${scenarioName} ===`);
+                try {
+                    const result = await runServerConformanceTest(serverUrl, scenarioName);
+                    allResults.push({ scenario: scenarioName, checks: result.checks });
```
pcarleton, line 91: hypothesis here is we'll want results logged as we go, and summary at the end for faster feedback loops

test-tool-call.ts
```diff
@@ -0,0 +1,39 @@
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
```
pcarleton: is this meant to be in here?

PR #10, opened 2025-11-06
docs: update server requirements

paoloricciuti:
This PR fixes some mistake in the expected output of `SERVER_REQUIREMENTS` while also updating the code blocks to be `json` rather than `typescript`.

## Motivation and Context
1. Implementing an `everything-server` while following the guide was resulting in a few errors because, for example, the tests for elicitation don't actually response with a `{ "respose": "" }` object but with a `{ username: "", email: "" }` object. Also the test for `test_error_handling` correctly checks for `isError` and `content` in the response but the guide was still expecting an error to be thrown.
2. While typescript looks a bit nicer (without a bunch of quotes) the downside is that since the typescript in the blocks it's just an object the formatter automatically adds a `;` at the end of the `content` array meaning copying them to paste them in the server will always require removing that semicolon to have proper syntax. Having them as JSON will allow to just copy paste them (and the formatter will remove the unnecessary quotes).

## How Has This Been Tested?
I have built an `everything-server` and verified the changes are correct.

## Breaking Changes
Nope.

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] New and existing tests pass locally

### DIFF
--- a/SERVER_REQUIREMENTS.md
+++ b/SERVER_REQUIREMENTS.md
@@ -31,7 +31,7 @@ This document specifies requirements based on the MCP specification. All feature
 
 ### Server Identity
 
-Your server MUST provide:
+Your server MUST provide (can substitute your SDK name as the server name):
 
 ```json
 {


SERVER_REQUIREMENTS.md
```diff
@@ -33,32 +33,32 @@ This document specifies requirements based on the MCP specification. All feature
 
 Your server MUST provide:
```
pcarleton:
```suggestion
Your server MUST provide (can substitute your SDK name as the server name):
```

paoloricciuti:
> I think it's likely in the future we'll want to find a way to generate this from the test code so we're not stuck manually updating it for every test change.

That's an interesting idea, not straightforward but definitely interesting, I'll see if I can come up with something

PR #14, opened 2025-11-10
Add conformance tests for SEP-1034 elicitation defaults

felixweinberger:
Adds conformance tests for [SEP-1034](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1034), which extends MCP elicitation support to include default values for all JSON Schema primitive types.

## Motivation and Context
[SEP-1034](https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1034) extends elicitation default value support beyond booleans to all primitive types (string, integer, number, enum). This PR adds conformance tests to validate this functionality.

## How Has This Been Tested?
All 5 conformance checks pass against the TypeScript reference server implementation.

CleanShot 2025-11-10 at 15 54 27

## Breaking Changes
None.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/examples/servers/typescript/everything-server.ts
+++ b/examples/servers/typescript/everything-server.ts
@@ -396,6 +396,82 @@ function createMcpServer() {
     }
   );
 
+  // SEP-1034: Elicitation with default values for all primitive types
+  mcpServer.registerTool(
+    'test_elicitation_sep1034_defaults',
+    {
+      description: 'Tests elicitation with default values per SEP-1034',
+      inputSchema: {}
+    },
+    async () => {
+      try {
+        // Request user input with default values for all primitive types
+        const result = await mcpServer.server.request(
+          {
+            method: 'elicitation/create',
+            params: {
+              message: 'Please review and update the form fields with defaults',
+              requestedSchema: {
+                type: 'object',
+                properties: {
+                  name: {
+                    type: 'string',
+                    description: 'User name',
+                    default: 'John Doe'
+                  },
+                  age: {
+                    type: 'integer',
+                    description: 'User age',
+                    default: 30
+                  },
+                  score: {
+                    type: 'number',
+                    description: 'User score',
+                    default: 95.5
+                  },
+                  status: {
+                    type: 'string',
+                    description: 'User status',
+                    enum: ['active', 'inactive', 'pending'],
+                    default: 'active'
+                  },
+                  verified: {
+                    type: 'boolean',
+                    description: 'Verification status',
+                    default: true
+                  }
+                },
+                required: []
+              }
+            }
+          },
+          z
+            .object({ method: z.literal('elicitation/create') })
+            .passthrough() as any
+        );
+
+        const elicitResult = result as any;
+        return {
+          content: [
+            {
+              type: 'text',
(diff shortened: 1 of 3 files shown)


examples/servers/typescript/everything-server.ts
```diff
@@ -396,6 +396,82 @@ function createMcpServer() {
     }
   );
 
+  // SEP-1034: Elicitation with default values for all primitive types
+  mcpServer.registerTool(
+    'test_elicitation_sep1034_defaults',
```
felixweinberger, line 401: we'll probably want to start thinking about some modularity pattern soon if many SEPs mean we need to add additional tools / tests.

What we could do for example is to have the everything-server top level stay the same, but then have folders per SEP. If a tool, resource, etc. is used for a specific SEP's conformance test, that could be a sub-folder in everything-server instead of making this a 10k line long file.

Probably premature right now and we can do such a refactoring later, but just floating it now in case anyone has a cleverer idea.

pcarleton: yea good call. agreed on both, that we'll want to break it up, and that we can always break it up at the point it gets annoying since we'll have a better idea of how to do it well.

One other thing in that bucket is whether/how we want to communicate things like tool names.  e.g. an issue came up with someone wanting to test a gateway, but the gateway prefixes all the tool names.  Would be great to be able to easily tweak the test behavior to accommodate that type of thing, but I want to get a better feel for the normal development flow before we commit on how we'll accommodate more niche things.

src/scenarios/server/elicitation-defaults.ts
```diff
@@ -0,0 +1,293 @@
...
+        if (!('default' in properties.name)) {
+          stringErrors.push('Missing default field');
+        } else if (properties.name.default !== 'John Doe') {
+          stringErrors.push(
+            `Expected default "John Doe", got "${properties.name.default}"`
+          );
+        }
+      }
+
+      checks.push({
```
pcarleton, line 85: at first look, this and below seems super repetitive, but I think it's fine and probably easier to follow this way, and also easier to CTRL-F for the failing test.

I think we'll want to write some scenario "best practices", and this could be a good candidate (i.e. optimize for Ctrl+F on the slug).

Another one that could use some attention is how to do a check for an expected but missing call.  I do it kind of jankily in a few places, and think there's probably a good pattern we can do. just mentioning to keep in mind.

examples/servers/typescript/everything-server.ts
ihrpr: @felixweinberger  Can we add this tool requiremets to https://github.com/modelcontextprotocol/conformance/blob/main/SERVER_REQUIREMENTS.md?

felixweinberger: I'll add a follow-up

paoloricciuti:
I also commented on discord but this revealed a bug in the official typescript sdk: the elicitation is expecting a type of type string for enums but that's wrong per spec

https://json-schema.org/understanding-json-schema/reference/enum

I'm gonna open a PR to fix that but then we should figure out what to do here

PR #16, opened 2025-11-11
Add publish step

pcarleton:
so we can `npx` (also add typecheck step)


.github/workflows/ci.yml
```diff
@@ -12,16 +12,53 @@ concurrency:
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 24
+          cache: npm
+          registry-url: 'https://registry.npmjs.org'
+
+      - run: npm ci
+
+      - run: npm publish --provenance --access public
```
mattzcarey, line 52: will need this 
```
permissions:
  id-token: write
```

PR #28, opened 2025-11-13
[scenario] authz metadata discovery tests: compatibility with 2025-03-26 spec, prm root, and www-authenticate

pcarleton:


package.json
```diff
@@ -26,7 +26,7 @@
     "dist"
   ],
   "bin": {
-    "conformance": "dist/index.mjs"
+    "conformance": "dist/index.js"
```
pcarleton, line 29: tsdown 0.15 creates `index.js`, 0.16 creates `index.mjs`, so since we're pinning to `0.15` for now (which seems to also be what pkg.pr.new uses), this needs to be `.js`

package.json
```diff
@@ -43,7 +43,7 @@
     "vitest": "^4.0.5"
   },
   "dependencies": {
-    "@modelcontextprotocol/sdk": "^1.20.1",
+    "@modelcontextprotocol/sdk": "^1.22.0",
```
pcarleton, line 46: this is to pick up the fix that makes these tests pass

src/scenarios/client/auth/basic-dcr.test.ts
```diff
@@ -1,28 +0,0 @@
-import {
```
pcarleton, line 1: this is in index.test.ts now

src/scenarios/client/auth/basic-metadata.ts
```diff
@@ -0,0 +1,211 @@
...
+    ];
+
+    for (const slug of expectedSlugs) {
+      if (!this.checks.find((c) => c.id === slug)) {
+        this.checks.push({
+          id: slug,
+          name: `Expected Check Missing: ${slug}`,
+          description: `Expected Check Missing: ${slug}`,
+          status: 'FAILURE',
+          timestamp: new Date().toISOString()
```
felixweinberger, line 58: nit: should these point to a spec/RFC? either MCP or OAuth?

felixweinberger: Same with some of the other checks in this file

src/scenarios/client/auth/index.test.ts
```diff
@@ -0,0 +1,37 @@
...
+      'examples/clients/typescript/auth-test-broken1.ts'
+    );
+    const runner = new SpawnedClientRunner(clientPath);
+    await runClientAgainstScenario(runner, 'auth/basic-dcr', [
+      // There will be other failures, but this is the one that matters
+      'prm-priority-order'
+    ]);
+  });
+
+  // TODO: Add more negative tests here
```
felixweinberger, line 36: 👀

src/scenarios/client/auth/march-spec-backcompat.ts
```diff
@@ -0,0 +1,325 @@
...
+          id: 'no-prm-root',
+          name: 'No PRM at Root',
+          description:
+            'Client attempted to fetch PRM at root location, but March spec does not have PRM',
+          status: 'SUCCESS',
+          timestamp: new Date().toISOString(),
+          details: {
+            url: req.url,
+            path: req.path,
+            note: 'March spec behavior: no PRM available'
```
felixweinberger: should we point to a specific link in the march spec?

PR #35, opened 2025-11-17
feat: add conformance tests for SEP-1699 SSE polling

felixweinberger:
Adds conformance tests for SEP-1699 which introduces SSE polling via server-side disconnect.

## Summary
- Client test: validates retry field timing respect (MUST requirement)
- Client test: validates Last-Event-ID sent on reconnection (SHOULD requirement)
- Server test: validates priming event with id and empty data (SHOULD requirement)
- Server test: validates retry field sent before disconnect (SHOULD requirement)
- Fixed timing validation to only track GET requests (not POST)
- Added "very late" threshold (>2x retry) as FAILURE vs "slightly late" as WARNING

## Motivation and Context
SEP-1699 introduces SSE polling behavior. These conformance tests ensure SDK implementations correctly handle retry field timing, priming events, and graceful disconnection.

## How Has This Been Tested?
- All conformance checks pass against TypeScript SDK implementation
- Verified timing tolerances work correctly (2000ms retry, ~50ms early tolerance, ~200ms late tolerance)

## Breaking Changes
None.

## Types of changes
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/scenarios/server/sse-polling.ts
+++ b/src/scenarios/server/sse-polling.ts
@@ -2,86 +2,95 @@
  * SSE Polling conformance test scenarios for MCP servers (SEP-1699)
  *
  * Tests that servers properly implement SSE polling behavior including:
- * - Sending priming events with event ID and empty data
- * - Sending retry field before closing connection
+ * - Sending priming events with event ID and empty data on POST SSE streams
+ * - Sending retry field in priming events when configured
+ * - Replaying events when client reconnects with Last-Event-ID
  */
 
 import { ClientScenario, ConformanceCheck } from '../../types.js';
 import { EventSourceParserStream } from 'eventsource-parser/stream';
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js';
 
 export class ServerSSEPollingScenario implements ClientScenario {
   name = 'server-sse-polling';
   description =
-    'Test server sends SSE priming event and retry field (SEP-1699)';
+    'Test server sends SSE priming events on POST streams and supports event replay (SEP-1699)';
 
   async run(serverUrl: string): Promise<ConformanceCheck[]> {
     const checks: ConformanceCheck[] = [];
 
+    let sessionId: string | undefined;
+    let client: Client | undefined;
+    let transport: StreamableHTTPClientTransport | undefined;
+
     try {
-      // Make a GET request to establish SSE stream
-      const response = await fetch(serverUrl, {
-        method: 'GET',
-        headers: {
-          Accept: 'text/event-stream'
+      // Step 1: Initialize session with the server
+      client = new Client(
+        {
+          name: 'conformance-test-client',
+          version: '1.0.0'
+        },
+        {
+          capabilities: {
+            sampling: {},
+            elicitation: {}
+          }
         }
+      );
+
+      transport = new StreamableHTTPClientTransport(new URL(serverUrl));
+      await client.connect(transport);
+
+      // Extract session ID from transport (accessing internal state)
+      sessionId = (transport as unknown as { sessionId?: string }).sessionId;
+
+      if (!sessionId) {
+        checks.push({
+          id: 'server-sse-polling-session',
+          name: 'ServerSSEPollingSession',
+          description: 'Server provides session ID for SSE polling tests',
+          status: 'WARNING',
+          timestamp: new Date().toISOString(),
+          specReferences: [
+            {
+              id: 'SEP-1699',
+              url: 'https://github.com/modelcontextprotocol/modelcontextprotocol/issues/1699'
+            }
+          ],
+          details: {
+            message:
+              'Server did not provide session ID - SSE polling tests may not work correctly'
+          }
+        });
+      }
+
+      // Step 2: Make a POST request that returns SSE stream
+      // We need to use raw fetch to observe the priming event
+      const postResponse = await fetch(serverUrl, {
+        method: 'POST',
+        headers: {
+          'Content-Type': 'application/json',
+          Accept: 'text/event-stream, application/json',
+          ...(sessionId && { 'mcp-session-id': sessionId }),
+          'mcp-protocol-version': '2025-03-26'
+        },
+        body: JSON.stringify({
+          jsonrpc: '2.0',
+          id: 1,
+          method: 'tools/list',
+          params: {}
+        })
       });
 
-      // Check if server supports SSE GET endpoint
-      if (response.status === 405) {
+      if (!postResponse.ok) {
         checks.push({
-          id: 'server-sse-polling-endpoint',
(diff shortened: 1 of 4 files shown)


src/scenarios/server/sse-polling.ts
```diff
@@ -0,0 +1,288 @@
...
+    const checks: ConformanceCheck[] = [];
+
+    try {
+      // Make a GET request to establish SSE stream
+      const response = await fetch(serverUrl, {
+        method: 'GET',
+        headers: {
+          Accept: 'text/event-stream'
+        }
+      });
```
paoloricciuti: I might be wrong, but by reading the SEP, it seems like the main reason it was proposed and accepted is to allow POST requests to disconnect after priming the ID. Since this test is only doing a GET request, it would pass even if the server was only sending the ID on the standalone server, no?

src/scenarios/server/sse-polling.ts
paoloricciuti, line 1: Trying to run this locally with a server that doesn't send the ID, I don't get a failure, but I get a success 0/0 which is a bit weird but maybe intentional since it's back compatible? Maybe we should also show warnings in the result, tho?

src/scenarios/server/sse-polling.ts
```diff
@@ -0,0 +1,288 @@
...
+          }
+        });
+        return checks;
+      }
+
+      if (!response.ok) {
+        checks.push({
+          id: 'server-sse-polling-connection',
+          name: 'ServerSSEPollingConnection',
+          description: 'Server accepts SSE GET request',
```
paoloricciuti: See comment above, I think this should either apply to both GET and POST or even specifically to POST

src/scenarios/server/sse-polling.ts
```diff
@@ -0,0 +1,288 @@
...
+          ],
+          details: {
+            serverUrl,
+            statusCode: response.status,
+            message:
+              'Server does not support SSE GET endpoint (405 Method Not Allowed)'
+          }
+        });
+        return checks;
+      }
```
paoloricciuti: How do we plan to provide tests for when this behaviour is actually expected? Do we expect people to build multiple testing server and run them against various conformance suites?

src/scenarios/server/sse-polling.ts
```diff
@@ -0,0 +1,288 @@
...
+      try {
+        while (true) {
+          const { value: event, done } = await reader.read();
+
+          if (done) {
+            disconnected = true;
+            break;
+          }
+
+          eventCount++;
```
paoloricciuti, line 180: If we do expect multiple events, shouldn't we describe which event the server is expected to send for implementors of the everything-server?

src/scenarios/server/sse-polling.ts
felixweinberger: That's because I added this test to the `Pending` list, as https://github.com/modelcontextprotocol/typescript-sdk/pull/1129 is still WIP and hasn't landed yet. 

But I only plan to land this conformance scenario after shipping that anyway so might as well keep it failing for now, updating.

src/scenarios/server/sse-polling.ts
felixweinberger: Thanks for the review, and you're right - great catch. Will amend before publishing.

src/scenarios/server/sse-polling.ts
felixweinberger: I'm not 100% sure what you mean, but the idea of the conformance test suite is to have an `everything-server` in whatever SDK you're working on that implements all the features and then run the suite against it.

The `everything-server` for typescript is in this repo, but for others (e.g. Python here: https://github.com/modelcontextprotocol/python-sdk/pull/1604) it should be in the respective SDK repo.

The simplest approach is to have a single `everything-server` that implements all the required features for the tests and then run the entire suite against it.

Does that answer your question?

src/scenarios/server/sse-polling.ts
felixweinberger: Yes, will fix.

src/scenarios/server/sse-polling.ts
paoloricciuti: I actually run this before it was in the pending list 😄 The scenario was there but had 0/0 tests failing

src/scenarios/server/sse-polling.ts
felixweinberger: The `everything-server` will need to be implemented in this repo at https://github.com/modelcontextprotocol/conformance/tree/main/examples/servers/typescript

I probably shouldn't have pushed this PR yet as it only really makes sense once https://github.com/modelcontextprotocol/typescript-sdk/pull/1129 lands and the everything-server here can actually implement the required features.

Apologies, probably caused more confusion than necessary by pushing this draft.

src/scenarios/server/sse-polling.ts
paoloricciuti: Not really... my question is: returning 405 when the GET stream is not supported is one of the spec points. It would be good to have that in test form. But a server can either accept SSE or not so a single everything-server can't satisfy both 🤔

felixweinberger: That's fair, not sure we have that crisply defined yet what the behavior in that case should be - wdyt @pcarleton?

pcarleton: yea good question. In the client testing world, I have not done an "everything client", e.g. the "basic" client and the "auth" client are different clients people have to implement.

For client's it's a bit different because they need to be proactive instead of reactive.  Thinking about it, we could theoretically do an "everything client" that runs through many different spec features in a specified order.

In any event, to address the issue a few ideas I've been considering:
* Separating tests into "suites" that can be run all with the same client or command (e.g. everything-server suite is all against 1 server, client auth is all against 1 client run multiple times)
* Specifying a "run config" that matches up commands with the scenario to test, so if you do decide to do separate commands for separate scenarios, you can easily run a bunch together
* Passing in some kind of json "context" as an argument to the server, that tells it how to run
  * can make it easier to pass dynamic context in, like auth might need a client id and secret that's dynamically provisioned
  * can also change behavior of the server e.g. a stateless or stateful

The "simplest" is just require a different command, and implicitly people will need to figure out how to run them both in their CI suites

paoloricciuti: Yeah I think having different suites could work...but I would also allow for pass everything as a configuration and have an --all command that runs the different suites against the configured servers

src/scenarios/server/sse-polling.ts
paoloricciuti: Don't worry no confusion at all...I was just thinking about what the requirements for this test would be: are we expecting multiple events to be sent? Would those event only be Id, retry and data? Basically: should the server have a specific tool that this test can test against?

src/scenarios/server/sse-polling.ts
paoloricciuti: Btw this probably could actually be used to further improve the spec: is the id required/suggested also for the standalone stream? It's not very clear.

Issue #32, opened 2025-11-17
Client Auth: Scope Selection & Challenge Handling


pcarleton:
## Overview
Add conformance tests for OAuth scope selection strategies and scope challenge handling in client authentication flows.

## Specification References
- [Scope Selection Strategy](https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-selection-strategy)
- [Scope Challenge Handling](https://modelcontextprotocol.io/specification/draft/basic/authorization#scope-challenge-handling)
- [Error Handling](https://modelcontextprotocol.io/specification/draft/basic/authorization#error-handling)
- [Access Token Usage](https://modelcontextprotocol.io/specification/draft/basic/authorization#access-token-usage)

### Scope Selection (SHOULD)
- Clients **SHOULD** follow principle of least privilege
- Clients **SHOULD** use `scope` parameter from `WWW-Authenticate` header if provided
- If `scope` not available, use all scopes from `scopes_supported` in Protected Resource Metadata
- Omit `scope` parameter if `scopes_supported` is undefined

### Scope Challenge Handling
- Servers **SHOULD** respond with HTTP 403 for insufficient scope errors
- Servers **SHOULD** include `WWW-Authenticate` header with `error="insufficient_scope"`
- Servers **SHOULD** include `scope` parameter with required scopes
- Clients **SHOULD** respond by requesting new token with increased scopes via step-up authorization
- Clients **SHOULD** implement retry limits for scope upgrade attempts

### 401 Handling on Non-Initialize (MUST)
- Clients **MUST** parse `WWW-Authenticate` headers
- Clients **MUST** respond appropriately to HTTP 401 responses
- Authorization **MUST** be included in every HTTP request

### Positive Tests
1. Client correctly uses scope from `WWW-Authenticate` header
2. Client falls back to `scopes_supported` when scope not in header
3. Client handles HTTP 403 with `insufficient_scope` error
4. Client performs step-up authorization flow successfully
5. Client handles 401 on non-initialize requests

### Negative Tests
1. Client fails to include scope from `WWW-Authenticate` header
2. Client requests excessive scopes beyond what's needed
3. Client doesn't retry after `insufficient_scope` error
4. Client enters infinite retry loop on scope challenges
5. Client fails to handle 401 on non-initialize requests

## Implementation Notes
[...]

PR #36, opened 2025-11-18
[auth] Scenarios for scope selection

pcarleton:
## Motivation and Context

fixes https://github.com/modelcontextprotocol/conformance/issues/32

## Breaking Changes

tests and negative tests

## Types of changes

- [ ] Documentation update

## Checklist

- [ ] New and existing tests pass locally
  - needs a typescript sdk change to pass

## Additional context

I did a few related refactors:
- Inlined tests
- allowed piping test output to jq
- some fixes around how we handled WARNING, since it's a bunch of SHOULD's in this case

### DIFF
--- a/src/scenarios/client/auth/index.test.ts
+++ b/src/scenarios/client/auth/index.test.ts
@@ -14,10 +14,25 @@ beforeAll(() => {
   setLogLevel('error');
 });
 
+const skipScenarios = new Set<string>([
+  // Waiting on typescript-sdk support in bearerAuth middleware to include
+  // scope in WWW-Authenticate header
+  // https://github.com/modelcontextprotocol/typescript-sdk/pull/1133
+  'auth/scope-from-www-authenticate',
+  // Waiting on typescript-sdk support for using scopes_supported from PRM
+  // to request scopes.
+  // https://github.com/modelcontextprotocol/typescript-sdk/pull/1133
+  'auth/scope-from-scopes-supported'
+]);
+
 describe('Client Auth Scenarios', () => {
   // Generate individual test for each auth scenario
   for (const scenario of authScenariosList) {
     test(`${scenario.name} passes`, async () => {
+      if (skipScenarios.has(scenario.name)) {
+        // TODO: skip in a native way?
+        return;
+      }
       const runner = new InlineClientRunner(goodClient);
       await runClientAgainstScenario(runner, scenario.name);
     });


src/runner/client.ts
```diff
@@ -86,10 +86,10 @@ export async function runConformanceTest(
   // Scenario is guaranteed to exist by CLI validation
   const scenario = getScenario(scenarioName)!;
 
-  console.log(`Starting scenario: ${scenarioName}`);
+  console.error(`Starting scenario: ${scenarioName}`);
```
pcarleton, line 89: this is to make it possible to pipe verbose output to `jq` or `jless`

src/scenarios/client/auth/helpers/createServer.ts
```diff
@@ -33,10 +50,30 @@ export function createServer(
...
+          inputSchema: { type: 'object' }
+        }
+      ]
     };
   });
 
+  server.setRequestHandler(
+    CallToolRequestSchema,
+    async (request): Promise<CallToolResult> => {
+      if (request.params.name === 'test-tool') {
```
pcarleton, line 65: this is so we can test an "auth only required on a tool call" scenario.  It's via setRequestHandler because we're in the low level interface

src/scenarios/client/auth/helpers/mockTokenVerifier.ts
```diff
@@ -4,10 +4,23 @@ import type { ConformanceCheck } from '../../../../types.js';
...
 export class MockTokenVerifier implements OAuthTokenVerifier {
-  constructor(private checks: ConformanceCheck[]) {}
+  private tokenScopes: Map<string, string[]> = new Map();
+
+  constructor(
+    private checks: ConformanceCheck[],
+    private expectedScopes: string[] = []
+  ) {}
+
+  registerToken(token: string, scopes: string[]) {
```
pcarleton, line 14: we're keeping track of the scopes here as a convenience. we could also fetch them from the AS, and might want to switch to doing that in the future.

src/scenarios/client/auth/index.test.ts
```diff
@@ -1,37 +1,64 @@
 import { authScenariosList } from './index.js';
 import {
   runClientAgainstScenario,
-  SpawnedClientRunner
+  InlineClientRunner
 } from './test_helpers/testClient.js';
-import path from 'path';
+import { runClient as goodClient } from '../../../../examples/clients/typescript/auth-test.js';
```
pcarleton, line 6: I refactored the client tests to allow running them "inline" rather than spawning a subshell.

A subshell took a good 600ms, so the full auth suite took about 13s.  With this, it's about 200ms for the full suite, so vitest watch is like a live refresh.

src/scenarios/client/auth/index.test.ts
```diff
@@ -1,37 +1,64 @@
...
+import { runClient as ignore403Client } from '../../../../examples/clients/typescript/auth-test-ignore-403.js';
+import { setLogLevel } from '../../../../examples/clients/typescript/helpers/logger.js';
 
-describe('Client Auth Scenarios', () => {
-  const clientPath = path.join(
-    process.cwd(),
-    'examples/clients/typescript/auth-test.ts'
-  );
+beforeAll(() => {
+  setLogLevel('error');
```
pcarleton, line 14: without this, we get a bunch of annoying stdout lines littering the test output

src/scenarios/client/auth/test_helpers/testClient.ts
```diff
@@ -136,7 +136,9 @@ export async function runClientAgainstScenario(
       }
 
       // Verify that only the expected checks failed
-      const failures = nonInfoChecks.filter((c) => c.status === 'FAILURE');
+      const failures = nonInfoChecks.filter(
+        (c) => c.status === 'FAILURE' || c.status === 'WARNING'
```
pcarleton, line 140: I'm counting warning as failure in this test because we want the examples to not have warnings

examples/clients/typescript/auth-test-ignore-403.ts
```diff
@@ -0,0 +1,87 @@
...
+import { withOAuthRetry } from './helpers/withOAuthRetry.js';
+import { ConformanceOAuthProvider } from './helpers/ConformanceOAuthProvider.js';
+import { runAsCli } from './helpers/cliRunner.js';
+import { logger } from './helpers/logger.js';
+
+/**
+ * Broken client that only responds to 401, not 403.
+ * BUG: Ignores 403 responses which are used for step-up auth.
+ */
+export async function runClient(serverUrl: string): Promise<void> {
```
felixweinberger, line 20: more of a style nit: all these "broken" examples look very similar (identical?) except for the specific way in which they're "broken".

Could consider having a single implementation and having the specific way it's actually broken be a param to `createBrokenHandler`, a higher level function that creates the `handle401broken`.

```
  type ClientBehavior =
    | { type: 'well-behaved' }  // or just omit for default
    | { type: 'ignore-resource-metadata' }
    | { type: 'ignore-scope' }
    | { type: 'partial-scopes'; scope: string }
    | { type: 'ignore-403' };
```

Could make this less verbose.

felixweinberger: Potentially the well-behaved client could also share the same implementation.

felixweinberger: Maybe we prefer explicitness over conciseness though for conformance tests, so definitely just a thought 🤔

package.json
```diff
@@ -33,6 +33,7 @@
     "@types/express": "^5.0.3",
     "@types/node": "^22.10.2",
     "@typescript/native-preview": "^7.0.0-dev.20251030.1",
+    "cors": "^2.8.5",
```
felixweinberger, line 36: Is adding `cors` here intentional for this PR? Doesn't seem like we're importing this anywhere?

examples/clients/typescript/auth-test-ignore-403.ts
pcarleton: agreed on the repetitiveness, i've been resisting the urge to de-dupe too aggressively in order to let abstractions fall out from a few more iterations on tests before committing to one.

package.json
pcarleton: when I ran `npm link` this started complaining, i think because the everything server uses it (even if we have a package.json in the server client repo).

I'd prefer to keep it to let the link flow be smoother even if it's a little odd.

Issue #34, opened 2025-11-17
Client Auth: Client Registration Methods


pcarleton:
## Overview
Add conformance tests for all three OAuth client registration mechanisms: Client ID Metadata Documents (CIMD), Pre-registration, and Dynamic Client Registration (DCR).

## Specification References
- [Client Registration Approaches](https://modelcontextprotocol.io/specification/draft/basic/authorization#client-registration-approaches)
- [Client ID Metadata Documents](https://modelcontextprotocol.io/specification/draft/basic/authorization#client-id-metadata-documents)
- [Preregistration](https://modelcontextprotocol.io/specification/draft/basic/authorization#preregistration)
- [Dynamic Client Registration](https://modelcontextprotocol.io/specification/draft/basic/authorization#dynamic-client-registration)
- [OAuth Client ID Metadata Document (draft-ietf-oauth-client-id-metadata-document-00)](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-client-id-metadata-document-00)
- [RFC 7591 - OAuth 2.0 Dynamic Client Registration Protocol](https://datatracker.ietf.org/doc/html/rfc7591)

### Client ID Metadata Documents (SHOULD)
**For MCP Clients:**
- Clients **MUST** host metadata document at HTTPS URL following RFC requirements
- `client_id` URL **MUST** use "https" scheme and contain path component
- Metadata document **MUST** include: `client_id`, `client_name`, `redirect_uris`
- Clients **MUST** ensure `client_id` value in metadata matches document URL exactly
- Clients **MAY** use `private_key_jwt` for client authentication with appropriate JWKS configuration

**For Authorization Servers:**
- **SHOULD** fetch metadata documents when encountering URL-formatted client_ids
- **MUST** validate fetched document's `client_id` matches the URL exactly
- **SHOULD** cache metadata respecting HTTP cache headers
- **MUST** validate redirect URIs against those in metadata document
- **MUST** validate document structure is valid JSON and contains required fields
- **SHOULD** follow security considerations in CIMD spec Section 6

### Pre-registration (SHOULD)
- Clients **SHOULD** support option for static client credentials
- Support hardcoded client ID or user-entered credentials
- Handle both confidential and public client types

### Dynamic Client Registration (MAY)
- Clients and authorization servers **MAY** support RFC 7591
- Included for backwards compatibility with earlier MCP versions

### Client ID Metadata Documents - Positive Tests
1. Client uses HTTPS URL as `client_id`
[...]

PR #37, opened 2025-11-18
[auth] Implement CIMD conformance test

pcarleton:
partially addresses https://github.com/modelcontextprotocol/conformance/issues/34

depends on typescript sdk change here:
https://github.com/modelcontextprotocol/typescript-sdk/pull/1127


examples/clients/typescript/auth-test-no-cimd.ts
```diff
@@ -0,0 +1,44 @@
+#!/usr/bin/env node
+
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js';
+import { withOAuthRetry } from './helpers/withOAuthRetry.js';
+import { runAsCli } from './helpers/cliRunner.js';
+import { logger } from './helpers/logger.js';
```
mattzcarey: nit: .js extension is not needed with tsx

examples/clients/typescript/auth-test-no-cimd.ts
```diff
@@ -0,0 +1,44 @@
...
+    { capabilities: {} }
+  );
+
+  // BUG: Not passing clientMetadataUrl, so client will use DCR
+  // even when server supports client_id_metadata_document_supported
+  const oauthFetch = withOAuthRetry(
+    'test-auth-client-no-cimd',
+    new URL(serverUrl)
+    // Missing: handle401, clientMetadataUrl
+  )(fetch);
```
mattzcarey, line 32: as a user am I expected to understand what this does? Like is this a learning moment for me. If so I think this method could be clearer.

pcarleton: yea the readability is a bit busted here for now. we need to upstream some things to the `withOauth` middleware in typescript and then refactor. for now i think this will be a little gross.

examples/clients/typescript/auth-test-no-cimd.ts
pcarleton: fixed all these 👍

PR #38, opened 2025-11-18
Set logging level before calling test_tool_with_logging

mikekistler:
## Motivation and Context

This PR makes a small change to the test_tool_with_logging test, to set the logLevel before calling the tool.

Currently the test does not set the logLevel and expects the default logLevel lets through at least informational messages.
I think this assumption and not clearly required in the spec.

The current behavior of the C# SDK is that no log messages are sent unless the client sets the log level.

## How Has This Been Tested?

Tested with a server built with the C# MCP SDK.

## Breaking Changes

This should not affect any implementations currently passing the tests.

## Types of changes

- [ ] Documentation update

## Checklist


pcarleton:
👍  makes sense, that's consistent with the flow chart in the spec:
https://modelcontextprotocol.io/specification/draft/server/utilities/logging

thanks for this change!

PR #55, opened 2025-11-24
Add client credentials conformance tests (JWT and Basic auth)

pcarleton:
## Summary

Adds two new conformance test scenarios for SEP-1046 OAuth client_credentials flow:

- **`auth/client-credentials-jwt`**: Tests client_credentials grant with `private_key_jwt` authentication (RFC 7523 Section 2.2)
- **`auth/client-credentials-basic`**: Tests client_credentials grant with `client_secret_basic` authentication

### Well-known test credentials

Both scenarios use well-known test credentials that SDK implementations should use:

**For JWT (private_key_jwt):**
- Client ID: `conformance-test-client`
- EC P-256 private key (in conformance client implementations)
- Public key JWK is in `client-credentials.ts`

**For Basic auth (client_secret_basic):**
- Client ID: `conformance-test-client`  
- Client Secret: `conformance-test-secret`

### Changes

- Add `ClientCredentialsJwtScenario` - verifies JWT assertion signature and claims
- Add `ClientCredentialsBasicScenario` - verifies HTTP Basic auth header
- Extend `createAuthServer` to pass Authorization header to `onTokenRequest` callback
- Update `mockTokenVerifier` to accept `cc-token-*` prefix

### Note

The existing TypeScript test client uses authorization_code flow, so these scenarios will fail with it. SDK implementations need to add client_credentials support to pass these tests.

### DIFF
--- a/src/scenarios/client/auth/client-credentials.ts
+++ b/src/scenarios/client/auth/client-credentials.ts
@@ -103,6 +103,28 @@ export class ClientCredentialsJwtScenario implements Scenario {
             clockTolerance: 30
           });
 
+          // Verify iss claim matches expected client_id
+          if (payload.iss !== CONFORMANCE_TEST_CLIENT_ID) {
+            this.checks.push({
+              id: 'client-credentials-jwt-iss',
+              name: 'ClientCredentialsJwtIss',
+              description: `JWT iss claim '${payload.iss}' does not match expected client_id '${CONFORMANCE_TEST_CLIENT_ID}'`,
+              status: 'FAILURE',
+              timestamp,
+              specReferences: [SpecReferences.RFC_JWT_CLIENT_AUTH],
+              details: {
+                expected: CONFORMANCE_TEST_CLIENT_ID,
+                actual: payload.iss
+              }
+            });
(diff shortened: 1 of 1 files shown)


src/scenarios/client/auth/client-credentials.ts
```diff
@@ -0,0 +1,378 @@
...
+        // Verify JWT signature and claims using the generated public key
+        try {
+          // Per RFC 7523bis, audience MUST be the issuer identifier
+          const { payload } = await jose.jwtVerify(clientAssertion, publicKey, {
+            audience: issuerUrl,
+            clockTolerance: 30
+          });
+
+          // Verify sub claim matches expected client_id
+          if (payload.sub !== CONFORMANCE_TEST_CLIENT_ID) {
```
felixweinberger, line 129: nit: Should we assert on `payload.iss` as well?

Looks like `iss` is required by this RFC: https://datatracker.ietf.org/doc/html/rfc7523#section-3

pcarleton: ah yea good call.

PR #80, opened 2025-12-15
fix: validate scope checks ran and report overall test status

pcarleton:
## Summary

- Add FAILURE checks to scope scenarios when authorization flow doesn't complete
- Report client timeout and exit errors in test results  
- Add OVERALL: PASSED/FAILED summary at end of test output
- Exit with code 1 on any failure (check failures, client timeout, or exit error)

## Problem

Previously, if the OAuth authorization flow didn't complete (e.g., client timed out or exited with an error before the authorization request reached the server), the scope checks were never triggered and the test would incorrectly report success like "Passed: 6/6, 0 failed".

## Solution

The following scope scenarios now emit a FAILURE if their expected scope check wasn't recorded:
- `ScopeFromWwwAuthenticateScenario`
- `ScopeFromScopesSupportedScenario`  
- `ScopeOmittedWhenUndefinedScenario`

Additionally, `printClientResults` now:
- Accepts optional `clientOutput` to check for timeout/exit errors
- Reports warnings for client timeout or non-zero exit codes
- Shows an overall PASSED/FAILED summary
- Returns `overallFailure` boolean for proper exit code

## Example output (failure case)

```
Test Results:
Passed: 6/7, 1 failed, 0 warnings

⚠️  CLIENT EXITED WITH ERROR (code 1) - Test may be incomplete

Failed Checks:
  - Client scope selection from scopes_supported: Client did not complete authorization flow - scope check could not be performed

❌ OVERALL: FAILED
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -125,7 +125,7 @@ program
           totalFailed += failed;
           totalWarnings += warnings;
 
-          const status = failed === 0 ? '✓' : '✗';
+          const status = failed === 0 && warnings === 0 ? '✓' : '✗';
           const warningStr = warnings > 0 ? `, ${warnings} warnings` : '';
           console.log(
             `${status} ${result.scenario}: ${passed} passed, ${failed} failed${warningStr}`
@@ -145,7 +145,7 @@ program
         console.log(
           `\nTotal: ${totalPassed} passed, ${totalFailed} failed, ${totalWarnings} warnings`
         );
-        process.exit(totalFailed > 0 ? 1 : 0);
+        process.exit(totalFailed > 0 || totalWarnings > 0 ? 1 : 0);
       }
 
       // Require either --scenario or --suite


src/scenarios/client/auth/scope-handling.ts
```diff
@@ -73,6 +73,21 @@ export class ScopeFromWwwAuthenticateScenario implements Scenario {
   }
 
   getChecks(): ConformanceCheck[] {
+    // Emit failure check if expected scope check didn't run
+    const hasScopeCheck = this.checks.some(
+      (c) => c.id === 'scope-from-www-authenticate'
+    );
+    if (!hasScopeCheck) {
+      this.checks.push({
```
felixweinberger, line 81: nit: mutations in a `getSomething` might be slightly weird, but I deduplication above makes it idempotent at least.

src/index.ts
```diff
@@ -173,8 +173,12 @@ program
         timeout
       );
 
-      const { failed } = printClientResults(result.checks, verbose);
-      process.exit(failed > 0 ? 1 : 0);
+      const { overallFailure } = printClientResults(
+        result.checks,
+        verbose,
+        result.clientOutput
```
felixweinberger, line 179: more of a suggestion: IIUC we're only showing these output summaries for single scenarios, could consider doing that for test suites as well.

src/scenarios/client/auth/scope-handling.ts
pcarleton: good callout. i don't love this pattern in general of throwing in failed checks if they're missing. It feels easy to forget, and I'd rather have you decide on the language wherever you're defining the check. like either the checks are fully "in-line", or they're fully after the fact. 

 I will have to think if we can swap in something better here, but will leave this as is for now.

PR #64, opened 2025-11-26
Add server OAuth protection conformance tests

tobinsouth:
## Summary

Starting a discussion on adding OAuth conformance tests for _servers_ behind an `--auth` flag. The scenarios are still somewhat basic and only tests resource server protection and metadata. This PR stops at the point where a bearer would need to be passed, only validating server OAuth infrastructure (discovery, metadata, HTTP responses) without requiring access tokens.

In general, the concerns of the auth server are separated from the resource server, which puts auth conformance somewhat out of scope of SDKs. Having said that, it's a huge pain in the butt to achieve full auth conformance for Auth vendors, open-source auth solutions, or custom auth setups. This repo feels like a great place to put a reliable test of auth conformance testing for AS' wrapping servers against a high quality client.

### New Test Scenarios (10 total)

| Scenario                              | Purpose                                         |
| ------------------------------------- | ----------------------------------------------- |
| `server/auth-prm-discovery`           | Protected Resource Metadata endpoint (RFC 9728) |
| `server/auth-as-metadata-discovery`   | Authorization Server metadata (RFC 8414)        |
| `server/auth-discovery-mechanism`     | Discovery endpoint availability                 |
| `server/auth-as-pkce-support`         | PKCE S256 support (RFC 7636)                    |
| `server/auth-as-cimd-supported`       | Client ID Metadata Document support             |
| `server/auth-as-token-auth-methods`   | Token endpoint auth methods                     |
| `server/auth-as-grant-types`          | Grant types (client_credentials/SEP-1046)       |
| `server/auth-prm-resource-validation` | Resource URI format validation                  |
| `server/auth-401-unauthorized`        | 401 response for unauthenticated requests       |
| `server/auth-www-authenticate-header` | WWW-Authenticate header format (RFC 6750)       |

### New Files

```
src/scenarios/server/auth/
├── helpers/
│   ├── auth-fetch.ts           # HTTP client with header parsing
│   └── as-metadata.ts          # AS metadata fetching utilities
├── scenarios/
│   ├── prm-discovery.ts
│   ├── as-metadata-discovery.ts
│   ├── discovery-mechanism.ts
│   ├── as-pkce-support.ts
│   ├── as-cimd-supported.ts
│   ├── as-token-auth-methods.ts
│   ├── as-grant-types.ts
│   ├── prm-resource-validation.ts
│   ├── unauthorized-response.ts
│   └── www-authenticate-header.ts
├── spec-references.ts          # RFC/spec reference definitions
└── index.ts
```

[...]

PR #105, opened 2026-01-14
feat: Add server auth conformance tests

pcarleton:
## Summary

Adds conformance testing for MCP server OAuth authentication implementations.

## Design Philosophy

The MCP authorization spec separates the **Authorization Server (AS)** from the **Resource Server (RS/MCP server)**. This reflects the reality that authorization is often better handled by dedicated tools (Auth0, Keycloak, better-auth, etc.) rather than reimplemented in every SDK.

This conformance suite tests **MCP server (RS) implementations**, not AS implementations. SDKs are not expected to provide a full AS - they just need to:

1. Return 401 + WWW-Authenticate for unauthenticated requests
2. Serve Protected Resource Metadata (PRM) pointing to an AS
3. Validate Bearer tokens (via introspection or other means)

The conformance suite provides a **fake AS** for testing, so SDK authors can verify their RS implementation without setting up real auth infrastructure. Alternatively, servers can integrate with their own AS if they prefer.

### CLI Commands

```bash
# Run auth tests against an MCP server
# The AS is spun up and URL passed via MCP_CONFORMANCE_AUTH_SERVER_URL environment variable. 
conformance server --suite auth --command 'npx path/to/test-server.ts'

# Run auth tests against an MCP server (with server providing AS)
conformance server --suite auth --url http://localhost:3000/mcp

# Interactive mode for servers requiring browser-based login
conformance server --suite auth --url http://localhost:3000/mcp --interactive

# Standalone fake AS for manual testing
conformance fake-auth-server [--port <port>]
```

### New Files

- `src/fake-auth-server.ts` - Standalone fake OAuth AS CLI
- `src/scenarios/server-auth/` - Server auth scenario directory
  - `basic-dcr-flow.ts` - Tests complete OAuth flow (401 → PRM → AS metadata → CIMD/DCR → Token → Auth'd request)
  - `helpers/oauth-client.ts` - Observation middleware and `ConformanceOAuthProvider`
  - `spec-references.ts` - RFC/spec references (2025-11-25)

### Modified Files

- `src/index.ts` - Extended server command with `--suite auth`, `--interactive`
- `src/runner/server.ts` - Added `runServerAuthConformanceTest()` and `startFakeAuthServer()`
- `src/scenarios/client/auth/helpers/createAuthServer.ts` - Added `/introspect` endpoint (RFC 7662), CIMD detection
- `src/types.ts` - Added `ClientScenarioOptions` with `interactive` flag
- `package.json` - Added `fake-auth-server` bin entry

### Interactive Mode

For servers that integrate with their own AS requiring browser-based login, use `--interactive`:

```bash
conformance server --suite auth --url http://localhost:3000/mcp --interactive
```

This starts a callback server on port 3333 and displays the authorization URL for you to open in a browser.

### CIMD vs DCR

The fake AS defaults to `client_id_metadata_document_supported: true`, making **CIMD (URL-based client IDs)** the preferred method over DCR.

## Architecture

```
┌─────────────────────────────────────────────────────────────────┐
│                    Conformance Suite                             │
│                                                                  │
│  ┌──────────────┐     ┌──────────────────────────────────────┐  │
│  │   Fake AS    │     │  Checker Client (typescript-sdk)      │  │
│  │  (provided)  │◄────│  - Observes HTTP requests             │  │
│  │              │     │  - Generates conformance checks       │  │
│  └──────────────┘     └──────────┬───────────────────────────┘  │
│        ▲                         │                               │
│        │ PRM points here         │ Auth'd MCP requests          │
│        │ (or to own AS)          │                               │
│        │                         ▼                               │
│  ┌──────────────────────────────────────────────────────────┐   │
│  │              MCP Server Under Test (RS)                   │   │
│  │  - Returns 401 + WWW-Authenticate for unauth requests    │   │
│  │  - Serves PRM at /.well-known/oauth-protected-resource   │   │
│  │  - Validates Bearer tokens                                │   │
│  └──────────────────────────────────────────────────────────┘   │
└─────────────────────────────────────────────────────────────────┘
```

## Conformance Checks

- `auth-401-response` - Server returns 401 for unauthenticated requests
- `auth-www-authenticate-header` - 401 includes WWW-Authenticate header with Bearer scheme
- `auth-resource-metadata-param` - WWW-Authenticate includes resource_metadata parameter
- `auth-prm-discovery` - PRM endpoint responds at `.well-known/oauth-protected-resource`
- `auth-prm-authorization-servers` - PRM contains authorization_servers array
- `auth-as-metadata-discovery` - AS metadata discovery works
- `auth-as-metadata-fields` - AS metadata includes required fields (issuer, authorization_endpoint, token_endpoint, PKCE S256)
- `auth-dcr-registration` - Dynamic Client Registration succeeds (when DCR used)
- `auth-dcr-response` - DCR response contains client credentials
- `auth-token-request` - Token acquisition succeeds
- `auth-token-response` - Token response contains access_token
- `auth-authenticated-request` - Authenticated MCP call succeeds
- `auth-flow-completion` - Complete OAuth flow succeeded

## Testing Example

```bash
# Start an MCP server with OAuth support
cd ~/code/mcp/typescript-sdk
pnpm exec tsx examples/server/src/simpleStreamableHttp.ts --oauth

# Run conformance tests (uses --interactive since example uses better-auth)
conformance server --suite auth --url http://localhost:3000/mcp --interactive
```

## Related PRs

- modelcontextprotocol/typescript-sdk#1384 - Adds `auth-test-server.ts` to typescript-sdk conformance directory


tobinsouth:
I think the fake AS is a great solution here and focusing on the RS is the right design choice (better than #64 !). If we do a AS specific conformance check, I think it should be separate tooling. Maybe something someone at Okta / Entra / Auth0 / Twillio / Arcade / WorkOS etc want to take on.

src/scenarios/server-auth/basic-dcr-flow.ts
```diff
@@ -0,0 +1,583 @@
...
+        });
+
+        // Check for resource_metadata parameter
+        if (wwwAuth.params.resource_metadata) {
+          checks.push({
+            id: 'auth-resource-metadata-param',
+            name: 'Resource Metadata URL in WWW-Authenticate',
+            description:
+              'WWW-Authenticate header includes resource_metadata parameter',
+            status: 'INFO',
```
nbarbettini, line 255:
```suggestion
            status: 'SUCCESS',
```

nbarbettini: My reading of [authorization s. 4.2](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#protected-resource-metadata-discovery-requirements) is that `resource_metadata` is a MUST in MCP if the AS returns WWW-Authenticate. I think that means there should also be a corresponding failure check if it is missing.

src/scenarios/server-auth/basic-dcr-flow.ts
```diff
@@ -0,0 +1,583 @@
...
+          issues.push('code_challenge_methods_supported must include S256');
+
+        const allValid = issues.length === 0;
+
+        checks.push({
+          id: 'auth-as-metadata-fields',
+          name: 'AS Metadata Required Fields',
+          description: allValid
+            ? 'Authorization Server metadata includes all required fields'
+            : `Authorization Server metadata issues: ${issues.join(', ')}`,
```
nbarbettini, line 416: Not blocking: would you be open to disaggregating this check into a check per field or per group of related fields? I think it could produce more easily-digestible results when specific clusters of things like PKCE fail.

src/scenarios/server-auth/basic-dcr-flow.ts
pcarleton: yep that sounds right

src/scenarios/server-auth/basic-dcr-flow.ts
pcarleton: yea that seems reasonable! let's make it play nicely with the way we're doing metadata checks in the AS-only tests, i think we should use the same checks in both places

PR #115, opened 2026-01-22
feat: add DNS rebinding protection conformance tests

### LINKED ISSUE #103: [Tiering] Add conformance tests for localhost DNS rebinding protection
## Summary

Localhost MCP servers without authentication are vulnerable to DNS rebinding attacks. The conformance suite should help developers verify their DNS rebinding protection is working correctly.

## Motivation

Many MCP servers run on localhost during development or for local-only use cases, often without authentication. In this configuration, the **only** protection against malicious websites interacting with the server is Origin header validation.

**The attack:**
1. User visits `evil.com` in their browser
2. `evil.com`'s JavaScript makes fetch requests to `http://localhost:3000/mcp`
3. Browser sends `Origin: https://evil.com` header
4. If server doesn't validate Origin, attacker can interact with user's local MCP server
5. Attacker can invoke tools, read resources, etc.

This conformance test helps developers answer: "Is my localhost server's DNS rebinding protection actually working?"

## Scope

These tests are **strictly scoped to localhost servers** (detected by server URL being `localhost`, `127.0.0.1`, or `[::1]`). Remote servers with authentication have different security models and are out of scope for this issue.

### Check: `localhost-origin-rebinding-rejected`
**Spec:** "Servers MUST validate the Origin header... If the Origin header is present and invalid, servers MUST respond with HTTP 403 Forbidden."

Send request with `Origin: https://attacker.example.com` to localhost server → expect HTTP 403.

### Check: `localhost-origin-valid-accepted`
Verify server accepts requests from legitimate localhost origins.

Send request with `Origin: http://localhost:PORT` → expect normal MCP response (not 403).

## Spec Requirements

| Keyword | Statement |
|---------|-----------|
| MUST | Servers MUST validate the Origin header on all incoming connections to prevent DNS rebinding attacks |
| MUST | If the Origin header is present and invalid, servers MUST respond with HTTP 403 Forbidden |
| SHOULD | When running locally, servers SHOULD bind only to localhost (127.0.0.1) rather than all network interfaces (0.0.0.0) |

## Implementation Notes

**Valid origins for localhost servers:**
- `http://localhost` / `http://localhost:PORT`
- `http://127.0.0.1` / `http://127.0.0.1:PORT`
- `http://[::1]` / `http://[::1]:PORT` (IPv6)

**Invalid origins (should 403):**
- Any non-localhost origin (e.g., `https://evil.com`, `https://example.com`)

**Edge cases to consider:**
- Missing Origin header: Non-browser clients don't send Origin. Servers likely need to accept missing Origin, meaning protection only applies to browser-originated requests.
- `null` Origin: Sent by file:// origins. Servers may reasonably accept or reject.

**Test activation:**
- Auto-detect: Only run when server URL is localhost/127.0.0.1/[::1]
- Or explicit flag: `--localhost-security`

## Spec References

- [HTTP Transport Security](https://modelcontextprotocol.io/specification/draft/basic/transports#security)

pcarleton:
## Summary

Add conformance tests to verify localhost MCP servers properly validate Host headers to prevent DNS rebinding attacks.

Closes #103

### New scenario: `dns-rebinding-protection`

Two checks:
- **`localhost-host-rebinding-rejected`**: Server returns 4xx for non-localhost Host headers (e.g., `Host: evil.example.com`)
- **`localhost-host-valid-accepted`**: Server accepts valid localhost Host headers

### Implementation details

- Uses `undici` for making HTTP requests with custom Host headers (native fetch doesn't allow this)
- Accepts any 4xx status code as valid rejection (403, 421, etc.)
- Tests require localhost server URL - non-localhost URLs return FAILURE with clear message

### Additional changes

- Updated `everything-server` example to use `createMcpExpressApp()` for DNS rebinding protection
- Added negative test case `no-dns-rebinding-protection.ts` that intentionally omits protection

## Testing

Verified against:
- ✅ TypeScript SDK (everything-server with `createMcpExpressApp()`)
- ✅ Python SDK (everything-server with auto-enabled protection)
- ✅ Negative test case (correctly fails)

## Related

- GHSA: https://github.com/modelcontextprotocol/typescript-sdk/security/advisories/GHSA-w48q-cv73-mx4w
- Spec: https://modelcontextprotocol.io/specification/draft/basic/transports#security

### DIFF
--- a/src/scenarios/server/dns-rebinding.ts
+++ b/src/scenarios/server/dns-rebinding.ts
@@ -1,17 +1,23 @@
 /**
  * DNS Rebinding Protection test scenarios for MCP servers
  *
- * Tests that localhost MCP servers properly validate Host headers to prevent
- * DNS rebinding attacks. See GHSA-w48q-cv73-mx4w for details on the attack.
+ * Tests that localhost MCP servers properly validate Host or Origin headers
+ * to prevent DNS rebinding attacks. See GHSA-w48q-cv73-mx4w for details.
  */
 
 import { ClientScenario, ConformanceCheck } from '../../types';
 import { request } from 'undici';
 
-const SPEC_REFERENCE = {
-  id: 'MCP-DNS-Rebinding-Protection',
-  url: 'https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices#local-mcp-server-compromise'
-};
+const SPEC_REFERENCES = [
+  {
+    id: 'MCP-DNS-Rebinding-Protection',
+    url: 'https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices#local-mcp-server-compromise'
+  },
+  {
+    id: 'MCP-Transport-Security',
+    url: 'https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#security-warning'
+  }
+];
 
 /**
  * Check if URL is a localhost URL
@@ -36,17 +42,20 @@ function getHostFromUrl(serverUrl: string): string {
 }
 
 /**
- * Send an MCP initialize request with a custom Host header
+ * Send an MCP initialize request with custom Host and Origin headers.
+ * Both headers are set to the same value so that servers checking either
+ * Host or Origin will properly detect the rebinding attempt.
  */
-async function sendRequestWithHost(
+async function sendRequestWithHostAndOrigin(
   serverUrl: string,
-  hostHeader: string
+  hostOrOrigin: string
 ): Promise<{ statusCode: number; body: unknown }> {
   const response = await request(serverUrl, {
     method: 'POST',
     headers: {
       'Content-Type': 'application/json',
-      Host: hostHeader,
+      Host: hostOrOrigin,
+      Origin: `http://${hostOrOrigin}`,
       Accept: 'application/json, text/event-stream'
     },
     body: JSON.stringify({
@@ -54,7 +63,7 @@ async function sendRequestWithHost(
       id: 1,
       method: 'initialize',
       params: {
-        protocolVersion: '2024-11-05',
+        protocolVersion: '2025-11-25',
         capabilities: {},
         clientInfo: { name: 'conformance-dns-rebinding-test', version: '1.0.0' }
       }
@@ -86,14 +95,14 @@ website tricks a user's browser into making requests to the local server.
 1. User visits malicious website (e.g., evil.com)
 2. evil.com's DNS is configured to resolve to 127.0.0.1
 3. Browser makes request to evil.com which actually goes to localhost
-4. Without Host header validation, the local MCP server processes the request
+4. Without Host/Origin header validation, the local MCP server processes the request
 
 **Requirements:**
-- Server **MUST** validate the Host header on incoming requests
-- Server **MUST** reject requests with non-localhost Host headers (HTTP 4xx)
-- Server **MUST** accept requests with valid localhost Host headers
+- Server **MUST** validate the Host or Origin header on incoming requests
+- Server **MUST** reject requests with non-localhost Host/Origin headers (HTTP 4xx)
+- Server **MUST** accept requests with valid localhost Host/Origin headers
 
-**Valid localhost hosts:** \`localhost\`, \`127.0.0.1\`, \`[::1]\` (with optional port)
+**Valid localhost values:** \`localhost\`, \`127.0.0.1\`, \`[::1]\` (with optional port)
 
 **Note:** This test requires a localhost server URL. Non-localhost URLs will fail.
 
@@ -107,17 +116,19 @@ See: https://github.com/modelcontextprotocol/typescript-sdk/security/advisories/
     const rejectedCheckBase = {
       id: 'localhost-host-rebinding-rejected',
       name: 'DNSRebindingRejected',
-      description: 'Server rejects requests with non-localhost Host headers',
+      description:
+        'Server rejects requests with non-localhost Host/Origin headers',
       timestamp,
-      specReferences: [SPEC_REFERENCE]
+      specReferences: SPEC_REFERENCES
     };
 
     const acceptedCheckBase = {
       id: 'localhost-host-valid-accepted',
       name: 'LocalhostHostAccepted',
-      description: 'Server accepts requests with valid localhost Host headers',
+      description:
+        'Server accepts requests with valid localhost Host/Origin headers',
       timestamp,
-      specReferences: [SPEC_REFERENCE]
+      specReferences: SPEC_REFERENCES
     };
 
     // First check: Is this a localhost URL?
@@ -128,13 +139,13 @@ See: https://github.com/modelcontextprotocol/typescript-sdk/security/advisories/
 
       checks.push({
         ...rejectedCheckBase,
-        status: 'FAILURE' as const,
+        status: 'FAILURE',
         errorMessage,
         details
       });
       checks.push({
         ...acceptedCheckBase,
-        status: 'FAILURE' as const,
+        status: 'FAILURE',
         errorMessage,
         details
       });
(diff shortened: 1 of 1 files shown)


examples/servers/typescript/everything-server.ts
```diff
@@ -18,6 +18,7 @@ import {
   EventId,
   StreamId
 } from '@modelcontextprotocol/sdk/server/streamableHttp.js';
+import { createMcpExpressApp } from '@modelcontextprotocol/sdk/server/express.js';
```
pcarleton, line 21: note: this example is going to get deleted in a follow-up PR to make the typescript-sdk example the source of truth.  updating it in place for now.

examples/servers/typescript/no-dns-rebinding-protection.ts
```diff
@@ -0,0 +1,157 @@
...
+import {
+  StreamableHTTPServerTransport,
+  EventStore,
+  EventId,
+  StreamId
+} from '@modelcontextprotocol/sdk/server/streamableHttp.js';
+import express from 'express';
+import { randomUUID } from 'crypto';
+
+// In-memory event store for SSE
```
pcarleton: this negative example shouldn't require an event-store or sessions or any of the complexity. it can be the most dead simple server imaginable.

src/scenarios/server/dns-rebinding.ts
```diff
@@ -0,0 +1,263 @@
...
+ * Tests that localhost MCP servers properly validate Host headers to prevent
+ * DNS rebinding attacks. See GHSA-w48q-cv73-mx4w for details on the attack.
+ */
+
+import { ClientScenario, ConformanceCheck } from '../../types';
+import { request } from 'undici';
+
+const SPEC_REFERENCE = {
+  id: 'MCP-DNS-Rebinding-Protection',
+  url: 'https://modelcontextprotocol.io/specification/draft/basic/transports#security'
```
pcarleton: let's change this to:
https://modelcontextprotocol.io/specification/2025-11-25/basic/security_best_practices#local-mcp-server-compromise

src/scenarios/server/dns-rebinding.ts
```diff
@@ -0,0 +1,263 @@
...
+  };
+}
+
+export class DNSRebindingProtectionScenario implements ClientScenario {
+  name = 'dns-rebinding-protection';
+  description = `Test DNS rebinding protection for localhost servers.
+
+**Server Implementation Requirements:**
+
+DNS rebinding attacks occur when an attacker's domain resolves to a localhost IP,
```
pcarleton: let's clarify here that the vulnerable servers are localhost servers w/o https and without auth (see the language in the GHSA for inspiration).

src/scenarios/server/dns-rebinding.ts
```diff
@@ -0,0 +1,263 @@
...
+See: https://github.com/modelcontextprotocol/typescript-sdk/security/advisories/GHSA-w48q-cv73-mx4w`;
+
+  async run(serverUrl: string): Promise<ConformanceCheck[]> {
+    const checks: ConformanceCheck[] = [];
+    const timestamp = new Date().toISOString();
+
+    // First check: Is this a localhost URL?
+    if (!isLocalhostUrl(serverUrl)) {
+      // Return failure for both checks when server is not localhost
+      checks.push({
```
pcarleton, line 140: these checks.push calls are super repetitive. let's DRY them up

src/scenarios/server/dns-rebinding.ts
```diff
@@ -0,0 +1,219 @@
...
+      'Content-Type': 'application/json',
+      Host: hostHeader,
+      Accept: 'application/json, text/event-stream'
+    },
+    body: JSON.stringify({
+      jsonrpc: '2.0',
+      id: 1,
+      method: 'initialize',
+      params: {
+        protocolVersion: '2024-11-05',
```
maxisbey: why 2024-11-05?

src/scenarios/server/dns-rebinding.ts
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * Check if URL is a localhost URL
+ */
+function isLocalhostUrl(serverUrl: string): boolean {
+  const url = new URL(serverUrl);
+  const hostname = url.hostname.toLowerCase();
+  return (
+    hostname === 'localhost' ||
+    hostname === '127.0.0.1' ||
```
maxisbey, line 30: could just check first octet here? wasn't sure what realistic scenarios someone would use a different address, so asked claude and it came up with a few scenarios:

```
  - Running multiple services on the same port but different loopback IPs (e.g., 127.0.0.1:3000 and 127.0.0.2:3000)
  - Some container/networking tools use them for isolation
  - macOS firewall rules sometimes redirect to alternate loopback addresses
```

honestly not super fussed on this one, so can dismiss if you want

src/scenarios/server/dns-rebinding.ts
```diff
@@ -0,0 +1,219 @@
...
+
+    // First check: Is this a localhost URL?
+    if (!isLocalhostUrl(serverUrl)) {
+      const errorMessage =
+        'DNS rebinding tests require a localhost server URL (localhost, 127.0.0.1, or [::1])';
+      const details = { serverUrl, reason: 'non-localhost-url' };
+
+      checks.push({
+        ...rejectedCheckBase,
+        status: 'FAILURE' as const,
```
maxisbey: I don't think `as const` is needed here

src/scenarios/server/dns-rebinding.ts
```diff
@@ -0,0 +1,219 @@
...
+        errorMessage,
+        details
+      });
+      checks.push({
+        ...acceptedCheckBase,
+        status: 'FAILURE' as const,
```
maxisbey: I don't think `as const` is needed here

src/scenarios/server/dns-rebinding.ts
```diff
@@ -0,0 +1,219 @@
...
+    }
+
+    const validHost = getHostFromUrl(serverUrl);
+    const attackerHost = 'evil.example.com';
+
+    // Check 1: Invalid Host header should be rejected with a 4xx error
+    try {
+      const response = await sendRequestWithHost(serverUrl, attackerHost);
+      const isRejected =
+        response.statusCode >= 400 && response.statusCode < 500;
```
maxisbey, line 165: The spec says 403 specifically, although it talks about Origin not HOST: https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#security-warning

src/scenarios/server/dns-rebinding.ts
pcarleton: 🙈 thanks, should be 2025-11-25
```suggestion
        protocolVersion: '2025-11-25',
```

src/scenarios/server/dns-rebinding.ts
pcarleton: hmm yea i don't anticipate `.2` etc. as an issue, but good flag if it does come up

src/scenarios/server/dns-rebinding.ts
pcarleton: yea, good call out.  We should really move that to security best practices and clarify what it's protecting against.  Right now it's not clear what "validate" means, because if you have auth enabled, allowing any origin is fine, it's just without Auth that it causes problems.  And also checking the Host header is fine too as a protection against DNS rebinding.

Let me add that spec reference too for now, and I'll set Origin header as well as host in the test so that either protection will count.

src/scenarios/server/dns-rebinding.ts
```diff
@@ -144,73 +155,81 @@ See: https://github.com/modelcontextprotocol/typescript-sdk/security/advisories/
...
+    // Check 2: Valid localhost Host/Origin headers should be accepted (2xx response)
     try {
-      const response = await sendRequestWithHost(serverUrl, validHost);
+      const response = await sendRequestWithHostAndOrigin(serverUrl, validHost);
       const isAccepted =
         response.statusCode >= 200 && response.statusCode < 300;
 
       const details = {
         hostHeader: validHost,
+        originHeader: `http://${validHost}`,
```
maxisbey, line 208: Maybe we should actually be testing these separately? Otherwise we don't know whether the server will reject _both_ a bad host and a bad origin, we only know that it will reject either:

1. both
2. only header
3. only origin

but not which specifically

pcarleton: that's intentional, we want either protection to be valid. Checking host or origin will prevent the attack.

PR #139, opened 2026-02-09
feat: add token refresh and rotation conformance scenarios

jdmaturen:
## Summary

Adds two new client auth conformance scenarios that test OAuth 2.1 refresh token behavior:

- **`auth/token-refresh-basic`** — Tests that clients use the `refresh_token` grant to obtain a new access token when the current one expires (OAuth 2.1 §6). Server issues 2-second TTL access tokens + refresh token; client must detect 401, send `grant_type=refresh_token`, and use the new access token successfully.

- **`auth/token-refresh-rotation`** — Same flow but the server rotates the refresh token on each use (OAuth 2.1 §6.1). Client must store the new refresh token and not reuse the old one.

## Motivation

Multiple MCP client implementations have been reported as not handling token refresh reliably — leading to auth loops, death spirals, and excessive retry RPM when access tokens expire. These scenarios provide a concrete conformance check against the OAuth 2.1 refresh token spec.

## Changes

| File | Change |
|---|---|
| `token-refresh.ts` | New file: `TokenRefreshBasicScenario` and `TokenRefreshRotationScenario` |
| `createAuthServer.ts` | `issueRefreshToken`, `rotateRefreshTokens`, `accessTokenExpiresIn`, `onRefreshTokenRequest` options; full `grant_type=refresh_token` handler |
| `createServer.ts` | `perRequestServer` option (fresh MCP Server per request, needed because `server.close()` on response end breaks subsequent requests across token boundaries) |
| `mockTokenVerifier.ts` | Token expiration tracking (`issuedAt`, `expiresIn` per token); expired tokens throw `InvalidTokenError` with INFO conformance check |
| `spec-references.ts` | `OAUTH_2_1_REFRESH_TOKEN` (§6) and `OAUTH_2_1_TOKEN_ROTATION` (§6.1) |
| `index.ts` | Register both new scenarios |
| `everything-client.ts` | `runTokenRefreshClient` exercising the full refresh flow |

## Test plan

- [x] All 79 tests pass (77 existing + 2 new token refresh scenarios)
- [x] `auth/token-refresh-basic` passes in ~3s (2s token TTL + 1s buffer)
- [x] `auth/token-refresh-rotation` passes in ~3s
- [x] `tsc --noEmit` clean

## Dependencies

Depends on #138 (`InvalidTokenError` fix) — without it, expired tokens return 500 instead of 401 and the refresh flow never triggers.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/examples/clients/typescript/everything-client.ts
+++ b/examples/clients/typescript/everything-client.ts
@@ -371,6 +371,66 @@ export async function runPreRegistration(serverUrl: string): Promise<void> {
 registerScenario('auth/pre-registration', runPreRegistration);
 
 // ============================================================================
+// Token refresh scenarios
+// ============================================================================
+
+/**
+ * Token refresh client: authenticates, makes a request, waits for the
+ * short-lived access token to expire, then makes another request to
+ * trigger the refresh_token grant.
+ */
+async function runTokenRefreshClient(serverUrl: string): Promise<void> {
+  const client = new Client(
+    { name: 'test-token-refresh-client', version: '1.0.0' },
+    { capabilities: {} }
+  );
+
+  const oauthFetch = withOAuthRetry(
+    'test-token-refresh-client',
+    new URL(serverUrl),
+    handle401,
+    CIMD_CLIENT_METADATA_URL
+  )(fetch);
+
+  const transport = new StreamableHTTPClientTransport(new URL(serverUrl), {
+    fetch: oauthFetch
+  });
+
+  await client.connect(transport);
+  logger.debug('Token refresh: connected');
+
+  // First request — should succeed with initial access token
+  const tools = await client.listTools();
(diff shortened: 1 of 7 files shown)


pcarleton (changes requested):
This generally looks great, thanks for adding this.

My one big note is let's just do the rotation example.  I don't see a ton of value of a client that passes the first one but not the second one. Progress on making the client better can be gained by passing more checks out of the single scenario (e.g. pass the first set of checks, fail the 2nd one when they re-use the old token).

Reasons for fewer tests:
* less code to maintain & update
* these suites will often run in CI, in some cases on every push for every sdk. each test often involves spinning up its own http server which takes time.  So in general we want to keep it fast.

On the 2nd bullet, we may add a "slow" test tag or something in the future to indicate that we shouldn't run it on CI, and only run occasionally or when doing certification.

pcarleton:
Closing as part of a maintainer backlog sweep

PR #165, opened 2026-02-24
fix: tools-call-simple-text now fails when tool returns isError: true

lux999:
Added an `isError` check to `ToolsCallSimpleTextScenario`, so the test `tools-call-simple-text` correctly fails when the tool returns an error result instead of a successful text response.

## Motivation and Context
The test `tools-call-simple-text` was passing for servers that don't implement `test_simple_text` at all.
When an unknown tool is called, MCP frameworks return `{ isError: true, content: [{ type: "text", text: "Tool not found: ..." }] }`. The existing assertions only checked for a non-empty text content item, which the "Tool not found" error message satisfies. The missing assertion was that `result.isError` must not be true.

## How Has This Been Tested?
Verified against the existing everything-server reference implementation, which correctly implements `test_simple_text` and continues to pass. Confirmed that servers without `test_simple_text` now correctly fail the test.

## Breaking Changes
Servers that were previously passing `tools-call-simple-text` without implementing `test_simple_text` will now correctly fail.

## Types of changes

- [ ] Documentation update

## Checklist

### DIFF
--- a/src/scenarios/server/tools.ts
+++ b/src/scenarios/server/tools.ts
@@ -127,6 +127,8 @@ Implement tool \`test_simple_text\` with no arguments that returns:
 
       // Validate response
       const errors: string[] = [];
+      if ((result as any).isError === true)
+        errors.push('Tool returned an error instead of content');
       const content = (result as any).content;
       if (!content) errors.push('Missing content array');
       if (!Array.isArray(content)) errors.push('content is not an array');
--- a/src/scenarios/server/tools.ts
+++ b/src/scenarios/server/tools.ts
@@ -121,8 +121,14 @@ Implement tool \`test_simple_text\` with no arguments that returns:
       const connection = await connectToServer(serverUrl);
 
       const result = await connection.client.callTool({
-        name: 'test_simple_text'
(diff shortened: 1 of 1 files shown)


pcarleton (dismissed):
LGTM, thanks!

lux999:
Hi @pcarleton, thanks for the approval!

Quick heads-up on the new commit: after merging main, CI caught one test failure. This scenario originally omitted `arguments` since they're not required for `test_simple_text`, but the upstream SDK server rejects `undefined` arguments against any inputSchema and returns `isError: true` - which the new check correctly flags. There's an upstream fix in [typescript-sdk#1404](https://github.com/modelcontextprotocol/typescript-sdk/pull/1404) but it wasn't backported to 1.x - see [typescript-sdk#1869](https://github.com/modelcontextprotocol/typescript-sdk/issues/1869).

I squashed a small workaround into the fix commit: pass `arguments: {}` explicitly, with an inline comment linking to the SDK issue. Matches the convention in the sibling tool-call scenarios. Happy to revisit if you'd prefer a different approach.

Thanks again!

PR #166, opened 2026-02-24
feat (auth): add sep-2207 client checks

wdawson:
Add SEP-2207 (OIDC Refresh Token Guidance) conformance tests as a draft suite   

## Motivation and Context

[SEP-2207](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2207) adds normative guidance for MCP clients around `offline_access` scope and `refresh_token` grant type handling when interacting with OIDC authorization servers. This PR adds conformance tests for the client-side requirements, so SDK authors can validate their implementations against the draft spec.                                                                                            

Server-side guidance (SHOULD NOT include `offline_access` in PRM/WWW-Authenticate) depends on #105 and #155 

## How Has This Been Tested?

- All tests pass locally
- Draft scenarios tested against the TypeScript everything-client (3 scenarios, 32 checks, 0 failures)
- Draft scenarios tested against the Python SDK conformance client (`modelcontextprotocol/python-sdk`)
- Verified draft scenarios are excluded from tier scoring via the existing `isTierScoring()` filtering
- Verified spec-version isolation: draft scenarios only appear under `specVersions: ['draft']`
- Negative path: confirmed clients that don't implement SEP-2207 get INFO/WARNING (not FAILURE) for MAY/SHOULD checks, and FAILURE for MUST NOT violations

## Breaking Changes

None. Draft scenarios are non-scoring and only run when explicitly requested via `--suite draft` or `--spec-version draft`.

## Types of changes

- [ ] Documentation update

## Checklist

## Additional context

**New scenarios:**

| Scenario | Checks | Description |
|---|---|---|
| `auth/offline-access-scope` | `sep-2207-client-metadata-grant-types`, `sep-2207-offline-access-requested` | AS advertises `offline_access` in `scopes_supported`. Verifies client includes `refresh_token` in grant_types (SHOULD, via DCR or CIMD) and requests `offline_access` scope (MAY). |
| `auth/offline-access-not-supported` | `sep-2207-offline-access-not-requested` | AS does not advertise `offline_access`. Verifies client does not request it (MUST NOT). |

**Also moved** `auth/resource-mismatch` from `authScenariosList` to `draftScenariosList` — it already had `specVersions: ['draft']` but was incorrectly in the tier-required list.

**Design decisions:**

- CIMD `grant_types` check is attempted in `stop()` (async) by fetching the client_id URL. Falls back to INFO if URL is unreachable (e.g., fake CIMD URLs in test environments).
- Handles DCR, CIMD, and pre-registered client paths gracefully — checks what it can, emits INFO for what it can't verify.
- Requesting `offline_access` when AS doesn't support it is FAILURE (unsupported scope), not WARNING.

**AI Disclosure:** AI assisted with implementation and test authoring. All changes were reviewed and guided by hand.

### DIFF
--- a/src/scenarios/client/auth/offline-access.ts
+++ b/src/scenarios/client/auth/offline-access.ts
@@ -25,7 +25,7 @@ export class OfflineAccessScopeScenario implements Scenario {
   name = 'auth/offline-access-scope';
   specVersions: SpecVersion[] = ['draft'];
   description =
-    'Tests that client handles offline_access scope and refresh_token grant type when AS supports them (SEP-2207)';
+    'Tests that a client that wants a refresh token handles offline_access scope and refresh_token grant type when AS supports them (SEP-2207)';
 
   private authServer = new ServerLifecycle();
   private server = new ServerLifecycle();


src/scenarios/client/auth/offline-access.ts
```diff
@@ -0,0 +1,308 @@
...
+ * Setup:
+ * - AS metadata: scopes_supported includes 'offline_access'
+ * - PRM: scopes_supported does NOT include 'offline_access' (per SEP-2207 server guidance)
+ * - Both CIMD and DCR paths available
+ */
+export class OfflineAccessScopeScenario implements Scenario {
+  name = 'auth/offline-access-scope';
+  specVersions: SpecVersion[] = ['draft'];
+  description =
+    'Tests that client handles offline_access scope and refresh_token grant type when AS supports them (SEP-2207)';
```
pcarleton:
```suggestion
    'Tests that a client that wants a refresh token handles offline_access scope and refresh_token grant type when AS supports them (SEP-2207)';
```

I want to be clear that this is only when a client wants to handle refresh tokens. As in, I want it to be clear that it is not required to set that grant_type for a client to be conformant

Issue #169, opened 2026-02-25
Add conformance test for authorization server metadata


Michito-Okai:
# Background
This repository contains conformance tests for MCP clients and MCP servers. However, there are currently no conformance tests for authorization servers.
It may be because conformance tests for authorization servers may have low priority, but they are necessary.

# First Step
Implementing conformance tests for authorization servers involves multiple steps. Therefore, as a first step, this issue focuses on implementing a test for OAuth 2.0 Authorization Server Metadata [RFC8414](https://datatracker.ietf.org/doc/html/rfc8414).

Michito-Okai:
@pcarleton 
Thanks for your review.

I have fixed [this comment](https://github.com/modelcontextprotocol/conformance/issues/169#issuecomment-4010709641) as follows. Please review it. 

(Before) 1. A new `authorization` CLI command that accepts `--url <metadata-endpoint-url>` (e.g., `--url https://example.com/.well-known/oauth-authorization-server`)

(After) 1. A new `authorization` CLI command that accepts `--url <authorization-server-issuer-url>` (e.g., `--url https://auth.example.com`)

Michito-Okai:
@pcarleton Since [PR#170](https://github.com/modelcontextprotocol/conformance/pull/170) has been merged, is it okay to close this issue?

PR #170, opened 2026-02-25
feat: add conformance test for authorization server metadata

Michito-Okai:
## Motivation and Context
Described in #169

## How Has This Been Tested?
- npm test (Test Files  6 passed, Tests  85 passed)
- npm test -- src/scenarios/authorization-server/authorization-server-metadata.test.ts (Test Files  1 passed, Tests  1 passed)
- npm run start -- authorization --url http://localhost:8080/.well-known/oauth-authorization-server/realms/mp
  --url URL of the authorization server metadata endpoint
- npm run start -- authorization --url http://localhost:8080/.well-known/oauth-authorization-server/realms/invalid
  --url invalid URL
- npm run start -- list

## Breaking Changes
- add authorization server mode (-- authorization)
- add url option for authorization server mode (--url <authorization server metadata endpoint>)
- add a conformance test for authorization server metadata

## Types of changes

## Checklist

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -11,6 +11,10 @@ import {
   runInteractiveMode
 } from './runner';
 import {
+  printAuthorizationServerSummary,
+  runAuthorizationServerConformanceTest
+} from './runner/authorization-server';
+import {
   listScenarios,
   listClientScenarios,
   listActiveClientScenarios,
@@ -24,11 +28,17 @@ import {
   listScenariosForSpec,
   listClientScenariosForSpec,
   getScenarioSpecVersions,
+  listClientScenariosForAuthorizationServer,
+  listClientScenariosForAuthorizationServerForSpec,
   resolveSpecVersion
 } from './scenarios';
 import type { SpecVersion } from './scenarios';
 import { ConformanceCheck } from './types';
-import { ClientOptionsSchema, ServerOptionsSchema } from './schemas';
+import {
+  AuthorizationServerOptionsSchema,
+  ClientOptionsSchema,
+  ServerOptionsSchema
+} from './schemas';
 import {
   loadExpectedFailures,
   evaluateBaseline,
@@ -44,12 +54,19 @@ import packageJson from '../package.json';
 function filterScenariosBySpecVersion(
   allScenarios: string[],
   version: SpecVersion,
-  command: 'client' | 'server'
+  command: 'client' | 'server' | 'authorization'
 ): string[] {
-  const versionScenarios =
-    command === 'client'
-      ? listScenariosForSpec(version)
-      : listClientScenariosForSpec(version);
+  let versionScenarios: string[];
+  if (command === 'client') {
+    versionScenarios = listScenariosForSpec(version);
+  } else if (command === 'server') {
+    versionScenarios = listClientScenariosForSpec(version);
+  } else if (command === 'authorization') {
+    versionScenarios =
+      listClientScenariosForAuthorizationServerForSpec(version);
+  } else {
+    versionScenarios = [];
+  }
   const allowed = new Set(versionScenarios);
   return allScenarios.filter((s) => allowed.has(s));
 }
@@ -437,6 +454,87 @@ program
     }
   });
 
+// Authorization command - tests an authorization server implementation
+program
+  .command('authorization')
+  .description(
+    'Run conformance tests against an authorization server implementation'
+  )
+  .requiredOption('--url <url>', 'URL of the authorization server issuer')
+  .option('-o, --output-dir <path>', 'Save results to this directory')
+  .option(
+    '--spec-version <version>',
+    'Filter scenarios by spec version (cumulative for date versions)'
+  )
+  .action(async (options) => {
+    try {
+      // Validate options with Zod
+      const validated = AuthorizationServerOptionsSchema.parse(options);
+      const outputDir = options.outputDir;
+      const specVersionFilter = options.specVersion
+        ? resolveSpecVersion(options.specVersion)
+        : undefined;
+
+      let scenarios: string[];
+      scenarios = listClientScenariosForAuthorizationServer();
+      if (specVersionFilter) {
+        scenarios = filterScenariosBySpecVersion(
+          scenarios,
+          specVersionFilter,
+          'authorization'
+        );
+      }
+      console.log(
+        `Running test (${scenarios.length} scenarios) against ${validated.url}\n`
+      );
+
+      const allResults: { scenario: string; checks: ConformanceCheck[] }[] = [];
+      for (const scenarioName of scenarios) {
+        console.log(`\n=== Running scenario: ${scenarioName} ===`);
+        try {
+          const result = await runAuthorizationServerConformanceTest(
+            validated.url,
+            scenarioName,
+            outputDir
+          );
+          allResults.push({ scenario: scenarioName, checks: result.checks });
+        } catch (error) {
+          console.error(`Failed to run scenario ${scenarioName}:`, error);
+          allResults.push({
+            scenario: scenarioName,
+            checks: [
+              {
+                id: scenarioName,
+                name: scenarioName,
+                description: 'Failed to run scenario',
+                status: 'FAILURE',
+                timestamp: new Date().toISOString(),
+                errorMessage:
+                  error instanceof Error ? error.message : String(error)
+              }
+            ]
+          });
+        }
+      }
+      const { totalFailed } = printAuthorizationServerSummary(allResults);
+      process.exit(totalFailed > 0 ? 1 : 0);
+    } catch (error) {
+      if (error instanceof ZodError) {
+        console.error('Validation error:');
+        error.issues.forEach((err) => {
+          console.error(`  ${err.path.join('.')}: ${err.message}`);
+        });
+        console.error('\nAvailable authorization server scenarios:');
+        listClientScenariosForAuthorizationServer().forEach((s) =>
+          console.error(`  - ${s}`)
+        );
+        process.exit(1);
+      }
+      console.error('Authorization server test error:', error);
+      process.exit(1);
+    }
+  });
+
 // Tier check command
 program.addCommand(createTierCheckCommand());
 
@@ -446,6 +544,7 @@ program
   .description('List available test scenarios')
   .option('--client', 'List client scenarios')
   .option('--server', 'List server scenarios')
+  .option('--authorization', 'List authorization server scenarios')
   .option(
     '--spec-version <version>',
     'Filter scenarios by spec version (cumulative for date versions)'
@@ -455,7 +554,10 @@ program
       ? resolveSpecVersion(options.specVersion)
       : undefined;
 
-    if (options.server || (!options.client && !options.server)) {
+    if (
+      options.server ||
+      (!options.client && !options.server && !options.authorization)
+    ) {
       console.log('Server scenarios (test against a server):');
       let serverScenarios = listClientScenarios();
       if (specVersionFilter) {
@@ -471,7 +573,10 @@ program
       });
     }
 
-    if (options.client || (!options.client && !options.server)) {
+    if (
+      options.client ||
+      (!options.client && !options.server && !options.authorization)
+    ) {
       if (options.server || (!options.client && !options.server)) {
         console.log('');
       }
@@ -485,6 +590,31 @@ program
         );
       }
       clientScenarioNames.forEach((s) => {
+        const v = getScenarioSpecVersions(s);
+        console.log(`  - ${s}${v ? ` [${v}]` : ''}`);
+      });
+    }
+
+    if (
+      options.authorization ||
+      (!options.authorization && !options.server && !options.client)
+    ) {
+      if (!(options.authorization && !options.server && !options.client)) {
+        console.log('');
+      }
+      console.log(
+        'Authorization server scenarios (test against an authorization server):'
+      );
+      let authorizationServerScenarios =
+        listClientScenariosForAuthorizationServer();
+      if (specVersionFilter) {
+        authorizationServerScenarios = filterScenariosBySpecVersion(
+          authorizationServerScenarios,
+          specVersionFilter,
+          'authorization'
+        );
+      }
+      authorizationServerScenarios.forEach((s) => {
         const v = getScenarioSpecVersions(s);
         console.log(`  - ${s}${v ? ` [${v}]` : ''}`);
       });
--- a/src/scenarios/authorization-server/authorization-server-metadata.ts
+++ b/src/scenarios/authorization-server/authorization-server-metadata.ts
@@ -1,0 +1,155 @@
+/**
+ * Authorization server metadata endpoint test scenarios for MCP authorization servers
+ */
+import {
+  ClientScenarioForAuthorizationServer,
+  ConformanceCheck,
+  SpecVersion
+} from '../../types';
+import { request } from 'undici';
+
+type Status = 'SUCCESS' | 'FAILURE';
+
+export class AuthorizationServerMetadataEndpointScenario implements ClientScenarioForAuthorizationServer {
+  name = 'authorization-server-metadata-endpoint';
+  specVersions: SpecVersion[] = ['2025-03-26', '2025-06-18', '2025-11-25'];
+  description = `Test authorization server metadata endpoint.
+
+**Authorization Server Implementation Requirements:**
+
+**Endpoint**: \`authorization server metadata\`
+
+**Requirements**:
+- HTTP response status code MUST be 200 OK
+- Content-Type header MUST be application/json
+- Return a JSON response including issuer, authorization_endpoint, token_endpoint and response_types_supported
+- The issuer value MUST match the URI obtained by removing the well-known URI string from the authorization server metadata URI.`;
+
+  async run(serverUrl: string): Promise<ConformanceCheck[]> {
+    let status: Status = 'SUCCESS';
+    let errorMessage: string | undefined;
+    let details: any;
+    let response: any | null = null;
+    try {
+      const wellKnownUrls = this.createWellKnownUrl(serverUrl);
+
+      for (const url of wellKnownUrls) {
+        try {
+          const checkResponse = await request(url, { method: 'GET' });
+          if (checkResponse.statusCode === 200) {
+            response = checkResponse;
+            break;
+          }
+        } catch {
+          // Ignore the error and proceed to the next loop.
+        }
+      }
+
+      if (!response) {
+        throw new Error(
+          'All authorization server metadata endpoints returned invalid status code.'
+        );
+      }
+
+      this.validateContentType(response.headers['content-type']);
+
+      const body = await this.parseJson(response);
+      const errors: string[] = [];
+      this.validateMetadataBody(body, serverUrl, errors);
+
+      if (errors.length > 0) {
+        status = 'FAILURE';
+        errorMessage = errors.join(', ');
+      }
+
+      details = {
+        contentType: response.headers['content-type'],
+        body
+      };
+    } catch (error) {
+      status = 'FAILURE';
+      errorMessage = error instanceof Error ? error.message : String(error);
+    }
+
+    return [
+      {
+        id: 'authorization-server-metadata',
+        name: 'AuthorizationServerMetadata',
+        description: 'Valid authorization server metadata response',
+        status,
+        timestamp: new Date().toISOString(),
+        errorMessage,
+        specReferences: [
+          {
+            id: 'Authorization-Server-Metadata',
+            url: 'https://datatracker.ietf.org/doc/html/rfc8414'
+          }
+        ],
+        ...(details ? { details } : {})
+      }
+    ];
+  }
+
+  private createWellKnownUrl(serverUrl: string): string[] {
+    const base = new URL(serverUrl);
+    const origin = base.origin;
+    const path = base.pathname.replace(/\/$/, '');
+
+    const urls = new Set<string>();
+    urls.add(`${origin}/.well-known/oauth-authorization-server${path}`);
+    urls.add(`${origin}/.well-known/openid-configuration${path}`);
+    urls.add(`${origin}${path}/.well-known/openid-configuration`);
+
+    return Array.from(urls);
+  }
+
+  private validateContentType(contentType?: string | string[]): void {
+    const valid =
+      typeof contentType === 'string' &&
+      contentType.toLowerCase().includes('application/json');
+
+    if (!valid) {
+      throw new Error(`Invalid Content-Type: ${contentType ?? '(missing)'}`);
+    }
+  }
+
+  private async parseJson(response: any): Promise<Record<string, any>> {
+    const body = await response.body.json();
+    if (typeof body !== 'object' || body === null) {
+      throw new Error('Response body is not an object');
+    }
+    return body;
+  }
+
+  private validateMetadataBody(
+    body: Record<string, any>,
+    serverUrl: string,
+    errors: string[]
+  ): void {
+    this.assertString(
+      body.authorization_endpoint,
+      'authorization_endpoint',
+      errors
+    );
+    this.assertString(body.token_endpoint, 'token_endpoint', errors);
+
+    if (
+      !Array.isArray(body.response_types_supported) ||
+      !body.response_types_supported.includes('code')
+    ) {
+      errors.push(
+        'Response body does not include valid "response_types_supported" claim'
+      );
+    }
+
+    if (body.issuer !== serverUrl) {
+      errors.push(`Invalid issuer: ${body.issuer ?? '(missing)'}`);
+    }
+  }
+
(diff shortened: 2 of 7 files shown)


src/scenarios/authorization-server/authorization-server-metadata.ts
```diff
@@ -0,0 +1,141 @@
...
+} from '../../types';
+import { request } from 'undici';
+
+type Status = 'SUCCESS' | 'FAILURE';
+
+export class AuthorizationServerMetadataEndpointScenario
+  implements ClientScenarioForAuthorizationServer
+{
+  name = 'authorization-server-metadata-endpoint';
+  specVersions: SpecVersion[] = ['2025-03-26', '2025-06-18', '2025-11-25'];
```
tnorimat, line 15: To avoid duplication, instead of defining spec version here, could you import `SpecVersion` of `types.ts` like 
```
import {
  ClientScenarioForAuthorizationServer,
  ConformanceCheck,
  SpecVersion 
} from '../../types';
``` 
?

src/scenarios/index.ts
```diff
@@ -178,6 +196,12 @@ export function getClientScenario(name: string): ClientScenario | undefined {
   return clientScenarios.get(name);
 }
 
+export function getClientScenarioForAuthorizationServer(
+  name: string
+): ClientScenarioForAuthorizationServer | undefined {
+  return clientScenariosForAuthorizationServer.get(name);
```
tnorimat, line 209: `getClientScenarioForAuthorizationServer` declares its return type as `ClientScenarioForAuthorizationServer` but that type isn't imported. Could you add it to the import from '../types' on line 1 ?

src/index.ts
```diff
@@ -52,12 +61,19 @@ function resolveSpecVersion(value: string): SpecVersion {
...
-      ? listScenariosForSpec(version)
-      : listClientScenariosForSpec(version);
+  let versionScenarios: string[];
+  if (command === 'client') {
+    versionScenarios = listScenariosForSpec(version);
+  } else if (command === 'server') {
+    versionScenarios = listClientScenariosForSpec(version);
+  } else if (command === 'authorization') {
+    versionScenarios =
+      listClientScenariosForAuthorizationServerForSpec(version);
```
tnorimat, line 66: The function is called inside `filterScenariosBySpecVersion` but was never imported from `'./scenarios'`. It's exported from `index.ts` but missing from the import block at line18-34 of `index.ts`. Could you import that ? like

```
import {
  listScenarios,
  listClientScenarios,
  listActiveClientScenarios,
  listPendingClientScenarios,
  listAuthScenarios,
  listMetadataScenarios,
  listCoreScenarios,
  listExtensionScenarios,
  listBackcompatScenarios,
  listScenariosForSpec,
  listClientScenariosForSpec,
  getScenarioSpecVersions,
  listClientScenariosForAuthorizationServer,
  listClientScenariosForAuthorizationServerForSpec,
  ALL_SPEC_VERSIONS
} from './scenarios';
```

src/index.ts
```diff
@@ -496,6 +598,31 @@ program
...
+      console.log(
+        'Authorization server scenarios (test against a authorization server):'
+      );
+      let authorizationServerScenarios =
+        listClientScenariosForAuthorizationServer();
+      if (specVersionFilter) {
+        authorizationServerScenarios = filterScenariosBySpecVersion(
+          authorizationServerScenarios,
+          specVersionFilter,
+          'client'
```
tnorimat: `authorization` instead of `client` ?

src/index.ts
```diff
@@ -444,6 +460,85 @@ program
...
+      const { totalFailed } = printAuthorizationServerSummary(allResults);
+      process.exit(totalFailed > 0 ? 1 : 0);
+    } catch (error) {
+      if (error instanceof ZodError) {
+        console.error('Validation error:');
+        error.errors.forEach((err) => {
+          console.error(`  ${err.path.join('.')}: ${err.message}`);
+        });
+        console.error('\nAvailable authorization server scenarios:');
+        listClientScenarios().forEach((s) => console.error(`  - ${s}`));
```
tnorimat: `listClientScenariosForAuthorizationServer` instead of `listClientScenarios` ?

src/scenarios/index.ts
tnorimat: Is it needed to add authorizationserverScenarios ? If not, when `list --authorization` prints spec versions per scenario, they'll show as undefined.

src/index.ts
```diff
@@ -444,6 +460,85 @@ program
     }
   });
 
+// Authorization command - tests a authorization server implementation
```
tnorimat: `an authorization server` instead of  `a authorization server` ?

src/index.ts
```diff
@@ -496,6 +598,31 @@ program
...
+
+    if (
+      options.authorization ||
+      (!options.authorization && !options.server && !options.client)
+    ) {
+      if (!(options.authorization && !options.server && !options.client)) {
+        console.log('');
+      }
+      console.log(
+        'Authorization server scenarios (test against a authorization server):'
```
tnorimat: `an authorization server` instead of `a authorization server` ?

src/types.ts
```diff
@@ -60,3 +60,10 @@ export interface ClientScenario {
   specVersions: SpecVersion[];
   run(serverUrl: string): Promise<ConformanceCheck[]>;
 }
+
+export interface ClientScenarioForAuthorizationServer {
```
tnorimat, line 64: Both interfaces `ClientScenario and ClientScenarioForAuthorizationServer` have the exact same shape. Consider using a type alias (`type ClientScenarioForAuthorizationServer = ClientScenario`) or just reusing `ClientScenario` directly?.

src/scenarios/authorization-server/authorization-server-metadata.ts
```diff
@@ -0,0 +1,141 @@
...
+- Content-Type header MUST be application/json
+- Return a JSON response including issuer, authorization_endpoint, token_endpoint and response_types_supported
+- The issuer value MUST match the URI obtained by removing the well-known URI string from the authorization server metadata URI.`;
+
+  async run(serverUrl: string): Promise<ConformanceCheck[]> {
+    let status: Status = 'SUCCESS';
+    let errorMessage: string | undefined;
+    let details: any;
+
+    const fail = (msg: string) => {
```
tnorimat: The `fail` is never called.
All validation methods throw instead, and the catch block sets `status` and `errorMessage` directly.

src/scenarios/authorization-server/authorization-server-metadata.ts
```diff
@@ -0,0 +1,141 @@
...
+    if (
+      !Array.isArray(body.response_types_supported) ||
+      body.response_types_supported.length === 0
+    ) {
+      throw new Error(
+        'Response body does not include valid "response_types_supported" claim'
+      );
+    }
+
+    const expectedIssuer = serverUrl.replace(
```
tnorimat: It only removes the first occurrence. If the URL were e.g. `https://example.com/.well-known/oauth-authorization-server/tenant`, the replace would produce `https://example.com//tenant` (note the path portion is kept). Per RFC 8414 §3, the issuer for a path-aware metadata URL `/.well-known/oauth-authorization-server/<path>` should be `https://example.com/<path>`. The current approach works correctly for the root case but may be incorrect for path-based issuers.

src/index.ts
```diff
@@ -462,7 +558,10 @@ program
       ? resolveSpecVersion(options.specVersion)
       : undefined;
 
-    if (options.server || (!options.client && !options.server)) {
+    if (
+      options.server ||
+      (!options.client && !options.server && !options.authorization)
```
tnorimat, line 559: The `server` command supports running a single scenario (`--scenario`), suite selection (`suite`), expected-failures baselines, and verbose output (--verbose). The new `authorization` command has none of these, making it less flexible. This may be intentional for an initial implementation but is worth noting. It is up to you to consider this point by this PR.

tnorimat (commented):
@Michito-Okai Hello, I added some review comments. Could you check them?

src/scenarios/index.ts
tnorimat: For example,
```
export function getScenarioSpecVersions(
  name: string
): SpecVersion[] | undefined {
  return (
    scenarios.get(name)?.specVersions ??
    clientScenarios.get(name)?.specVersions ??
    clientScenariosForAuthorizationServer.get(name)?.specVersions
  );
}
```

src/scenarios/authorization-server/authorization-server-metadata.ts
tnorimat: For example,
```
    } catch (error) {
      status = 'FAILURE';
      errorMessage = error instanceof Error ? error.message : String(error);
    }
```

src/scenarios/authorization-server/authorization-server-metadata.ts
tnorimat: For example,
```
const url = new URL(serverUrl);
const wellKnownPrefix = '/.well-known/oauth-authorization-server';
const suffix = url.pathname.startsWith(wellKnownPrefix + '/')
  ? url.pathname.slice(wellKnownPrefix.length)
  : '';
const expectedIssuer = `${url.origin}${suffix}`;
```

src/index.ts
Michito-Okai: I fixed. Could you check this?

src/scenarios/authorization-server/authorization-server-metadata.ts
Michito-Okai: I fixed. Could you check this?

src/scenarios/index.ts
Michito-Okai: I fixed. Could you check this?

src/index.ts
Michito-Okai: I fixed. Could you check this?

src/index.ts
Michito-Okai: I fixed. Could you check this?

src/index.ts
Michito-Okai: I fixed. Could you check this?

src/scenarios/index.ts
Michito-Okai: I fixed. Could you check this?

src/index.ts
Michito-Okai: I fixed. Could you check this?

src/index.ts
Michito-Okai: I fixed. Could you check this?

src/types.ts
Michito-Okai: The shape of ClientScenarioForAuthorizationServer is likely to change with enhancements, so I defined ClientScenarioForAuthorizationServer.

src/scenarios/authorization-server/authorization-server-metadata.ts
Michito-Okai: I fixed. Could you check this?

src/scenarios/authorization-server/authorization-server-metadata.ts
Michito-Okai: I think that if the URL were e.g. https://example.com/.well-known/oauth-authorization-server/tenant, the replace would produce not https://example.com//tenant but https://example.com/tenant. Therefore I think that the current approach works correctly for path-based issuers.

src/index.ts
Michito-Okai: The scenario, suite, expected-failures and verbose options are outside the scope of this PR. I am considering to add these options in a future enhancement. How does that sound?

src/scenarios/authorization-server/authorization-server-metadata.ts
tnorimat: To avoid duplication, could you remove ['2025-03-26', '2025-06-18', '2025-11-25'] ?

src/scenarios/authorization-server/authorization-server-metadata.ts
tnorimat: Thank you for the clarification.

src/scenarios/index.ts
tnorimat: Thank you.

src/index.ts
tnorimat: Thank you.

src/index.ts
tnorimat: Yes I understand.

src/types.ts
tnorimat: Thank you for the clarification.

src/scenarios/index.ts
tnorimat: Thank you.

src/index.ts
tnorimat: Thank you.

src/index.ts
tnorimat: Thank you.

src/scenarios/authorization-server/authorization-server-metadata.ts
tnorimat: Thank you.

src/index.ts
tnorimat: I confirmed it.

src/scenarios/authorization-server/authorization-server-metadata.ts
```diff
@@ -0,0 +1,138 @@
...
+      throw new Error('Response body is not an object');
+    }
+    return body;
+  }
+
+  private validateMetadataBody(
+    body: Record<string, any>,
+    serverUrl: string
+  ): void {
+    this.assertString(body.authorization_endpoint, 'authorization_endpoint');
```
pcarleton: I like the `throw / catch` pattern for issues that block further progress (e.g. a non-200 response), but for issues where there may be multiple distinct issues, I'd prefer we either add multiple failing checks to the list of checks returned OR include all validation failures in one check so that the user gets feedback on the maximum possible list of failures to check when they get a test failure.

e.g. I want to avoid this situation:
* Run test, fails w/ missing `authorization_endpoint`
* Run test again, fails w/ missing `token_endpoint`
* Run test again, fails with missing `response_types_supported`

And instead have:
* Run test, fails w/ missing `authorization_endpoint`, `token_endpoint`, `response_types_supported`

src/scenarios/authorization-server/authorization-server-metadata.ts
```diff
@@ -0,0 +1,138 @@
...
+
+**Endpoint**: \`authorization server metadata\`
+
+**Requirements**:
+- HTTP response status code MUST be 200 OK
+- Content-Type header MUST be application/json
+- Return a JSON response including issuer, authorization_endpoint, token_endpoint and response_types_supported
+- The issuer value MUST match the URI obtained by removing the well-known URI string from the authorization server metadata URI.`;
+
+  async run(serverUrl: string): Promise<ConformanceCheck[]> {
```
pcarleton, line 28: mentioned in the issue, but I think we want serverUrl to be the issuerUrl, and we test that the metadata is at one of the .well-known paths.

pcarleton (changes requested):

src/scenarios/authorization-server/authorization-server-metadata.ts
Michito-Okai: `SpecVersion` of `types.ts` is a type declaration. Therefore, these are not duplicates.

src/scenarios/authorization-server/authorization-server-metadata.ts
Michito-Okai: @pcarleton 
I fixed as follows. Could you check this?
When test fails in `validateMetadataBody`, failures are not handled with a `throw / catch` pattern, but instead all checks are consolidated and verified within a single test run.

src/scenarios/authorization-server/authorization-server-metadata.ts
Michito-Okai: @pcarleton 
I fixed as follows. Could you check this?
`run` receives argument as the authorization server issuer.
`createWellKnownUrl` create the authorization server metadata url based the authorization server issuer

pcarleton: This looks good for OAuth metadata, but I was assuming we would also check OIDC metadata routes.

Basically I would imagine we wouldn't want a server to marked as failing if it's using OIDC metadata, since we treat that as valid.

I believe that would require us probing URL paths.

tnorimat: @Michito-Okai Hello, as we discussed, to incorporate @pcarleton comments, how about the following by following the MCP 2025-11-25 spec https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery ?

- The test receives an issuer(server) url (mandatory)  including or not including a tenant/realm part.
- If the tenant/realm part exist, creating an well-know url and access the url by the following the order that the MCP 2025-11-25 spec described.

```
1. OAuth 2.0 Authorization Server Metadata with path insertion: https://auth.example.com/.well-known/oauth-authorization-server/tenant1
2. OpenID Connect Discovery 1.0 with path insertion: https://auth.example.com/.well-known/openid-configuration/tenant1
3. OpenID Connect Discovery 1.0 path appending: https://auth.example.com/tenant1/.well-known/openid-configuration
```
If the test can successfully fetches the metadata from one of them, the test proceeds (no need to do all the three patterns above). If the test cannot fetch the metadata from any of them, the test returns a failure.

- If the tenant/realm part does not exist, creating an well-know url and access the url by the following the order that the MCP 2025-11-25 spec described.

```
1. OAuth 2.0 Authorization Server Metadata: https://auth.example.com/.well-known/oauth-authorization-server
2. OpenID Connect Discovery 1.0: https://auth.example.com/.well-known/openid-configuration
```
If the test can successfully fetches the metadata from one of them, the test proceeds (no need to do all the two patters above). If the test cannot fetch the metadata from any of them, the test returns a failure.

Michito-Okai: @pcarleton @tnorimat 
Thank you for your comments. I see. I added OIDC metadata routes.
If the test cannot fetch the metadata from OAuth metadata route, the test fetches from OIDC metadata routes.

tnorimat (commented):
@Michito-Okai Hello, I reviewed the PR and it seems good to me. I think that the revised PR incorporates @pcarleton the last comment.

pcarleton (dismissed):
This looks great, thank you!

src/scenarios/authorization-server/authorization-server-metadata.ts
```diff
@@ -0,0 +1,157 @@
...
+    this.assertString(
+      body.authorization_endpoint,
+      'authorization_endpoint',
+      errors
+    );
+    this.assertString(body.token_endpoint, 'token_endpoint', errors);
+
+    if (
+      !Array.isArray(body.response_types_supported) ||
+      body.response_types_supported.length === 0
```
nbarbettini: We should also require that `response_types_supported` include `code`. The AS may also support other response types (that's okay), but MCP requires the auth code flow + PKCE which should be `code` in OAuth 2.1.

src/scenarios/authorization-server/authorization-server-metadata.ts
```diff
@@ -0,0 +1,157 @@
...
+    ) {
+      errors.push(
+        'Response body does not include valid "response_types_supported" claim'
+      );
+    }
+
+    if (body.issuer !== serverUrl) {
+      errors.push(`Invalid issuer: ${body.issuer ?? '(missing)'}`);
+    }
```
nbarbettini, line 147: Exact issuer equality is correct, but we do need to be careful about path normalization before this check.

[RFC 8414 §3.1](https://www.rfc-editor.org/rfc/rfc8414.html#section-3.1) says that if the issuer identifier contains a path component, any trailing slash MUST be removed before building the metadata URL. 

In other words, we should not assume that `serverUrl` is already normalized; we should normalize it earlier in this scenario and then this exact-issuer check will be correct in all cases.

src/scenarios/authorization-server/authorization-server-metadata.ts
Michito-Okai: @nbarbettini 
Thank you for your review. I modified the check from verifying that the array has one or more elements to verifying whether the array contains `code`.

src/scenarios/authorization-server/authorization-server-metadata.ts
Michito-Okai: @nbarbettini 
> [RFC 8414 §3.1](https://www.rfc-editor.org/rfc/rfc8414.html#section-3.1) says that if the issuer identifier contains a path component, any trailing slash MUST be removed before building the metadata URL.

The removal of the trailing slash from the issuer (`serverUrl`) when constructing the metadata URL is performed in [L98](https://github.com/modelcontextprotocol/conformance/pull/170/changes#diff-e8d3e80a369c80a2a5bca70310c8e5b8fb59471a5b083aa4ff2deaf2d0baecd1R98). Since [RFC 8414 §3.1](https://www.rfc-editor.org/rfc/rfc8414.html#section-3.1) does not says that the issuer identifier must exclude a trailing slash, I think that removing the slash during issuer comparison is not necessary. What do you think?

src/scenarios/authorization-server/authorization-server-metadata.ts
pcarleton: good catch, ty!

src/scenarios/authorization-server/authorization-server-metadata.ts
pcarleton: hm yea maybe we just want to assume the URL you pass in is the `issuer` URL? 

agree with @Michito-Okai that it doesn't say the issuer itself can't have a trailing slash, just that you have to remove it before inserting the `/.well-known` to find the metadata

src/scenarios/authorization-server/authorization-server-metadata.ts
nbarbettini: @Michito-Okai Looks good, thanks!

src/scenarios/authorization-server/authorization-server-metadata.ts
nbarbettini: @Michito-Okai I missed L98, sry! That looks correct on the discovery side.

I still think there is technically a gap on the comparison side, but it may not matter in the real world. My reasoning chain is:
1.  [RFC 8414 §3.1](https://www.rfc-editor.org/rfc/rfc8414.html#section-3.1) says that if the issuer identifier contains a path component, any trailing / “MUST be removed” before constructing the metadata URL. (L98 ✅)
2. [RFC 8414 §3.3](https://www.rfc-editor.org/rfc/rfc8414.html#section-3.3) then says the returned issuer “MUST be identical” to the issuer identifier value _into which the well-known string was inserted_ to retrieve the metadata.
3. Additionally, [OIDC Discovery §4.3](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationValidation) says it this way: "The issuer value returned MUST be identical to the Issuer URL that was used as the prefix to `/.well-known/openid-configuration`". 

In both RFC 8414 and OIDC Discovery the language is slightly vague because it does not explicitly say "issuers with a path component MUST NOT have a trailing slash in their canonical form". But I do think that is strongly implied by (1) and (2), and reinforced by (3) - if the issuer value must be identical to the _prefix_, the returned issuer value can't have a trailing slash because `GET /issuer1//.well-known/openid-configuration` is invalid.

Note that this _only_ applies to issuers with a path component: `https://as.example.com/issuer1` not `https://as.example.com`

Now, it may not practically matter because many authorization servers are _not_ path-based, and the ones that are _probably_ do this correctly already. MCP conformance isn't trying to be a canonical OIDC conformance suite. If we don't want to tackle this right now, or it's too edgy of an edge case, we could revisit later.

Michito-Okai: Thank you for your explanation.
I agree that the language is slightly vague in both RFC 8414 and OIDC Discovery, and that the issuer URL should not include a trailing slash.
However, since the specifications do not explicity state that issuer URL should not include a trailing slash, I think that we need not to arbitrarily remove a trailing slash.
For example, there may be a case where the issuer URL could be `https://auth.example.com/issuer/` and metadata URL could be `https://auth.example.com/.well-known/openid-configuration/issuer`.
I would like to perform the check assuming that the value set in the `url` parameter is the issuer, i.e., the iss claim returned in the metadata.
What do you think? @nbarbettini @pcarleton

pcarleton: Yea let's go with what's here. 

My interpretation of"“MUST be identical” to the issuer identifier value into which the well-known string was inserted to retrieve the metadata."  is that the removing of the slash is a part of the "insertion", so that still would count. That would mean if you have a trailing slash in one, you have to have it in the other, and vice versa, which seems strict enough and simple enough.

Let's revisit if we see an issue though.

pcarleton (dismissed):
LGTM, thank you!

km-anthropic:
@claude review

pcarleton (approved):

PR #203, opened 2026-03-30
fix: rename routePrefix to issuerPath and add issuer-mismatch scenario

### LINKED ISSUE #140: AS issuer validation failure
**Describe the bug**
I'm currently working on implementing client-side OAuth in Go SDK and I'm running into a conformance test failure which I'm not sure is warranted, so I would like to consult.

The issue is related to [Authorization Server Metadata](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-server-metadata-discovery) handling. In `metadata-var2` and `metadata-var3` tests, the authorization server's issuer identifier contains a non-empty path component `/tenant1`. According to [RFC 8141 section 3.3](https://datatracker.ietf.org/doc/html/rfc8414#section-3.3) a client must validate if the issuer value returned by the authorization server is the same as the identifier used to create the URL for the metadata. However, the AS returns a value with no path component, for example:

```json
 {
    "id": "outgoing-auth-response",
    "name": "Outgoing-auth-response",
    "description": "Sent 200 response for GET /.well-known/oauth-authorization-server/tenant1",
    "status": "INFO",
    "timestamp": "2026-02-10T14:02:08.022Z",
    "details": {
      "method": "GET",
      "path": "/.well-known/oauth-authorization-server/tenant1",
      "statusCode": 200,
      "headers": {
        "x-powered-by": "Express",
        "content-type": "application/json; charset=utf-8",
        "content-length": "417",
        "etag": "W/\"1a1-vdDFr5xugGjES09D01A3xMuA7xw\""
      },
      "body": {
        "issuer": "http://localhost:35909",  <--- HERE
        "authorization_endpoint": "http://localhost:35909/tenant1/authorize",
        "token_endpoint": "http://localhost:35909/tenant1/token",
        "registration_endpoint": "http://localhost:35909/tenant1/register",
        "response_types_supported": [
          "code"
        ],
        "grant_types_supported": [
          "authorization_code",
          "refresh_token"
        ],
        "code_challenge_methods_supported": [
          "S256"
        ],
        "token_endpoint_auth_methods_supported": [
          "none"
        ]
      }
    }
  },
```

Should this behavior be adjusted?

**To Reproduce**
1. Run `metadata-var2` or `metadata-var3` conformance test with a working SDK and inspect `checks.json`.

**Expected behavior**
I would expect the issuer field to contain the path component used to construct the AS metadata URL.

pcarleton:
The previous fix (#152) used routePrefix for the issuer path, conflating two concepts: (1) where OAuth endpoints are mounted and (2) the issuer path component per RFC 8414. This worked for metadata-var2/var3 where both were /tenant1, but broke auth/2025-03-26-oauth-metadata-backcompat where routePrefix was /oauth (endpoint namespacing) but the issuer should be root.

Changes:
- Rename routePrefix -> issuerPath in createAuthServer. Endpoints are mounted under the issuer path (the real-world multi-tenant model).
- Drop the /oauth prefix from march-spec-backcompat entirely. No route collision exists with createServer, and test integrity against fallback-bypass is already guaranteed by expectedSlugs requiring authorization-server-metadata.
- Rename authRoutePrefix -> authIssuerPath in discovery-metadata config.
- Add issuerOverride option to createAuthServer for negative testing.
- Add auth/issuer-mismatch scenario: AS returns a mismatched issuer (https://evil.example.com) and the client is expected to reject per RFC 8414 section 3.3. Skipped in CI for now as the TS SDK reference client does not yet validate issuer.

Fixes #140

### DIFF
--- a/examples/clients/typescript/everything-client.ts
+++ b/examples/clients/typescript/everything-client.ts
@@ -147,6 +147,8 @@ registerScenarios(
     'auth/token-endpoint-auth-none',
     // Resource mismatch (client should error when PRM resource doesn't match)
     'auth/resource-mismatch',
+    // Issuer mismatch (client should error when AS metadata issuer doesn't match, RFC 8414 §3.3)
+    'auth/issuer-mismatch',
     // SEP-2207: Offline access / refresh token guidance (draft)
     'auth/offline-access-scope',
     'auth/offline-access-not-supported'
--- a/src/scenarios/client/auth/discovery-metadata.ts
+++ b/src/scenarios/client/auth/discovery-metadata.ts
@@ -22,8 +22,8 @@ interface MetadataScenarioConfig {
   prmLocation: string;
   inWwwAuth: boolean;
   oauthMetadataLocation: string;
-  /** Route prefix for the auth server (e.g., '/tenant1') */
-  authRoutePrefix?: string;
+  /** Issuer path component for the auth server (e.g., '/tenant1' for multi-tenant) */
+  authIssuerPath?: string;
   /** If true, add a trap for root PRM requests */
   trapRootPrm?: boolean;
 }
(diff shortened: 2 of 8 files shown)


pcarleton:
@claude review

pcarleton:
@claude review

Issue #114, opened 2026-01-22
Server lifecycle management: eliminate boilerplate bash scripts from SDK repos


maxisbey:
## Problem

Every SDK repo that runs server conformance tests needs a bash script that does the same thing:

1. Start the server process in the background
2. Poll until healthy (curl loop with retries)
3. Run `npx @modelcontextprotocol/conformance server --url <url>`
4. Trap + kill the server on exit

Here are the current scripts:

- **TypeScript SDK:** [`scripts/run-server-conformance.sh`](https://github.com/modelcontextprotocol/typescript-sdk/blob/bf6f16c8be10c185e434e068c2a67f6d3baee6b7/scripts/run-server-conformance.sh)
- **Python SDK:** [`.github/actions/conformance/run-server.sh`](https://github.com/modelcontextprotocol/python-sdk/blob/d77292fb0660072f82edb27709375d319b1ec958/.github/actions/conformance/run-server.sh)

These are nearly identical — the only SDK-specific part is the command to start the server. The health-check loop, process management, and cleanup are pure boilerplate that gets copy-pasted across repos.

This creates maintenance burden: if we want to change the health-check behavior (e.g., longer timeout, different readiness check), every SDK repo needs updating. It also means each SDK has to figure out the bash incantation independently.

## Context

The `client` subcommand already manages process lifecycle — it takes a `--command` flag, spawns the client, and handles cleanup. The `server` subcommand currently only connects to an already-running server via `--url`.

With the GitHub Action (#99, PR #113), server workflows still require a separate "start server and wait" step before invoking the action. This is the last piece of per-repo boilerplate that could be eliminated.

## Desired outcome

A single SDK-specific value (the command to start the server) should be all that's needed. The conformance tool should handle spawning, health-checking, and cleanup — the same way it already does for client testing.

## Suggestion

Adding `--command` support to the `server` subcommand (symmetric with how `client` already works) seems like the natural fit. The action could then wire it through for server mode as well.

pcarleton:
Related extension: once `server --command` exists, the next friction point is that not every scenario can run against the same fixture. Some SEP scenarios need servers that don't belong in everything-server, and everything-server is harder to extend than everything-client.

Specifically I'm thinking of the AS-RS combo servers which need to have different server URL's, and therefore can't be tested via a single "Everything server" URL.

A per-SDK manifest mapping scenarios → command would let a single run spawn different fixtures for different scenarios:

```yaml
scenarios:
  - match: "*"
[...]

PR #255, opened 2026-04-24
feat: support "draft" as a first-class spec version target

### LINKED ISSUE #253: Support "draft" as a first-class spec version target
## Problem

SEP authors testing not-yet-released spec changes hit two blockers when running the suite with `protocolVersion: "draft"`:

1. **`initialize` scenario rejects `"draft"`** — `src/checks/client.ts` hard-codes `VALID_PROTOCOL_VERSIONS = ['2025-06-18', '2025-11-25']`, so a client sending `"draft"` fails the `mcp-client-initialization` check with "Version mismatch".
2. **`--spec-version draft` only runs draft-tagged scenarios** — the `initialize` scenario (and most of the core suite) is tagged `['2025-06-18', '2025-11-25']`, so `--spec-version draft` skips it entirely. Tagging it `[..., 'draft']` is blocked by the `spec-version.test.ts` isolation check.

Net effect: there's no way to run the full suite against an SDK that's tracking the draft spec without forking the conformance repo.

Raised by @mikekistler in [#conformance-testing-wg](https://discord.com/channels/1358869848138059966/1428369612361175070/1492186964684701828) (thread, Apr 10–21) while testing SEP-2243. He proposed a long-lived `draft` branch as a workaround; preference here is to make `draft` a first-class spec-version target instead.

## Proposed fix

- Add `LATEST_SPEC_VERSION` and `DATED_SPEC_VERSIONS` constants so the next spec release is a one-line change.
- Treat `"draft"` as a valid protocol version in the `initialize` check and mock-server response (echoes back `"draft"` if the client sends it).
- Make `--spec-version draft` resolve to **latest-dated ∪ draft-tagged** scenarios — i.e., draft is a superset of the current release plus draft-only additions. The `spec-version.test.ts` isolation check is updated to match (a scenario is *tagged* draft-only, but *selected* by `--spec-version draft` if it's in the latest release).
- Pass the resolved spec version to the client process via `MCP_CONFORMANCE_SPEC_VERSION` so SDK examples can pick the right `protocolVersion` without hard-coding.

## Out of scope

- Migrating the ~30 hard-coded `'2025-11-25'` literals across server scenarios — separate cleanup.
- Relationship to the `experimental/` folder proposal (#244 comment thread) — that's about *where scenarios live*, this is about *which version they target*.

pcarleton:
Closes #253.

## Problem

SEP authors testing against the in-progress spec can't run the full suite: the `initialize` check rejects draft `protocolVersion` strings, and `--spec-version draft` only selects the handful of explicitly draft-tagged scenarios (skipping `initialize` and the rest of the core suite). Tagging core scenarios with `'draft'` is blocked by the spec-version isolation test.

Raised by @mikekistler in [#conformance-testing-wg](https://discord.com/channels/1358869848138059966/1428369612361175070/1492186964684701828) while testing SEP-2243.

## Design

`SpecVersion` is now always a **wire `protocolVersion` string** — the same value an SDK sends in `initialize` and the value `--spec-version` accepts:

```ts
export const DRAFT_PROTOCOL_VERSION = 'DRAFT-2026-v1';
export type SpecVersion = DatedSpecVersion | typeof DRAFT_PROTOCOL_VERSION;
```

`DRAFT_PROTOCOL_VERSION` mirrors `LATEST_PROTOCOL_VERSION` in the spec repo's [`schema/draft/schema.ts`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/schema/draft/schema.ts) (which has cycled `DRAFT-2025-v1` → `-v2` → `-v3` → `DRAFT-2026-v1`). When the spec repo bumps it, this constant gets a matching bump and a conformance release — consumers updating to that release see the new value in the type, in `--spec-version` validation, and in scenario tags simultaneously. Stale draft strings are rejected, which is a useful signal that the SDK and conformance package are out of sync.

The CLI also accepts `--spec-version draft` as a one-line alias that resolves to `DRAFT_PROTOCOL_VERSION`, purely for ergonomics; the literal `'draft'` does not appear in the type system.

A `'2026-06-DRAFT'` form was considered and rejected: the spec repo doesn't use it anywhere and the release date isn't known in advance.

## Changes

- **`types.ts`**: add `DATED_SPEC_VERSIONS`, `LATEST_SPEC_VERSION`, `DRAFT_PROTOCOL_VERSION`, and `NEGOTIABLE_PROTOCOL_VERSIONS`. `SpecVersion = DatedSpecVersion | typeof DRAFT_PROTOCOL_VERSION`. `'extension'` moved to a separate `ScenarioSpecTag` type — it's a scenario category, not a point on the version timeline.
- **`checks/client.ts` / `scenarios/client/initialize.ts`**: both use the shared `NEGOTIABLE_PROTOCOL_VERSIONS` constant; the mock server accepts and echoes `DRAFT_PROTOCOL_VERSION`.
- **`scenarios/index.ts`**: new `matchesSpecVersion()` helper. Selecting the draft version resolves to **latest-dated ∪ draft-tagged**, so SEP authors can run the full suite without retagging core scenarios. `resolveSpecVersion` adds the `'draft'` CLI alias. `--spec-version extension` is no longer valid; extension scenarios are reachable via `--suite extensions`.
- **`runner/client.ts` / `index.ts`**: forward the resolved `--spec-version` value to the client process as `MCP_CONFORMANCE_PROTOCOL_VERSION`. Example clients can use it directly as their `protocolVersion`; SDKs that hard-code their version can ignore it.
- **Draft-tagged scenarios** (`offline-access`, `resource-mismatch`, `sep-2164-resource-not-found`): now reference the `DRAFT_PROTOCOL_VERSION` constant, so a draft revision bump is a one-line change in `types.ts`.
- **`tier-check`**: derives tier-scoring versions from `DATED_SPEC_VERSIONS`; draft and extension remain non-scoring. A tier-check run at the draft version scores against latest ∪ draft (i.e., "would this SDK pass tier N if draft shipped today").
- Tests: superset relationship and tag-level isolation asserted separately; positive test for `DRAFT_PROTOCOL_VERSION` acceptance, negative tests for stale `'DRAFT-2025-v1'` and bare `'draft'` on the wire; `'draft'` CLI alias resolution; extension scenarios excluded from all `--spec-version` selections.
- README documents `--spec-version` and the new env var.

## Testing

```
$ npx vitest run
Test Files  8 passed (8)
     Tests  116 passed (116)
```

```
$ node dist/index.js list --spec-version DRAFT-2025-v1
Unknown spec version: DRAFT-2025-v1
Valid versions: 2025-03-26, 2025-06-18, 2025-11-25, DRAFT-2026-v1 (or 'draft' as an alias for DRAFT-2026-v1)
```

Typecheck and lint clean.

## Follow-ups (not in this PR)

- **#256 — range-based tagging**: replace per-scenario `specVersions: [...]` lists with `introducedIn` / optional `removedIn`. The superset model in this PR can't express a SEP that *tightens or removes* an existing requirement (e.g., SEP-986 narrowed valid tool-name format) — there's no subtraction arm. `matchesSpecVersion()` is the seam where range logic slots in.
- Migrating remaining hard-coded `'2025-11-25'` literals in server scenarios to `LATEST_SPEC_VERSION`.

### DIFF
--- a/src/scenarios/client/initialize.ts
+++ b/src/scenarios/client/initialize.ts
@@ -3,7 +3,9 @@ import {
   Scenario,
   ScenarioUrls,
   ConformanceCheck,
-  SpecVersion
+  SpecVersion,
+  LATEST_SPEC_VERSION,
+  DRAFT_PROTOCOL_VERSION
 } from '../../types';
 import { clientChecks } from '../../checks/index';
 
@@ -117,11 +119,15 @@ export class InitializeScenario implements Scenario {
     this.checks.push(clientChecks.createServerInfoCheck(serverInfo));
 
     // Echo back client's version if valid, otherwise use latest
-    const VALID_VERSIONS = ['2025-06-18', '2025-11-25'];
+    const VALID_VERSIONS = [
+      '2025-06-18',
+      LATEST_SPEC_VERSION,
+      DRAFT_PROTOCOL_VERSION
+    ];
     const clientVersion = initializeRequest?.protocolVersion;
     const responseVersion = VALID_VERSIONS.includes(clientVersion)
       ? clientVersion
-      : '2025-11-25';
+      : LATEST_SPEC_VERSION;
 
     const response = {
       jsonrpc: '2.0',
--- a/src/scenarios/client/initialize.ts
+++ b/src/scenarios/client/initialize.ts
@@ -5,7 +5,7 @@ import {
   ConformanceCheck,
   SpecVersion,
   LATEST_SPEC_VERSION,
-  DRAFT_PROTOCOL_VERSION
+  NEGOTIABLE_PROTOCOL_VERSIONS
 } from '../../types';
 import { clientChecks } from '../../checks/index';
 
@@ -119,13 +119,8 @@ export class InitializeScenario implements Scenario {
     this.checks.push(clientChecks.createServerInfoCheck(serverInfo));
 
     // Echo back client's version if valid, otherwise use latest
-    const VALID_VERSIONS = [
-      '2025-06-18',
-      LATEST_SPEC_VERSION,
-      DRAFT_PROTOCOL_VERSION
-    ];
     const clientVersion = initializeRequest?.protocolVersion;
-    const responseVersion = VALID_VERSIONS.includes(clientVersion)
+    const responseVersion = NEGOTIABLE_PROTOCOL_VERSIONS.includes(clientVersion)
       ? clientVersion
       : LATEST_SPEC_VERSION;
 
--- a/src/types.ts
+++ b/src/types.ts
@@ -23,12 +23,34 @@ export interface ConformanceCheck {
   logs?: string[];
 }
 
-export type SpecVersion =
-  | '2025-03-26'
-  | '2025-06-18'
-  | '2025-11-25'
-  | 'draft'
-  | 'extension';
+export const DATED_SPEC_VERSIONS = [
+  '2025-03-26',
+  '2025-06-18',
+  '2025-11-25'
+] as const;
+
+export type DatedSpecVersion = (typeof DATED_SPEC_VERSIONS)[number];
+
+export const LATEST_SPEC_VERSION: DatedSpecVersion = '2025-11-25';
+
+// Mirrors LATEST_PROTOCOL_VERSION in the spec repo's schema/draft/schema.ts.
+// Bump when that constant changes.
+export const DRAFT_PROTOCOL_VERSION = 'DRAFT-2026-v1';
+
+export type SpecVersion = DatedSpecVersion | 'draft' | 'extension';
+
+export function specVersionToProtocolVersion(
+  version: SpecVersion
+): string | undefined {
+  if (version === 'draft') return DRAFT_PROTOCOL_VERSION;
+  // TODO(#253 follow-up): 'extension' isn't a spec version — it's a scenario
+  // category that got lumped into SpecVersion so `--spec-version extension`
+  // could reuse the filter plumbing. It has no corresponding wire
+  // protocolVersion. Split it out of this type when moving to
+  // introducedIn/removedIn tagging.
+  if (version === 'extension') return undefined;
+  return version;
+}
 
 export interface ScenarioUrls {
   serverUrl: string;
--- a/src/types.ts
+++ b/src/types.ts
@@ -33,24 +33,30 @@ export type DatedSpecVersion = (typeof DATED_SPEC_VERSIONS)[number];
 
 export const LATEST_SPEC_VERSION: DatedSpecVersion = '2025-11-25';
 
-// Mirrors LATEST_PROTOCOL_VERSION in the spec repo's schema/draft/schema.ts.
-// Bump when that constant changes.
+/**
+ * Wire `protocolVersion` for the in-progress spec. Mirrors
+ * `LATEST_PROTOCOL_VERSION` in the spec repo's `schema/draft/schema.ts`;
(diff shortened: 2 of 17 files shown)


src/scenarios/client/initialize.ts
```diff
@@ -117,11 +119,15 @@ export class InitializeScenario implements Scenario {
     this.checks.push(clientChecks.createServerInfoCheck(serverInfo));
 
     // Echo back client's version if valid, otherwise use latest
-    const VALID_VERSIONS = ['2025-06-18', '2025-11-25'];
+    const VALID_VERSIONS = [
+      '2025-06-18',
+      LATEST_SPEC_VERSION,
+      DRAFT_PROTOCOL_VERSION
+    ];
```
felixweinberger: could consider consolildating a `VALID_VERSIONS` constant in `types.ts` instead of having 2 copies here and in `client.ts`

src/types.ts
```diff
@@ -23,12 +23,34 @@ export interface ConformanceCheck {
...
+export function specVersionToProtocolVersion(
+  version: SpecVersion
+): string | undefined {
+  if (version === 'draft') return DRAFT_PROTOCOL_VERSION;
+  // TODO(#253 follow-up): 'extension' isn't a spec version — it's a scenario
+  // category that got lumped into SpecVersion so `--spec-version extension`
+  // could reuse the filter plumbing. It has no corresponding wire
+  // protocolVersion. Split it out of this type when moving to
+  // introducedIn/removedIn tagging.
+  if (version === 'extension') return undefined;
```
felixweinberger: feels like something worth cleaning up, spec vs protocolVersion might get more confusing over time the longer we have both of these 🤔

felixweinberger (dismissed):
LGTM, optional nit on consolidating a single constant instead of having 2 copies

src/types.ts
pcarleton: 👍 we need a different way to refer to extensions, it doesn't super make sense to run "all extensions" as a selector.  ties a bit into my comment on #114 about making it easier to selectively run scenarios via a config file

src/types.ts
```diff
@@ -23,12 +23,34 @@ export interface ConformanceCheck {
...
+  '2025-11-25'
+] as const;
+
+export type DatedSpecVersion = (typeof DATED_SPEC_VERSIONS)[number];
+
+export const LATEST_SPEC_VERSION: DatedSpecVersion = '2025-11-25';
+
+// Mirrors LATEST_PROTOCOL_VERSION in the spec repo's schema/draft/schema.ts.
+// Bump when that constant changes.
+export const DRAFT_PROTOCOL_VERSION = 'DRAFT-2026-v1';
```
mikekistler, line 41: I'm confused. What protocol version should we be using for conformance tests of draft features? This one or "draft"? If it is this one, then why the special casing of "draft" for the --spec-version parameter?

pcarleton: fair question. I was thinking that `draft` would be a convenient alias, but looking at it again using the actual draft spec version will mean it's an easier string replace when we finalize the actual spec, and the alias just adds another indirection.

I've reworked it to be this one everywhere.  so you can pass `--spec-version DRAFT-2026-v1` and it will work. It will also work to pass `--spec-version draft` as a convenience, but tagging scenarios you need to use the explicit one now (not `draft`).

felixweinberger (approved):
Thanks for the changes!

mikekistler (approved):
Looks good! 👍 

Just confirmed that I can layer my tests for SEP-2243 on top of this and successfully test the MCP C# reference implementation.

PR #188, opened 2026-03-17
Conformance Tests for SEP-2322 MRTR

CaitieM20:
Draft Conformance tests for the [SEP-2322: Multi Round-Trip Requests](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2322)

Also added code to client-helper.ts to make rawMCP Requests (i.e. basic json requests) this will be generally useful for draft features that may not have reference implementations yet. 

## Motivation and Context
See SEP

## How Has This Been Tested?
Conformance Tests & Reference Implementation in progress work

## Breaking Changes
yes see SEP

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)
- [X] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

- [X] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [X] My code follows the repository's style guidelines
- [X] New and existing tests pass locally - existing tests pass draft tests do not since we don't have an implementation yet
- [X] I have added appropriate error handling
- [X] I have added or updated documentation as needed

PR #259, opened 2026-05-01
Add conformance tests for SEP-2243 HTTP Standardization

mikekistler:
## Summary

Adds conformance test scenarios for [SEP-2243 (HTTP Standardization)](https://modelcontextprotocol.io/specification/draft/basic/transports), covering the new standard and custom MCP request headers.

The PR also adds a "prepare" script to the package.json to allow "npm install" from a branch for pre-merge testing.

### Server scenarios (test server — inspects client requests)
- **`http-standard-headers`** — Verifies clients send `Mcp-Method` and `Mcp-Name` headers on HTTP POST requests
- **`http-custom-headers`** — Verifies clients handle `x-mcp-header` annotations: mirror param values into `Mcp-Param-*` headers, apply Base64 encoding for non-ASCII values, and convert number/boolean values correctly
- **`http-invalid-tool-headers`** — Verifies clients reject invalid `x-mcp-header` tool definitions (empty values, duplicates, non-primitive types, illegal characters, etc.)

### Client scenarios (connect to server under test)
- **`http-header-validation`** — Verifies servers reject mismatched/missing `Mcp-Method`/`Mcp-Name` headers, accept case-insensitive header names, reject case-mismatched values, trim whitespace, validate Base64-encoded custom headers, and return `-32001` (HeaderMismatch)
- **`http-custom-header-server-validation`** — Verifies servers validate custom `Mcp-Param-*` headers against the request body

### Traceability
- `src/scenarios/sep-2243.yaml` maps 30+ spec requirements to check IDs

## Notes
- Server-side scenarios are registered as **pending** until the everything-server implements SEP-2243 header validation
- Includes a `prepare` script in `package.json` for git-based dependency installs

## Checklist
- [x] `npm run build` passes
- [x] `npm test` passes
- [x] Scenarios registered in `src/scenarios/index.ts`

PR #262, opened 2026-05-05
Feat/tasks mrtr extension

### LINKED ISSUE #261: Add SEP-2663 Tasks Extension conformance scenarios
## Proposal

Add server-conformance scenarios for SEP-2663 (Tasks Extension), with incidental coverage of SEP-2575 (per-request
capability override) and SEP-2243 (Mcp-Method/Mcp-Name request headers) in the parts of the surface where they bind to
tasks. Plus one new MRTR-adjacent check (`mrtr-tasks-composition`, currently `SKIPPED`) for the SEP-2663 commit `451f5e1`
MRTR→Tasks promotion flow.

Tagged `['extension', DRAFT_PROTOCOL_VERSION]` per #255 conventions.

**Implementation:** Draft PR (link to be added after open) from `panyam/mcpconformance:feat/tasks-mrtr-extension`. 8
ClientScenario classes / ~33 checks for tasks plus 1 class / 7 checks + 1 SKIPPED for the MRTR↔Tasks composition
placeholder. Full scope, design notes, and test instructions live in the PR description.

## Relationship to PR #188 (SEP-2322 MRTR)

Complementary, not overlapping. SEP-2663 builds on SEP-2322's base types, so a few of the tasks scenarios touch the MRTR
shape (`inputRequests`, `requestState`, `resultType`) in their tasks-on-the-wire form (`status:"input_required"` on
`tasks/get`, `tasks/update` resume path, partial inputResponses fulfillment). The standalone-ephemeral-MRTR coverage stays
in #188.

If #188 lands first (expected order — SEP-2322 ships before SEP-2663), I'll drop the duplicative `mrtr-ephemeral-flow.ts`
checks in favor of #188's scenarios. The `mrtr-tasks-composition` check (currently `SKIPPED`) is the genuinely new
contribution to MRTR coverage — it asserts SEP-2663's MRTR→Tasks promotion flow.

## Open spec questions

1. **MRTR `resultType` discriminator value.** SEP-2322's draft uses `"input_required"`; SEP-2663's draft uses
`"incomplete"`. Centralized in code as `MRTR_INCOMPLETE_RESULT_TYPE` for a one-line flip when the SEP authors converge.

2. **`mrtr-tasks-composition`.** SEP-2663 commit `451f5e1` made the MRTR→Tasks promotion flow normative on the wire: a
single `tools/call` MAY exchange one or more `IncompleteResult` rounds and *then* return `CreateTaskResult` on a subsequent
 round. Implementing this requires the server middleware to defer task creation until the handler signals async-promotion —
 the natural alternative (mint the task up-front the moment a tool advertises task support) doesn't fit, because by the
time the handler's `IsIncomplete` signal is observable, the `CreateTaskResult` is already on the wire. This is a
wire-contract requirement, not an SDK-specific implementation choice; existing SDKs across languages that took the up-front
 pattern will need refactoring before this check can pass anywhere. Combined with #1 above, that's why the check is
`SKIPPED` today.

panyam:
## Summary

Adds server-conformance scenarios for SEP-2663 (Tasks Extension), with incidental coverage of SEP-2575 (per-request  capability override) and SEP-2243 (Mcp-Method/Mcp-Name request headers) in the parts of the surface where they bind to  tasks. Plus one new MRTR-adjacent check (`mrtr-tasks-composition`, currently `SKIPPED`) for the SEP-2663 commit `451f5e1` MRTR→Tasks promotion flow. 8 ClientScenario classes / ~33 internal checks for tasks plus 1 class / 7 SUCCESS + 1 SKIPPED for the MRTR↔Tasks composition placeholder. Tagged `['extension', DRAFT_PROTOCOL_VERSION]` per #255 conventions and registered in `pendingClientScenariosList` so default `everything-server` runs stay green.

## Motivation and Context
SEP-2663 (Tasks Extension), SEP-2575, and SEP-2243 are in active draft and currently have no conformance coverage in this repo. SDKs implementing them - including ones already shipping reference servers - have nothing to validate against, so wire-shape regressions and edge-case behavior (cancellation semantics, requestState handling, capability gating) slip
through SDK-internal tests. This PR fills that gap. The new scenarios assert what the spec text says, not what any specific  implementation does, so any SDK can run them.

The MRTR↔Tasks composition placeholder (`mrtr-tasks-composition`, `SKIPPED`) is a forward-looking marker for SEP-2663 commit `451f5e1`, which made the "MRTR rounds then promote to a task" flow normative on the wire - see the open spec questions below for why it's deferred.

## How Has This Been Tested?

Run end-to-end against a reference Go fixture from the in-flight `panyam/mcpkit` SDK:

```
TASKS_SERVER_URL=http://localhost:18092/mcp \
TASKS_SERVER_CMD="/path/to/tasks-fixture --serve --addr :18092" \
MRTR_SERVER_URL=http://localhost:18093/mcp \
MRTR_SERVER_CMD="/path/to/mrtr-fixture --serve --addr :18093" \
  npx vitest run src/scenarios/server/
```

Branch results:
- **Tasks:** 8/8 scenarios (~33 internal checks)
- **MRTR:** 1/1 scenario (7 SUCCESS + 1 SKIPPED - see Open Question 2 in Additional Context)

The runner is brand-neutral and language-agnostic - fixture wired via env vars, spawn via `sh -c`, readiness via TCP polling, no log-line scanning. Anyone's server in any language works. Reference fixtures:

* https://github.com/panyam/mcpkit/tree/main/examples/tasks-v2 and 
* https://github.com/panyam/mcpkit/tree/main/examples/mrtr.

`npm test` against the upstream `everything-server` continues to pass - the new scenarios live in `pendingClientScenariosList` so `all-scenarios.test.ts` skips them until `everything-server` grows extension support.

## Breaking Changes

None. All new scenarios are additive and tagged as `'extension'` + `DRAFT_PROTOCOL_VERSION`, so they're invisible to dated  `--spec-version` runs and only appear under `--suite extensions` or `--spec-version draft`. Default CI runs against `everything-server` are unaffected (the new scenarios are filtered out via `pendingClientScenariosList`).

## Types of changes

- [ ] Documentation update

## Checklist

### Relationship to PR #188 (SEP-2322 MRTR)

Complementary, not overlapping. SEP-2663 builds on SEP-2322's base types, so a few of the tasks scenarios touch the MRTR shape (`inputRequests`, `requestState`, `resultType`) in their tasks-on-the-wire form (`status:"input_required"` on `tasks/get`, `tasks/update` resume path, partial inputResponses fulfillment). The standalone-ephemeral-MRTR coverage stays in #188.

The branch also contains a `src/scenarios/server/mrtr/` folder with ephemeral-flow scenarios mirroring some of #188's checks. Those exist because the tasks reference fixture exercises the full MRTR base, and running them locally caught a real bug. **For this upstream merge:**

- If #188 lands first (expected order - SEP-2322 ships before SEP-2663), I'll drop the duplicative `mrtr-ephemeral-flow.ts` checks in favor of #188's scenarios.
- The `mrtr-tasks-composition` check (currently `SKIPPED`) is the genuinely new contribution this PR makes to MRTR coverage.

## Scope (8 ClientScenario classes, ~33 checks)

- `tasks-lifecycle` - sync vs task dispatch, DetailedTask shape, tool errors vs protocol errors, cancel ack, cancel-on-terminal -32602
- `tasks-capability-negotiation` - extension advertised under `capabilities.extensions`; `tasks/*` gated behind negotiation; SEP-2575 per-request opt-in
- `tasks-wire-fields` - `ttlSeconds` / `pollIntervalMilliseconds` renames, no early TTL expiry, no `related-task` _meta on inlined result
- `tasks-request-state` - optional emission, echo acceptance, stale-but-valid tolerance (tasks-surface form)
- `tasks-mrtr-input` - inputRequests on tasks/get, tasks/update resume, partial-fulfillment with multi-input fixture
- `tasks-request-headers` - SEP-2243 server tolerates routing headers; body authoritative when conflicting
- `tasks-dispatch-and-envelope` - removed v1 methods (-32601), legacy `task` param ignored, `resultType:"complete"` on every non-task response, strong-consistency immediate tasks/get, unknown taskId -32602
- `tasks-status-notifications` - optional INFO check (notifications are MAY per spec)

Plus `mrtr-ephemeral-flow` (1 class / 7 SUCCESS + 1 SKIPPED) under `src/scenarios/server/mrtr/`.

## Design highlights

- **Platform/Language agnostic runner.** Fixture configured via env vars `TASKS_SERVER_URL` / `TASKS_SERVER_CMD` (and
 `MRTR_SERVER_URL` / `MRTR_SERVER_CMD`). Spawn via `sh -c`, readiness via TCP polling, no log-line scanning. Suite is
`describe.skip`'d when env vars are unset.
- **Raw-fetch escape hatch.** SDK's typed schemas strip the SEP-2663 wire fields (`resultType`, `taskId`, `inputRequests`,
`requestState`, inlined result/error). Helpers in `src/scenarios/server/tasks/helpers.ts` provide `initRawSession` +
`rawRequest`/`rawRequestFull` so scenarios read those fields directly. When the SDK gains schemas for SEP-2663 shapes, the
call sites switch back to `client.request(..., AnyResult)` and the helper shrinks. Similar in spirit to the raw-MCP
additions in #188 - could converge on a shared helper.
- **Registered in `pendingClientScenariosList`** - `all-scenarios.test.ts` skips them since `everything-server` doesn't
implement the extension yet. CLI lookup (`getClientScenario(name)`) still finds them.

## Open spec questions

1. **MRTR `resultType` discriminator value.** SEP-2322's draft uses `"input_required"`; SEP-2663's draft uses `"incomplete"`. Centralized as `MRTR_INCOMPLETE_RESULT_TYPE` for a one-line flip when SEP authors converge. Tracked at modelcontextprotocol/modelcontextprotocol PR 2663 comment 4381885336 and PR 2322 comment 4381884825.

2. **`mrtr-tasks-composition`.** SEP-2663 commit `451f5e1` made the MRTR→Tasks promotion flow normative on the wire: a single `tools/call` MAY exchange one or more `IncompleteResult` rounds and *then* return `CreateTaskResult` on a subsequent round. Implementing this requires the server middleware to defer task creation until the handler signals async-promotion - the natural alternative (mint the task up-front the moment a tool advertises task support) doesn't fit, because by the time the handler's `IsIncomplete` signal is observable, the `CreateTaskResult` is already on the wire. This is a wire-contract requirement, not an SDK-specific implementation choice; existing SDKs across languages that took the up-front pattern will need refactoring before this check can pass anywhere. Combined with #1 above, that's why the check is `SKIPPED` today.

## Closes: https://github.com/modelcontextprotocol/conformance/issues/261


panyam:
@LucaButBoring

src/scenarios/server/tasks/wire-fields.ts
```diff
@@ -106,40 +111,52 @@ export class TasksWireFieldsScenario implements ClientScenario {
...
-        // ttlSeconds — required, positive (or null = unlimited; treat
-        // either as well-formed). Legacy `ttl` MUST be absent.
-        if (!('ttlSeconds' in result)) {
+        // ttlMs — required, positive integer (or null = unlimited; treat
+        // either as well-formed). Legacy `ttl` MUST be absent. Interim
+        // `ttlSeconds` MUST also be absent on a post-2026-05-07 server.
+        if (!('ttlMs' in result)) {
           errs.push(
-            'CreateTaskResult MUST carry ttlSeconds (renamed from v1 `ttl`)'
+            'CreateTaskResult MUST carry ttlMs (renamed from v1 `ttl` and from the interim `ttlSeconds`)'
```
LucaButBoring: We should be sure to clean up these interim comments before finalizing this (they're useful for keeping up with the SEP changes though)

src/scenarios/server/tasks/wire-fields.ts
```diff
@@ -106,40 +111,52 @@ export class TasksWireFieldsScenario implements ClientScenario {
...
           );
         } else if (
-          result.ttlSeconds !== null &&
-          (typeof result.ttlSeconds !== 'number' || result.ttlSeconds <= 0)
+          result.ttlMs !== null &&
+          (typeof result.ttlMs !== 'number' || result.ttlMs <= 0)
```
LucaButBoring: We should add a `Number.isInteger()` check to this and `pollIntervalMs`.

LucaButBoring (changes requested):

src/scenarios/server/tasks/wire-fields.ts
panyam: Done

src/scenarios/server/tasks/wire-fields.ts
panyam: Done - time to start cleaning up anyway

panyam:
Hi @CaitieM20 can I get you to review this too since we are looking at a subset of mrtr tests which intersect with Tasks.  Thanks heaps.

panyam:
SEP-2663 merged Final at modelcontextprotocol/modelcontextprotocol#2663 on 2026-05-15. The branch is now aligned: the suite asserts the four merged normative changes (notification rename, progress/message MUST-NOT on tasks, `requestState` removal from the v2 wire, `-32003` for required-mode without negotiation) as MUSTs rather than tolerating divergence.

Also rebased against `main` to clear the conflict from #259 / #276, and migrated existing tasks + MRTR scenarios to the `source: ScenarioSource` shape from #265 (added `'io.modelcontextprotocol/tasks'` to `EXTENSION_IDS` - happy to move that constant).

Ready for another review pass.

src/scenarios/server/tasks/notifications.ts
```diff
@@ -1,11 +1,17 @@
...
+ * post-merge. Three normative requirements observed here:
+ *   1. When a server emits status notifications, the method MUST be
+ *      `notifications/tasks` (renamed from `notifications/tasks/status`
+ *      before SEP-2663 merged Final).
+ *   2. The legacy method name `notifications/tasks/status` MUST NOT
+ *      appear on the v2 tasks surface.
+ *   3. `notifications/progress` and `notifications/message` MUST NOT be
+ *      sent on the task stream. The check is FAILURE-on-presence —
+ *      servers that fail to filter progress emissions inside their task
+ *      runtime regress here.
```
LucaButBoring: Depends on #271 but flagging this for follow-up as `notifications/tasks` now arrives in response to `subscriptions/listen` rather than opening an SSE stream on `tasks/get`. Under the [updated proposal](https://modelcontextprotocol.io/seps/2663-tasks-extension#task-status-notifications), if the server accepts `subscriptions/listen` for a task, it returns task status notifications there instead, and does not require clients to additionally poll.

src/scenarios/server/tasks/notifications.ts
```diff
@@ -149,51 +162,201 @@ notification params MUST carry:
...
+    // 3f1c3cfc), so the DetailedTask shape on every task-bearing message
+    // — tools/call CreateTaskResult, tasks/get, and these notifications
+    // — must omit it. The CreateTaskResult and tasks/get absence-asserts
+    // live in request-state.ts; this check covers the notification path
+    // by reusing the SSE-observation buffer above.
+    {
+      const id = 'tasks-status-notifications-no-request-state';
+      const name = 'TasksStatusNotificationsNoRequestState';
+      const description =
+        'notifications/tasks payloads MUST NOT carry `requestState` (the merged SEP-2663 removed the field from the Task base interface)';
```
LucaButBoring: We should probably remove these retroactive negative checks before merging, I think I might've mentioned that already at some point - a negative test case for something that was never in v1 is only useful for people who implemented v2 before it was actually merged.

src/scenarios/server/tasks/request-state.ts
```diff
@@ -1,16 +1,24 @@
...
+ * interface and deletes that entire section, so the tasks-v2 wire
+ * carries no `requestState`.
  *
- * If the server does not issue requestState at all (it's optional per
- * SEP-2322), the dependent checks emit INFO rather than failing — the
- * spec allows omission.
+ * This scenario asserts the absence:
+ *   - CreateTaskResult MUST NOT carry `requestState`.
+ *   - DetailedTask (tasks/get response) MUST NOT carry `requestState`,
+ *     regardless of status.
```
LucaButBoring: I'm going to be inconsistent with my reviews here and recommend leaving _these_ retroactive negative tests in, in contrast with the notification ones I commented on before. The reason for that is [SEP-2322](https://modelcontextprotocol.io/seps/2322-MRTR#schema-changes) happens to define `requestState` in exactly the same place we used to define it in `DetailedTask`, so it is actually a foreseeable mistake for a fresh implementation of v2 Tasks to accidentally leave in `requestState` in this context.

src/scenarios/server/tasks/required-task-error.ts
```diff
@@ -0,0 +1,233 @@
...
+ *
+ * The merged SEP-2663 says that a server which cannot service a request
+ * without returning `CreateTaskResult` — i.e. the tool's declared task
+ * support is "required" — MUST return JSON-RPC error code `-32003`
+ * ("Missing Required Client Capability") when the client did not
+ * declare the `io.modelcontextprotocol/tasks` extension during
+ * `initialize`. The error data SHOULD carry a `requiredCapabilities`
+ * object whose shape mirrors the `InitializeRequest` capabilities, so
+ * the client can self-describe what to add without needing out-of-band
+ * documentation.
```
LucaButBoring: This is actually a **MUST** requirement [here](https://modelcontextprotocol.io/seps/2575-stateless-mcp#missing-required-capabilities).

src/scenarios/server/tasks/required-task-error.ts
```diff
@@ -0,0 +1,233 @@
+/**
+ * SEP-2663 Tasks Extension — required-task error conformance.
+ *
+ * The merged SEP-2663 says that a server which cannot service a request
```
LucaButBoring: I recommend a cleanup pass over the whole PR at some point to tidy up these comments which refer to requirements that were dropped midway through the SEP lifecycle and thus won't make sense to people reading this with only knowledge of the merged v1 and v2 states (and not all of the incremental changes made between those points).

src/scenarios/server/mrtr/ephemeral-flow.ts
LucaButBoring: Flagging this for followup - and in the updated composition test, we'll have to ensure that the MRTR phase _does_ allow and maintain `requestState`, while the Task phase _does not_ allow it.

LucaButBoring (changes requested):

src/scenarios/server/tasks/request-state.ts
```diff
@@ -0,0 +1,206 @@
...
+            : undefined,
+          specReferences: [SEP_2663_REF],
+          details: { observedStatus: detailed.status }
+        });
+      } catch (error) {
+        checks.push(failureCheck(id, name, description, error, [SEP_2663_REF]));
+      }
+    }
+
+    // Cleanup so the fixture doesn't leak a 60-second goroutine.
```
LucaButBoring: nit: remove decontextualized comment (also elsewhere)

LucaButBoring (changes requested):

src/scenarios/server/tasks/notifications.ts
panyam: Replaced the whole run() body with a single skipCheck per our Discord agreement to leave the scenario in and skip it. PR 271 has now merged into this branch, so the citation no longer leans on it as a dependency.  The harness will be in a followup once 2575 is implemented in our reference sdk.

src/scenarios/server/tasks/notifications.ts
panyam: Full scenario is now skipped, so none of those inline negative checks execute. Its just 50 ish lines now.

src/scenarios/server/tasks/request-state.ts
panyam: Haha - all good.    I dont know how you are even keeping up with what is consistent and what is not :)

I went ahead and kept the absence-asserts as you suggested and reframed the doc block around the SEP-2322 lexical-adjacency motivation.  This way a fresh reader now sees the "this is a foreseeable copy-paste from `InputRequiredResult`" framing instead of pre-merge history.

src/scenarios/server/tasks/required-task-error.ts
panyam: +1.  Did the cleanup pass across the tasks/* scenarios — dropped the various framings and updated each doc block in terms of the current spec only. The one exception is the SEP-2322 lexical-adjacency context on `request-state.ts`, which I kept because it's the current-spec motivation for the negative test, not incremental SEP history.

src/scenarios/server/tasks/required-task-error.ts
panyam: Fixed.   The JSDoc and the public scenario description now lead with SEP-2575 §"Missing Required Capabilities" as the canonical MUST home of `-32003` and with SEP-2663 §"Required Capabilities" applying it to the tasks extension.  Ive made sure that `specReferences` on the `-32003` checks now carry both SEP refs.

src/scenarios/server/mrtr/ephemeral-flow.ts
panyam: Acked.   This asymmetric invariant (requestState allowed in MRTR but not in tasks) is exactly what  I was hoping the composition test needs to encode across phases. I'll file it as a follow-up alongside the `subscriptions/listen` rewrite once this lands.

src/scenarios/server/tasks/request-state.ts
panyam: Done

src/scenarios/server/tasks/lifecycle.ts
```diff
@@ -0,0 +1,565 @@
...
+          // Status settles to cancelled — observe via tasks/get.
+          const after = (await client.request(
+            { method: 'tasks/get', params: { taskId: cancelTaskId } },
+            AnyResult
+          )) as any;
+          if (after.status !== 'cancelled') {
+            errs.push(
+              `tasks/get after cancel MUST report cancelled; got ${after.status}`
+            );
+          }
```
LucaButBoring: I believe this is missing a wait, given cancellation is eventually-consistent? It should indeed settle to `cancelled` as the comment suggests, but won't necessarily be `cancelled` until it hits its terminal status. Technically it doesn't even need to settle to `cancelled` but we can make that part of the tool contract here so that this test remains reliable.

src/scenarios/server/tasks/lifecycle.ts
```diff
@@ -0,0 +1,565 @@
...
+            specReferences: [SEP_2663_REF, SEP_2322_REF],
+            details: { cancelAck: ack, statusAfterCancel: after.status }
+          });
+        }
+      } catch (error) {
+        checks.push(failureCheck(id, name, description, error, [SEP_2663_REF]));
+      }
+    }
+
+    // Check 8: tasks/cancel on a terminal task MUST return -32602.
```
LucaButBoring: This is actually not required, this is still an empty-ack case - otherwise clients would have to deal with race conditions on this.

src/scenarios/server/tasks/mrtr-input.ts
```diff
@@ -0,0 +1,418 @@
...
+              content: { confirm: true }
+            };
+          }
+          const ack = (await client.request(
+            {
+              method: 'tasks/update',
+              params: {
+                taskId,
+                inputResponses: responses,
+                requestState: inputTask.requestState
```
LucaButBoring: Missed a `requestState` removal here

src/scenarios/server/tasks/README.md
```diff
@@ -0,0 +1,197 @@
...
+| -------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------- |
+| SEP-2663 | Tasks Extension — `io.modelcontextprotocol/tasks` capability, flat `CreateTaskResult` (`Result & Task`), `DetailedTask` on `tasks/get` (with inlined result/error/inputRequests/requestState), `tasks/update` for MRTR resume, ack-only `tasks/cancel`, wire-field renames (`ttlMs`, `pollIntervalMs`, both integer milliseconds per the 2026-05-07 spec commit aligning duration suffixes) | every scenario                      |
```
LucaButBoring: Also missed a `requestState` removal here and at the top of this file.

src/scenarios/server/tasks/lifecycle.ts
```diff
@@ -0,0 +1,565 @@
...
+          });
+        } else {
+          const ack = (await client.request(
+            { method: 'tasks/cancel', params: { taskId: cancelTaskId } },
+            AnyResult
+          )) as any;
+          const errs: string[] = [];
+          // Ack carries only the SEP-2322 discriminator — no task envelope.
+          if (
+            JSON.stringify(ack) !== JSON.stringify({ resultType: 'complete' })
```
LucaButBoring: This check is actually too strict, as we do still permit metadata (all result shapes do)

src/scenarios/server/tasks/README.md
```diff
@@ -0,0 +1,197 @@
...
+| `tasks-no-early-ttl-expiry`                    | Task remains accessible via `tasks/get` for the duration of its `ttlMs`                          |
+| `tasks-no-related-task-meta-on-inlined-result` | v1 `io.modelcontextprotocol/related-task` `_meta` key absent on tasks/get's inlined `result`     |
+
+### `tasks-request-state` (`request-state.ts`)
+
+| Check                                 | What it tests                                                                                                     |
+| ------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
+| `tasks-request-state-shape`           | When emitted, `requestState` is a non-empty string (`INFO` if server omits it; emission is optional per SEP-2322) |
+| `tasks-request-state-echo`            | Server accepts `tasks/get` with the previously-emitted `requestState` echoed back                                 |
+| `tasks-request-state-stale-tolerance` | Earlier (stale-but-still-valid) `requestState` MUST still be accepted after a newer one is minted                 |
```
LucaButBoring: This section is stale now

src/scenarios/server/tasks/capability.ts
```diff
@@ -0,0 +1,309 @@
...
+            { method: tc.method, params: tc.params },
+            AnyResult
+          );
+          errs.push(`${tc.method} MUST reject (it returned a result)`);
+        } catch (e: any) {
+          if (e.code !== -32601) {
+            errs.push(
+              `${tc.method} MUST return -32601; got ${e.code ?? '<missing>'}`
+            );
+          }
```
LucaButBoring: Thinking about it, this actually isn't specified, we should probably return `-32003` here. Making a note of this - `-32003` is also what we use for `subscriptions/listen` on tasks.

LucaButBoring (changes requested):

src/scenarios/server/tasks/lifecycle.ts
panyam: Done sir.   Now the post-cancel obsrv now via `waitForTerminal()` rather than reading immediately, and this is also documented in the the README's `slow_compute` row that it settles to `cancelled` after cancellation so the assertion has a tool-contract anchor.

src/scenarios/server/tasks/capability.ts
panyam: Ive gone ahead and flipped this to `-32003` in both the conformance assertion and the mcpkit fixture  (`gateOnTasksExtension` + the `*RejectedWithoutExtension` tests).   Just checking quickly while we are at it - does the merged SEP-2663 (or SEP-2575) actually reflect `-32003` for this path?  I havent fully checked the latest but was operating under `-32601`.   Wanted to confirm before this conformance lands ahead of the  spec.  Happy to revert to `-32601` if not, or leave as-is with the open-questions flag I added to the README

src/scenarios/server/tasks/README.md
panyam: Ok the `tasks-request-state` table now describes the current absence-asserts scenario (`tasks-create-result-no-request-state` and `tasks-get-detailed-no-request-state`). The prior shape/echo/stale-tolerance rows are gone since they assumed `requestState` lived on the wire

src/scenarios/server/tasks/lifecycle.ts
panyam: Hmm true.   Loosened this now.  The ack check now asserts `resultType === 'complete'` plus absence of the task-envelope fields (`taskId`, `status`, `result`, `error`, `inputRequests`), instead of strict-equality against the literal `{resultType:"complete"}`. `_meta` and any future result-shape metadata pass through.

Also applied the same to the `tasks/update` ack check in `mrtr-input.ts` which had the same  anti-pattern.

src/scenarios/server/tasks/README.md
panyam: +1 `requestState` is no longer listed as a SEP-2663 wire field anywhere (top blurb, Specs table SEP-2663 row, wire-format diff, raw-fetch design note).  Also kept it on the SEP-2322 row where it correctly describes the ephemeral MRTR surface.

src/scenarios/server/tasks/mrtr-input.ts
panyam: Fixed now.   Also `tasks/update` params here now carry only `taskId` + `inputResponses`.  `requestState` belongs only to the  SEP-2322 ephemeral InputRequiredResult surface.

src/scenarios/server/tasks/lifecycle.ts
panyam: Yep flipped this now.

src/scenarios/server/tasks/capability.ts
LucaButBoring: It's unspecified either way - just opened a PR to specify `-32003` in particular on the grounds that `-32601` implies the server doesn't support tasks at all: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2756

LucaButBoring:
Just about everything looks good now, final two issues I think:
1. `mrtr-tasks-composition` is still skipped - we aligned both SEPs on `resultType: "input_required"` before they were merged (we can handle that in the same followup as (2)).
2. `subscriptions/listen` - we discussed this one on Discord already, skip is intentional and we'll have a followup PR for it since the reference implementation hasn't picked up SEP-2575 yet.

LucaButBoring:
@pcarleton Can you take a look at this? We're planning to get this in in its current state with the two cross-SEP suites skipped and ship a fast-follow with the rest (2322+2663 and 2575+2663).

Randgalt:
I did a quick test against the server we're developing. The protocol version header is always `2025-11-25` and fails our server. FYI

panyam:
> I did a quick test against the server we're developing. The protocol version header is always `2025-11-25` and fails our server. FYI

Hey thanks for reaching out.   Can you share what failures you are seeing (the latest tasks is a breaking change as it also deprecates the old tasks which was part of the core).

panyam:
Thanks for surfacing this @Randgalt.   Reproduced the failure and just pushed a fix to the branch. 

The root cause was that the the SDK's `Client.connect()` pins `protocolVersion` to its package-level `LATEST_PROTOCOL_VERSION` constant (currently `"2025-11-25"`) inside the `initialize` body. The scenarios in this PR are tagged with `DRAFT_PROTOCOL_VERSION` for filtering purposes, but that tag never reached the wire. So a draft-only server like yours correctly rejected the handshake with `unsupported protocol version`. The earlier PR body's reference to a "raw-fetch escape hatch" was supposed to be done anticipating this but the the helper never landed.

The fix is in now.  it is a new `initRawSession` helper in `src/scenarios/server/tasks/helpers.ts` that does a raw fetch initialize carrying the draft version, captures the session ID, posts `notifications/initialized`, and exposes a small `request` / `requestFull` / `notification` / `close` surface to the scenarios. SSE response parsing is folded in so the previous per-scenario raw helper in `headers.ts` is retired.

All 8 SEP-2663 tasks scenarios + the MRTR ephemeral-flow scenario migrate are migrated and errors are still thrown as `McpError` so existing `instanceof McpError` and `.code` checks keep working. tsc + eslint + the local server suite stay clean.

Do you mind retesting against the updated branch when you have a moment? If your server uses a different draft tag (RC, dated variant, etc.) we can wire that through `initRawSession`'s `protocolVersion` option - it's a parameter now, not a constant.

@LucaButBoring Can you take a look at this param to accept the new draft in the initialize negotiation as well?

Randgalt:
> Thanks for surfacing this @Randgalt. Reproduced the failure and just pushed a fix to the branch.
> 
> Do you mind retesting against the updated branch when you have a moment? If your server uses a different draft tag (RC, dated variant, etc.) we can wire that through `initRawSession`'s `protocolVersion` option - it's a parameter now, not a constant.

The protocol version is correct however most of the required headers/_meta are missing. I mentioned this in Discord regarding the other new tests. https://discord.com/channels/1358869848138059966/1508044920680611912

panyam:
@LucaButBoring  @pcarleton Ive gone ahead and also added tests for validating tasks in the stateless mode too (SEP 2575).   Part of this was setting up the harness to create sessions in various modes that are not currently in the ts-sdk or in the conf harness.   (thanks @Randgalt for pointing this).

As per our original plan we are eyeing a merge of this (The Task+MRTR and the Tasks+SEP "wire-up" via subscriptions/listen would still be a follow up as planned).

panyam:
> > Thanks for surfacing this @Randgalt. Reproduced the failure and just pushed a fix to the branch.
> > Do you mind retesting against the updated branch when you have a moment? If your server uses a different draft tag (RC, dated variant, etc.) we can wire that through `initRawSession`'s `protocolVersion` option - it's a parameter now, not a constant.
> 
> The protocol version is correct however most of the required headers/_meta are missing. I mentioned this in Discord regarding the other new tests. https://discord.com/channels/1358869848138059966/1508044920680611912

Thanks for getting back. The missing `_meta` envelope and SEP-2243 routing headers are addressed in [panyam/mcpconformance PR 5](https://github.com/panyam/mcpconformance/pull/5), which makes `initRawSession` SEP-2575-aware (per-request `_meta`, `MCP-Protocol-Version` header, auto-emitted `Mcp-Method` / `Mcp-Name`) and flips the defaults to `legacy+stateless` (from just testing against legacy).

it is being reviewed right now.  I would value your thoughts/review too since your strict-draft server is a good target to test this  cleanly against. I'll ping back once it merges into this branch.   If the Discord thread has specifics worth folding in before merge, happy to land them on the PR 5 branch directly.

Randgalt:
I did some testing with the latest. There is still a problem where tests don't isolate the condition they are testing for. For example:

```
{
    "id": "tasks-removed-tasks-result",
    "name": "TasksRemovedTasksResult",
    "description": "tasks/result is removed in v2 and MUST reject with -32601",
    "status": "FAILURE",
    "timestamp": "2026-05-29T09:32:28.766Z",
    "errorMessage": "expected -32601; got -32602",
    "specReferences": [
      {
        "id": "SEP-2663",
        "url": "https://modelcontextprotocol.io/seps/2663-tasks-extension"
      }
    ]
  }
```

Our server is sending `-32602` because there are missing headers/_meta, which we check first. The tests need to ensure that only the condition they are testing for is set.

panyam:
thanks @Randgalt Good catch. You're right that the test wasn't isolating the dispatch dimension - fixing it to send a fully-formed v2 request envelope (incl. Mcp-Name: <taskId> mirroring params.taskId per SEP-2663 (Streamable HTTP: Routing Headers), so the only thing servers can reject on is the removed method itself.

One adjacent thing Ive been thinking and might be worth considering is when your server rejects on routing-header validation, the spec-mandated code I beleive is -32001 HeaderMismatch [SEP-2243 - Server Behavior](https://modelcontextprotocol.io/seps/2243-http-standardization#server-behavior) , not -32602. Worth a look on your side - would have made the test failure self-explanatory.  IIRC -32602 are the 404 equivalents for tasks and not missing headers

So couple of quick questions (@LucaButBoring your read here would be useful).   SEP-2663 (Streamable HTTP: Routing Headers) says:

>> "the client MUST set the Mcp-Name header to the value of params.taskId"

on tasks/get / tasks/update / tasks/cancel. The mismatch case is already covered by SEP-2243 (Server Behavior) - its MUST-reject-on-mismatch rule is method-agnostic and picks up any header that mirrors a body field. So a server seeing Mcp-Name != params.taskId on tasks/get MUST return -32001. No question there.

But a couple of narrower questions still I think not settled are:

1. Missing header on the live tasks methods. Does SEP-2663's "client MUST set" elevate Mcp-Name to "required standard header" status for tasks/get / tasks/update / tasks/cancel - so missing it triggers -32001 per SEP-2243's Validation Failure Conditions? Or is missing-only allowed to pass through to dispatch?  This would determine whether reference servers (like mcpkit and other SDKs that don't gate tasks/* on Mcp-Name) need to add the missing-header check.

2. Removed methods and validation order. When tasks/result arrives at a server that validates routing headers before method dispatch, which code wins - -32001 (header validation, if scope 1 expands that far) or -32601 (method removed, per SEP-2663 line 974)? Or is precedence intentionally unspecified?   I would assume headers first before body helps with fail-fast.

Randgalt:
> the spec-mandated code I beleive is -32001 HeaderMismatch

I having trouble following all of this tbh. I've been using the new alpha conformance tests for validation.

LucaButBoring:
> Does SEP-2663's "client MUST set" elevate Mcp-Name to "required standard header" status for tasks/get / tasks/update / tasks/cancel - so missing it triggers -32001 per SEP-2243's Validation Failure Conditions?

Correct, this is the intent of that verbiage.

> When tasks/result arrives at a server that validates routing headers before method dispatch, which code wins - -32001 (header validation, if scope 1 expands that far) or -32601 (method removed, per SEP-2663 line 974)? Or is precedence intentionally unspecified?

This returns `-32601` - we never introduced a requirement that `tasks/result` carry `Mcp-Name`, so we fortunately don't need to worry about this case.

panyam:
Quick update since @LucaButBoring confirmed both spec questions raised above:

1. **Missing `Mcp-Name` on `tasks/get` / `tasks/update` / `tasks/cancel`** --> server MUST reject with `-32001` (HeaderMismatch), not `-32602`.
2. **`tasks/result`** --> server MUST return `-32601` regardless of routing headers; the SEP doesn't require `Mcp-Name` on that removed method.

For servers -  where you're returning `-32602` for routing-header issues, the spec-mandated code is `-32001` per SEP-2243. And rejecting `tasks/result` for missing `Mcp-Name` isn't spec-required - it should fall through to dispatch and return `-32601`.

panyam:
@LucaButBoring @CaitieM20  FYI I just added `sep-2663.yaml` as we discussed.  Plus `mrtr-tasks-composition` is now active (one fewer skip from the list).

Randgalt:
FYI - I just pulled the latest and am getting this:

```
{
    "id": "tasks-sync-tool-call",
    "name": "TasksSyncToolCall",
    "description": "Sync tool returns ToolResult (resultType:\"complete\"), no taskId at top level",
    "status": "FAILURE",
    "timestamp": "2026-05-31T08:09:03.534Z",
    "errorMessage": "MCP error -32602: Missing required metadata: io.modelcontextprotocol/protocolVersion",
    "specReferences": [
      {
        "id": "SEP-2663",
        "url": "https://modelcontextprotocol.io/seps/2663-tasks-extension"
      }
    ]
  }
```

panyam:
Thanks for the repro @randgalt - it turned out your server was right and our harness was lenient here. SEP-2575 removes the initialize handshake on DRAFT-2026-v1 and mandates per-request _meta.io.modelcontextprotocol/protocolVersion, but the harness was still driving the legacy initialize+session wire on draft (no _meta), which any spec-strict server correctly rejected with -32602. Fixed in latest - the legacy-wire describe block no longer runs on draft, and the stateless-wire run already injects _meta correctly, so a re-pull should clear these.

panyam:
@pcarleton @CaitieM20 @LucaButBoring - Just merged main into this PR - this has the option of selected legacy vs stateless wire modes so we can modulate tests against a couple of versions going back.   Tasks needs this for validating conformance with stateless and legacy.

LucaButBoring (approved):

Randgalt:
@panyam I just rebased, rebuilt and ran and it's still missing metadata. For example, I ran:

```
npm  start  -- server --scenario tasks-lifecycle  --url <my-url> --verbose
```

And when `tools/list` is called there is no metadata

src/connection/index.ts
```diff
@@ -49,6 +49,14 @@ export interface RunContext {
...
    */
   connect(): Promise<Connection>;
+  /**
+   * Wire mode override for suites that exercise both the legacy and
+   * SEP-2575 stateless wires against a single spec version (notably
+   * SEP-2663 tasks and SEP-2322 MRTR). When absent the wire is implied
+   * by `specVersion` — `2025-x` ⇒ legacy, draft ⇒ stateless. Scenarios
+   * outside the tasks/mrtr suites ignore this knob.
+   */
+  wire?: 'legacy' | 'stateless';
```
pcarleton: hmm can you explain this? I don't think this should be necessary, i.e. it should be implied by the spec version

src/scenarios/server/_shared/raw-session.ts
```diff
@@ -0,0 +1,692 @@
+/**
```
pcarleton: I don't think you need this for tasks. Tasks is only 2026-07-28 and beyond, so it should be fine using the stuff setup in `connection/`  there are maybe a few features to be merged in there, but let's not create a whole new abstraction for it

src/scenarios/server/_shared/checks.ts
```diff
@@ -0,0 +1,67 @@
+/**
```
pcarleton: this `_shared` directory is not a pattern we use in this repo. you can include it in `../server` or a helpers dir.

src/scenarios/server/_shared/wire-mode.ts
```diff
@@ -0,0 +1,93 @@
+/**
```
pcarleton: this also seems to duplicate the #318 behavior, please rebase on top of that

src/scenarios/server/tasks/capability.ts
```diff
@@ -0,0 +1,278 @@
...
+  ConformanceCheck,
+  ScenarioSource
+} from '../../../types';
+import type { RunContext } from '../../../connection';
+import { SEP_2575_REF, SEP_2663_REF } from '../_shared/sep-refs';
+import { errMsg, failureCheck } from '../_shared/checks';
+import { initRawSession, type RawSession } from '../_shared/raw-session';
+import { TASKS_EXTENSION_ID } from './helpers';
+
+export class TasksCapabilityNegotiationScenario implements ClientScenario {
```
pcarleton, line 19: here and elsewhere: the scenario description needs to describe what the server has to implement, e.g. a tool with this name that's expected to do this behavior.

src/scenarios/server/mrtr/ephemeral-flow.ts
```diff
@@ -0,0 +1,645 @@
+/**
```
pcarleton: please dedupe this with #188

src/scenarios/server/mrtr/README.md
```diff
@@ -0,0 +1,82 @@
...
+| `mrtr-multiple-input-requests-one-round`         | A single `InputRequiredResult` MAY carry inputRequests for `elicitation/create` + `sampling/createMessage` + `roots/list` together                                                  |
+| `mrtr-multi-round-flow`                          | A handler MAY take 2+ rounds; each round mints a fresh `requestState`; final result reflects answers from every round                                                               |
+| `mrtr-wrong-input-key-rerequests`                | When client sends a wrong `inputResponses` key, server SHOULD re-request via `InputRequiredResult` rather than erroring                                                             |
+| `sep-2663-mrtr-synchronous-before-task-creation` | MRTR → Tasks composition: handler exchanges `InputRequiredResult` rounds first, then returns `CreateTaskResult` on a subsequent round (SEP-2663 §"Task Creation", commit `451f5e1`) |
+
+## Required server fixtures
```
pcarleton: this belongs in the scenario description, not in a readme

src/scenarios/server/mrtr/README.md
```diff
@@ -0,0 +1,82 @@
...
+| `test_incomplete_result_request_state`   | Exercises `requestState` validation; final result includes `state-ok` to confirm validation |
+| `test_incomplete_result_multiple_inputs` | Emits 3+ inputRequests of different methods in one round                                    |
+| `test_incomplete_result_multi_round`     | Drives 2+ MRTR rounds, final result references every answer                                 |
+| `test_incomplete_result_elicitation`     | Emits inputRequest for `user_name`; server re-requests on wrong-key responses               |
+
+The fixture can be implemented in any language; one example reference
+implementation lives at
+[`panyam/mcpkit/examples/mrtr`](https://github.com/panyam/mcpkit/tree/main/examples/mrtr).
+
+## Running
```
pcarleton: this duplicates root behavior

src/scenarios/server/mrtr/README.md
```diff
@@ -0,0 +1,82 @@
+# SEP-2322 MRTR — Server Conformance
```
pcarleton: generally this readme seems repetitive, please remove

src/scenarios/server/tasks/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+Tests any MCP server that implements the `io.modelcontextprotocol/tasks`
+extension (SEP-2663) plus the SEP-2322 base types it builds on, the
+SEP-2575 per-request capability override, and the SEP-2243 routing
+headers.
+
+The scenarios assert what the spec text says — not what any particular
+implementation does. When the SDK schemas in
+`@modelcontextprotocol/sdk/types.js` lag the spec, scenarios bypass
+the SDK and use raw `fetch` so the SEP-2663 wire fields (`resultType`,
```
pcarleton: this is al redundant

src/scenarios/server/tasks/all-scenarios.test.ts
```diff
@@ -0,0 +1,167 @@
+/**
```
pcarleton: i'm not sure what this file is for. this seems to duplicate the test running harness

src/scenarios/server/mrtr/all-scenarios.test.ts
```diff
@@ -0,0 +1,142 @@
+/**
```
pcarleton: similar here, we don't need a separate runner per scenario. if you want to group scenarios, do it in a suite.

src/scenarios/server/_shared/test-runner.ts
```diff
@@ -0,0 +1,56 @@
+/**
```
pcarleton: if you want changes to the harness, add it to the harness, not per category

pcarleton (changes requested):

src/scenarios/server/_shared/raw-session.ts
panyam: I agree.  I started this PR before Felix's #318 was merged.  Il try to port this to use that instead

src/scenarios/server/mrtr/ephemeral-flow.ts
panyam: Ok Dedup done in https://github.com/panyam/mcpconformance/pull/8/.

Now down frm 8 -> 1 test.  The other 7 had started before mrtr had merged in.

src/connection/index.ts
panyam: Yep. The field was added to drive a (wire × scenario) matrix back when both wires were spec-valid on the draft.   Since SEP-2575 removed the legacy `initialize` on `DRAFT-2026-v1` collapsing that to a single col and turning the field into a way for callers to go out of sync with `specVersion`. 

Dropped it - wire now derives from `specVersion` via `isStateless(ctx)` in `connection/select.ts`. Follow-up PR: panyam/mcpconformance#8.

src/scenarios/server/_shared/raw-session.ts
panyam: Also now rebased on the Connection abstraction in the follow-up. `raw-session.ts` is gone, its routing-headers + `_meta` envelope behavior lives in `connection/stateless.ts`, and the three things scenarios actually used beyond the existing Connection surface (`extraHeaders`, `discover()`, capability bootstrap via `ConnectOptions`) are now on the Connection interface.

src/scenarios/server/_shared/checks.ts
panyam: Yep, `_shared/` is now gone and the three helper modules (`checks`, `sep-refs`, `wire-format`) are consolidated into `scenarios/server/tasks-mrtr-helpers.ts` matching the existing `input-required-result-helpers.ts` precedent.

src/scenarios/server/_shared/wire-mode.ts
panyam: Done. Helper is now one function (`isStateless(ctx)`) which lives in `connection/select.ts` next to `connectFor`, sharing the single `STATEFUL_VERSIONS` source of truth.

src/scenarios/server/_shared/test-runner.ts
panyam: Gone along with the per-suite test files. Will provide the fixture-spawn convenience a `--spawn-cmd` flag on the CLI in a separate harness PR.

src/scenarios/server/mrtr/all-scenarios.test.ts
panyam: Yep deleted

src/scenarios/server/tasks/all-scenarios.test.ts
panyam: Deleted. The existing CLI is now the entry point.

src/scenarios/server/mrtr/README.md
panyam: Yep deleted

src/scenarios/server/mrtr/README.md
panyam: Removed

src/scenarios/server/mrtr/README.md
panyam: Removed

src/scenarios/server/tasks/capability.ts
panyam: Sure.  Ive now swept every tasks + MRTR scenario and the `description` strings now ends with an explicit "Required server fixtures" section listing each tool name + behavior so an implementer reading the runner's output has it without going to source.

Randgalt:
FYI - I just pulled the latest and tests are missing `io.modelcontextprotocol/clientInfo`.

jokerghotam21-dotcom (changes requested):
Is fixed

pcarleton (approved):
LGTM with the three small commits applied (runner ConnectOptions forwarding, description/spec alignment, connection nits).

One fast-follow needed post-merge: `capability.ts`, `headers.ts`, and `required-task-error.ts` still hardcode `-32003`/`-32001`; main's `spec-types/draft.ts` now exports `MISSING_REQUIRED_CLIENT_CAPABILITY = -32021` and `HEADER_MISMATCH = -32020` per modelcontextprotocol#2907. Import those constants and decide whether to rename the `*-32003` check-ID slugs. Scenarios are in `pendingClientScenariosList` so default CI is unaffected meanwhile.

PR #263, opened 2026-05-06
fix(tier-check): swap server/client scenario lists in conformance reconciliation

blackwell-systems:
## Summary

`tier-check` reports 0% server conformance even when all server tests pass because `checkConformance` (server check) reconciles results against the **client** scenario list, and `checkClientConformance` (client check) reconciles against the **server** scenario list.

Server scenario names (`initialize`, `tools_call`, etc.) have zero overlap with client scenario names (`server-initialize`, `tools-list`, etc.), so every passed scenario is counted as "missing."

```
Server scenarios: 26 names  (initialize, tools_call, ...)
Client scenarios: 31 names  (server-initialize, tools-list, ...)
Overlap: 0
```

## Fix

Swap the scenario list functions so each check reconciles against its own scenario type:
- `checkConformance` now uses `listScenarios()` / `listScenariosForSpec()` (server scenarios)
- `checkClientConformance` now uses `listActiveClientScenarios()` / `listClientScenariosForSpec()` (client scenarios)

Fixes #182

### DIFF
--- a/src/tier-check/checks/test-conformance-results.ts
+++ b/src/tier-check/checks/test-conformance-results.ts
@@ -211,12 +211,9 @@ export async function checkConformance(options: {
     // Non-zero exit is expected when tests fail — results are still in outputDir
   }
 
-  const activeScenarios = new Set(listActiveClientScenarios());
   const expectedScenarios = options.specVersion
-    ? listClientScenariosForSpec(options.specVersion).filter((s) =>
-        activeScenarios.has(s)
-      )
-    : [...activeScenarios];
+    ? listScenariosForSpec(options.specVersion)
+    : listScenarios();
 
   return reconcileWithExpected(
     parseOutputDir(outputDir),
@@ -268,9 +265,12 @@ export async function checkClientConformance(options: {
     // Non-zero exit is expected when tests fail — results are still in outputDir
   }
 
+  const activeScenarios = new Set(listActiveClientScenarios());
   const expectedScenarios = options.specVersion
-    ? listScenariosForSpec(options.specVersion)
-    : listScenarios();
+    ? listClientScenariosForSpec(options.specVersion).filter((s) =>
+        activeScenarios.has(s)
+      )
+    : [...activeScenarios];
 
   return reconcileWithExpected(parseOutputDir(outputDir), expectedScenarios);
 }


marcus-kepler-92:
I think this swap would reproduce the `0/N` it's meant to fix, in both directions —
and #182 looks like it was already fixed by a different change.

The naming that trips this up is on the *types*, not the directories. Scenarios in
`src/scenarios/server/` are the ones that test a server, driven by the `server`
command (`src/index.ts:426-429`) — but each class implements `ClientScenario`,
because the harness plays the client role, so the list function is called
`listActiveClientScenarios()`. `checkConformance` reaching for something named
"client scenarios" reads wrong while being right. The mirror holds for
`src/scenarios/client/` → `Scenario` → `listScenarios()` → the `client` command
(`src/index.ts:130`). There's a note to this effect at `src/index.ts:58-61`.

The two name spaces are disjoint — `listScenarios()` and
`listActiveClientScenarios()` share zero entries — which is what makes the swap
fatal. Measured on `main` (81eb1c3), running each command so it emits results, then
resolving names the way the harness does (`globSync('**/checks.json')` +
`stripTimestamp` + the `'server'` `resultPrefix` that `reconcileWithExpected`
strips):

```
node dist/index.js server --url http://127.0.0.1:9/mcp -o /tmp/sv
  → 31 results
     in listActiveClientScenarios()  [main]      31 / 31    exact match
     in listScenarios()              [this PR]    0 / 44

node dist/index.js client --command /usr/bin/false --suite all -o /tmp/cl
  → 44 results
     in listScenarios()              [main]      44 / 44    exact match
     in listActiveClientScenarios()  [this PR]    0 / 31
```

(The glob is recursive — 32 of the 44 client results are nested under `auth/`, so a
top-level directory count undercounts them.)

Both pairings currently match exactly, 0 missing and 0 extra. After the swap
neither matches anything, so `reconcileWithExpected` counts every expected scenario
as missing — the `0/N` symptom.

On #182 itself: it looks already fixed, by e702a25 (#175), which `git tag
--contains` puts in v0.1.16. At v0.1.15 the subprocess was

```ts
execSync(
  `node dist/index.js server --url ${options.serverUrl} -o ${outputDir}`,
  { cwd: process.cwd(), ... }
)
```

wrapped in a bare `catch {}`. The reporter used
`npx @modelcontextprotocol/conformance@0.1.15`, so unless they ran it from a
checkout of this repo, `./dist/index.js` wouldn't resolve; the throw would be
swallowed, the output directory would stay empty, and reconciliation would count
every expected scenario as missing. e702a25 replaced the path with
`node "${process.argv[1]}"`.

The report's own denominator supports that the lists were never the problem. At
v0.1.15, `listActiveClientScenarios()` returned 30 and `listScenarios()` returned
24. #182 reports `0/30` — so the server check was already reconciling against the
30-entry list this PR would replace. Had it been using `listScenarios()`, the
report would have read `0/24`.

Not my call whether #182 gets closed, but it may be worth re-testing against a
current release before this merges.

pcarleton:
Closing as part of a maintainer backlog sweep

Issue #208, opened 2026-04-03
Add positive tests for the Authorization Code Grant


Michito-Okai:
# Background
[The MCP specification 2025-11-25](https://modelcontextprotocol.io/specification/2025-11-25/basic/authorization#authorization-flow-steps) adheres to the Authorization Code Grant in [OAuth 2.1](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-15.html#name-grant-types). Therefore, the conformance test for  authorization servers must verify whether the authorization server correctly implements the Authorization Code Grant.

# First Step
As a first step, this issue focuses on implementing positive test cases for the Authorization Code Grant.

Michito-Okai:
@pcarleton Hello, if the issue makes sense, could you assign me to the issue? I cannot assign myself to the issue.

nbarbettini:
@Michito-Okai We need to get https://github.com/modelcontextprotocol/conformance/pull/105 merged in. It is not the same thing as what you want to test here (testing the RS with a mocked/fake AS in #105 vs. testing the AS itself here), but I think it would give you some of the harness mechanics you need for this. What do you think?

Michito-Okai:
@nbarbettini Thank you very much. I think that #105, including `oauth-client.ts`, cloud be used to implement a fake client.
Should I wait until #105 is merged, or should I extract the parts I need from #105 for this positive tests? What do you think?

tnorimat:
@nbarbettini @Michito-Okai Hello, I prefer the option "extracting the parts from #105" because if waiting for #105 being merged, the work for the issue cannot go forward now.

nbarbettini:
@Michito-Okai @tnorimat I hope to merge #105 soon, but in the meantime I agree it makes sense to extract what you need so you are not blocked. Good plan!

Michito-Okai:
@pcarleton @nbarbettini @tnorimat 
I updated [this thread](https://github.com/modelcontextprotocol/conformance/issues/208#issue-4198961133) and submitted [PR](https://github.com/modelcontextprotocol/conformance/pull/267).
Could you please check them.

PR #228, opened 2026-04-15
feat: add JSON-based setup option to authorization server conformance…

Michito-Okai:
## Motivation and Context
Described in #225

## How Has This Been Tested?
- npm test (Test Files  7 passed, Tests  92 passed)
- npm test -- src/scenarios/authorization-server/authorization-server-metadata.test.ts (Test Files  1 passed, Tests  3 passed)
- npm run start -- authorization --url http://localhost:8080/.well-known/oauth-authorization-server/realms/mcp
- npm run start -- authorization --url http://localhost:8080/.well-known/oauth-authorization-server/realms/mcp --file authorization-server-settings.json
- npm run start -- authorization --file authorization-server-settings.json
- npm run start -- authorization

## Breaking Changes
- add file option for authorization server mode (--file <authorization server settings file>)

## Types of changes
- [x] New feature (non-breaking change which adds functionality)

## Checklist

- [x] I have read the [MCP Documentation](https://modelcontextprotocol.io)
- [x] My code follows the repository's style guidelines
- [x] New and existing tests pass locally
- [x] I have added appropriate error handling
- [x] I have added or updated documentation as needed

PR #267, opened 2026-05-12
feat: add positive tests for the Authorization Code Grant

Michito-Okai:
## Motivation and Context
Described in #208

## How Has This Been Tested?
- npm test (Test Files  10 passed, Tests  126 passed)
- npm run start -- authorization --url http://localhost:8080/realms/mcp --client-id testClient --secret qeVFKk0g4qVTyeiX93ZW9ffdOA19L0QO --port 3000

## Breaking Changes

- add file options for authorization server mode (--client-id <client ID> --secret <client secret> --port <redirect uri port>)

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

### DIFF
--- a/src/scenarios/authorization-server/auth/helpers/createCallbackServer.ts
+++ b/src/scenarios/authorization-server/auth/helpers/createCallbackServer.ts
@@ -1,0 +1,40 @@
+import express from 'express';
+
+export interface CallbackServer {
+  waitForCallback: (timeoutMs: number) => Promise<string>;
+}
+
+export function startCallbackServer(port: number): CallbackServer {
+  const app = express();
+
+  let resolveFn: (url: string) => void;
+
+  const promise = new Promise<string>((resolve) => {
+    resolveFn = resolve;
+  });
+
+  const server = app.listen(port, '127.0.0.1', () => {
+    console.log(`Callback server started: http://localhost:${port}`);
+  });
+
+  app.use((req, res) => {
+    const fullUrl = `${req.protocol}://${req.get('host')}${req.originalUrl}`;
+    res.send('OK. You can close this page.');
+
+    server.close();
+    resolveFn(fullUrl);
+  });
+
+  return {
+    waitForCallback: (timeoutMs: number) =>
+      Promise.race([
+        promise,
+        new Promise<string>((_, reject) =>
+          setTimeout(() => {
+            server.close();
+            reject(new Error('Timeout: No callback received'));
+          }, timeoutMs)
+        )
+      ])
+  };
+}
--- a/src/scenarios/authorization-server/authorization-code-grant.ts
+++ b/src/scenarios/authorization-server/authorization-code-grant.ts
@@ -1,0 +1,309 @@
+/**
+ * Authorization code grant test scenarios for MCP authorization servers
+ */
+import {
+  ClientScenarioForAuthorizationServer,
+  ConformanceCheck
+} from '../../types';
+import { startCallbackServer } from '../authorization-server/auth/helpers/createCallbackServer';
+import { request } from 'undici';
+import { randomBytes } from 'crypto';
+import { AuthorizationServerOptions } from '../../schemas';
+import { SpecReferences } from '../authorization-server/auth/spec-references';
+
+const REDIRECT_URI_ORIGIN = 'http://localhost';
+const REDIRECT_URI_PATH = '/callback';
+// These values are from RFC 7636 Appendix B.
+const CODE_VERIFIER = 'dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk';
+const CODE_CHALLENGE = 'E9Melhoa2OwvFrEMTJguCHaoeK1t8URWbuGJSstw-cM';
+
+export class AuthorizationCodeGrantScenario implements ClientScenarioForAuthorizationServer {
+  private state = randomBytes(32).toString('base64url');
+  name = 'authorization-code-grant';
+  readonly source = { introducedIn: '2025-03-26' } as const;
+  description = `Test authorization code grant.
+
+**Authorization Server Implementation Requirements:**
+
+**Endpoint**: \`authorization endpoint\`, \`token endpoint\`
+
+**Requirements**:
+- The URI in the authorization response MUST match the redirect_uri parameter in the authorization request
+- The code parameter MUST be present in the authorization response query parameters
+- The code parameter MUST have a value
+- The state parameter in the authorization response MUST match the state parameter in the authorization request query parameters if the state parameter is present in the authorization request query parameters
+- The iss parameter in the authorization response MUST match the issuer claim of authorization server metadata if the iss parameter is present in the authorization response query parameters
+- The code, state and iss parameters MUST NOT appear more than once
+- The code_challenge parameter MUST NOT be present in the authorization response query parameters
+- The error parameter MUST NOT be present in the authorization response query parameters
+- HTTP response status code of token response MUST be 200 OK
+- Content-Type header of token response MUST be application/json
+- Cache-Control header of token response MUST be no-store
+- Token response MUST return a JSON response including access_token and token_type`;
+
+  async run(
+    option: AuthorizationServerOptions,
+    details: Record<string, unknown>
+  ): Promise<ConformanceCheck[]> {
+    try {
+      this.state = randomBytes(32).toString('base64url');
+      const resultMetadata = details[
+        'authorization-server-metadata-endpoint'
+      ] as { body?: Record<string, unknown> };
+      if (!resultMetadata) {
+        throw new Error('Invalid authorization server metadata');
+      }
+      const body = resultMetadata.body;
+
+      const callback = startCallbackServer(option.port);
+      const authorizationRequest = this.buildAuthorizationRequest(body, option);
+      console.log(
+        'Access the following URL in your browser and complete the authentication process.'
+      );
+      console.log(authorizationRequest);
+      console.log('');
+
+      const authorizationResponseUrl = await callback.waitForCallback(300_000);
+
+      const errors: string[] = [];
+      const code = this.validateAuthorizationResponse(
+        authorizationResponseUrl,
+        body,
+        option,
+        errors
+      );
+
+      const tokenResponse = await this.requestToken(body, option, code);
+      if (tokenResponse === null) {
+        return [
+          this.skippedCheck(
+            'Server does not support client_secret_post or client_secret_basic auth methods'
+          )
+        ];
+      }
+      this.validateTokenResponse(tokenResponse, errors);
+
+      if (errors.length > 0) {
+        return [this.failureCheck(errors.join(', '))];
+      }
+
+      return [
+        this.successCheck({
+          authorizationRequest,
+          authorizationResponseUrl,
+          body: tokenResponse.body
+        })
+      ];
+    } catch (error) {
+      return [this.failureCheck(error)];
+    }
+  }
+
+  private buildAuthorizationRequest(
+    metadata: any,
+    option: AuthorizationServerOptions
+  ): string {
+    if (!metadata?.authorization_endpoint) {
+      throw new Error('Unable to obtain authorization endpoint from metadata');
+    }
+
+    const redirectUri = encodeURIComponent(
+      `${REDIRECT_URI_ORIGIN}:${option.port}${REDIRECT_URI_PATH}`
+    );
+    const params =
+      `response_type=code&client_id=${option.clientId}&state=${this.state}` +
+      `&redirect_uri=${redirectUri}&code_challenge=${CODE_CHALLENGE}` +
+      `&code_challenge_method=S256&resource=https%3A%2F%2Fapi.example.com%2Fapp%2F`;
+
+    return `${metadata.authorization_endpoint}?${params}`;
+  }
+
+  private validateAuthorizationResponse(
+    responseUrl: string,
+    metadata: any,
+    option: AuthorizationServerOptions,
+    errors: string[]
+  ): string {
+    const url = new URL(responseUrl);
+
+    if (url.origin !== REDIRECT_URI_ORIGIN + ':' + option.port) {
+      errors.push(`Invalid origin of redirect URL: ${url.origin}`);
+    }
+    if (url.pathname !== REDIRECT_URI_PATH) {
+      errors.push(`Invalid path of redirect URL: ${url.pathname}`);
+    }
+
+    const code = url.searchParams.getAll('code');
+    if (code.length !== 1 || code[0] === '') {
+      throw new Error(`Invalid code parameter: ${code ?? 'missing'}`);
+    }
+
+    const state = url.searchParams.getAll('state');
+    if (state.length !== 1 || state[0] !== this.state) {
+      errors.push(`Invalid state parameter: ${state ?? 'missing'}`);
+    }
+
+    const iss = url.searchParams.getAll('iss');
+    if (iss.length > 0) {
+      if (iss.length !== 1 || iss[0] !== metadata.issuer) {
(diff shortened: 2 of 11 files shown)


src/schemas.ts
```diff
@@ -40,7 +40,14 @@ export type ServerOptions = z.infer<typeof ServerOptionsSchema>;
 
 // Authorization server command options schema
 export const AuthorizationServerOptionsSchema = z.object({
-  url: z.string().url('Invalid authorization server URL')
+  url: z.string().url('Invalid authorization server URL'),
+  clientId: z.string().min(1, 'Client id cannot be empty'),
```
tnorimat, line 44: The schema REQUIRES clientId and secret while CLI requires them as optional parameter, which causes contradiction.
One way for resolving is that the schema requires the parameters as optional.

src/scenarios/authorization-server/authorization-code-grant.ts
```diff
@@ -0,0 +1,264 @@
...
+- The code, state and iss parameters MUST NOT appear more than once
+- The code_challenge parameter MUST NOT be present in the authorization response query parameters
+- The error parameter MUST NOT be present in the authorization response query parameters
+- HTTP response status code of token response MUST be 200 OK
+- Content-Type header of token response MUST be application/json
+- Cache-Control header of token response MUST be no-store
+- Token response MUST return a JSON response including access_token and token_type`;
+
+  async run(
+    option: any,
```
tnorimat: I think it might be better to be type-safe as much as we can.
How about using `AuthorizationServerOptions` defined by  `schemas.ts` ?

src/scenarios/authorization-server/authorization-code-grant.ts
```diff
@@ -0,0 +1,264 @@
...
+import {
+  ClientScenarioForAuthorizationServer,
+  ConformanceCheck,
+  SpecVersion
+} from '../../types';
+import { startCallbackServer } from '../client/auth/helpers/createCallbackServer';
+import { request } from 'undici';
+import { randomBytes } from 'crypto';
+
+const STATE = randomBytes(32).toString('base64url');
```
tnorimat: When only the module is loaded, a pseudo-random number is generated and assigned to `STATE`, which means that multiple runs share that. For now, it might not cause any issue, but in the future, it might cause the issue.

If possible, I think it might be better to generate the value and assign it to `STATE` per run.

src/scenarios/authorization-server/authorization-code-grant.ts
```diff
@@ -0,0 +1,264 @@
...
+const REDIRECT_URI_ORIGIN = 'http://localhost';
+const REDIRECT_URI_PATH = '/callback';
+const CODE_VERIFIER = 'dBjftJeZ4CVP-mB92K27uhbUJU1p1r_wW1gFWFOEjXk';
```
tnorimat, line 17: It seems that `CODE_VERIFIER` and `CODE_CHALLENGE` values are from https://datatracker.ietf.org/doc/html/rfc7636#appendix-B as a valid combination of code_challenge and code_verifier.
How about mentioning that here?

src/scenarios/client/auth/helpers/createCallbackServer.ts
```diff
@@ -0,0 +1,40 @@
...
+export function startCallbackServer(port: number): CallbackServer {
+  const app = express();
+
+  let resolveFn: (url: string) => void;
+
+  const promise = new Promise<string>((resolve) => {
+    resolveFn = resolve;
+  });
+
+  const server = app.listen(port, () => {
```
tnorimat: `app.listen(port)` binds to 0.0.0.0 by default, which means that it listens all IP addresses of the machine running the test. I am afraid that it may cause some security issue.

How about binding explicitly to the loopback address?
```
const server = app.listen(port, '127.0.0.1', () => { ... });
```

src/scenarios/authorization-server/authorization-code-grant.ts
```diff
@@ -0,0 +1,264 @@
...
+    try {
+      const resultMetadata = details[
+        'authorization-server-metadata-endpoint'
+      ] as { body?: Record<string, unknown> };
+      if (!resultMetadata) {
+        throw new Error('Invalid authorization server metadata');
+      }
+      const body = resultMetadata.body;
+
+      const callback = startCallbackServer(option.port);
```
tnorimat, line 58: If any request hits the server on a path other than /callback, express returns a default 404 but the server stays open indefinitely (until timeout).

If possible, it might be good if the server receives a request other than /callback, then it server closes and the test finishes as failure.

src/scenarios/authorization-server/authorization-code-grant.ts
```diff
@@ -0,0 +1,264 @@
...
+  ): Promise<{ body: any; headers: any }> {
+    if (!metadata?.token_endpoint) {
+      throw new Error('Unable to obtain token endpoint from metadata');
+    }
+
+    const params = new URLSearchParams({
+      grant_type: 'authorization_code',
+      code,
+      redirect_uri: REDIRECT_URI_ORIGIN + ':' + option.port + REDIRECT_URI_PATH,
+      client_id: option.clientId,
```
tnorimat: It includes client_id and client_secret are included into HTTP Form body and assumes that a target authorization server supports `client_secret_post` as client authentication method.

I am afraid that if an authorization server does not support `client_secret_post` but support `client_secret_basic`,  then this token request fails.

Therefore, it might be better to firstly look at `token_endpoint_auth_methods_supported `of server metadata and determine which authentication method is used.

src/scenarios/authorization-server/authorization-code-grant.ts
```diff
@@ -0,0 +1,264 @@
...
+    return {
+      id: 'authorization-code-grant',
+      name: 'AuthorizationCodeGrant',
+      description: 'Valid authorization code grant',
+      status: 'SUCCESS',
+      timestamp: new Date().toISOString(),
+      specReferences: [
+        {
+          id: 'Authorization-Code-Grant',
+          url: 'https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-15#section-4.1'
```
tnorimat: I think It might be better to point to MCP spec's reference that mandate https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-15#section-4.1 .

src/scenarios/authorization-server/authorization-code-grant.ts
```diff
@@ -0,0 +1,264 @@
...
+      id: 'authorization-code-grant',
+      name: 'AuthorizationCodeGrant',
+      description: 'Valid authorization code grant',
+      status: 'FAILURE',
+      timestamp: new Date().toISOString(),
+      errorMessage: error instanceof Error ? error.message : String(error),
+      specReferences: [
+        {
+          id: 'Authorization-Code-Grant',
+          url: 'https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-15#section-4.1'
```
tnorimat: I think It might be better to point to MCP spec's reference that mandate https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-15#section-4.1 .

tnorimat (changes requested):

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Also, it might be good to refer OAuth 2.1 v13 instead of v15 because:
- the latest MCP revision, namely 2025-11-25, refers v13.
- other MCP client and server tests refers v13.

tnorimat: Also, how about following the MCP client and server test convention for the way of referring the spec?
E.g,  spec-refercenses.ts:
```
export const SpecReferences: { [key: string]: SpecReference } = {
  RFC_PRM_DISCOVERY: {
    id: 'RFC-9728',
    url: '[https://www.rfc-editor.org/rfc/rfc9728.html#section-3.1'](https://www.rfc-editor.org/rfc/rfc9728.html#section-3.1%27)
  },
```

src/schemas.ts
Michito-Okai: Assuming [this PR](https://github.com/modelcontextprotocol/conformance/pull/228) will be merged, the clientId and secret were optional CLI parameters. I have made clientId and secret mandatory CLI parameters.

src/scenarios/authorization-server/authorization-code-grant.ts
Michito-Okai: I changed it from `any` to `AuthorizationServerOptions`.

src/scenarios/authorization-server/authorization-code-grant.ts
Michito-Okai: I added a comment about `CODE_VERIFIER` and `CODE_CHALLENGE` values.

src/scenarios/authorization-server/authorization-code-grant.ts
Michito-Okai: I modified it so that the value of `state` changes each time `run` is called.

src/scenarios/client/auth/helpers/createCallbackServer.ts
Michito-Okai: I fixed.

src/scenarios/authorization-server/authorization-code-grant.ts
Michito-Okai: I fixed the version of the OAuth 2.1 URL.
Also, I added  `OAUTH_2_1_AUTHORIZATION_CODE_GRANT` to `spec-refercenses.ts`.

src/scenarios/authorization-server/authorization-code-grant.ts
Michito-Okai: I fixed the version of the OAuth 2.1 URL.
Also, I added  `OAUTH_2_1_AUTHORIZATION_CODE_GRANT` to `spec-refercenses.ts`.

src/scenarios/authorization-server/authorization-code-grant.ts
Michito-Okai: The callback accepts any path, and the path check is performed during the `redirect_uri` validation to cause failure if it does not match.

src/scenarios/authorization-server/authorization-code-grant.ts
Michito-Okai: I fixed.
If the authorization server does not support `client_secret_post` and `client_secret_basic`, it is set to `SKIPPED`. The implementations of `client_secret_jwt`, `private_key_jwt`, and `tls_client_auth` are out of scope and have been added to the To do list of [this ISSUE](https://github.com/modelcontextprotocol/conformance/issues/208).

src/schemas.ts
tnorimat: Thank you for the clarification.

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Thank you.

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Thank you.

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Thank you.

src/scenarios/client/auth/helpers/createCallbackServer.ts
tnorimat: Thank you.

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Does it means that the callback receives a request with any path, for example, `/anothercallback` ?

Michito-Okai: That is correct. After receiving the callback, the `redirect_uri` and the callback path are compared, so any path other than /callback will result in failure.

src/scenarios/authorization-server/auth/helpers/createCallbackServer.ts
tnorimat, line 1: It seems that the files under `client` directory should be for MCP client tests. Therefore, it might be good that the file be moved to `authorization-server/auth/helpers` directory.
What do you think about that?

src/scenarios/client/auth/spec-references.ts
```diff
@@ -65,6 +65,10 @@ export const SpecReferences: { [key: string]: SpecReference } = {
     id: 'RFC-7523-JWT-Client-Auth',
     url: 'https://datatracker.ietf.org/doc/html/rfc7523#section-2.2'
   },
+  OAUTH_2_1_AUTHORIZATION_CODE_GRANT: {
```
tnorimat: It seems that the files under `client` directory should be for MCP client tests. Therefore, it might be good that the file be moved to `authorization-server/auth/spec-references.ts` file.
What do you think about that?

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Thank you.

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Thank you.

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: It might be better to comment on the source file to notify the reader of the codes of why the codes does not accept an authorization server supporting client_secret_jwt, private_key_jwt, and tls_client_auth.

An authorization server supporting client_secret_jwt, private_key_jwt, and tls_client_auth is valid one. Therefore the tests should have accept such the authorization server but the PR would not accept that because supporting each client_secret_jwt, private_key_jwt, and tls_client_auth client authentication method by the tests requires us to implement a lot of codes so it is better to be handled by other follow-up PRs.

tnorimat (changes requested):

src/scenarios/client/auth/spec-references.ts
Michito-Okai: I added `authorization-server/auth/spec-references.ts` and removed `OAUTH_2_1_AUTHORIZATION_CODE_GRANT` in `client/auth/spec-references.ts`.

src/scenarios/authorization-server/auth/helpers/createCallbackServer.ts
Michito-Okai: I added `authorization-server/auth/helpers/createCallbackServer.ts` and removed `client/auth/helpers/createCallbackServer.ts`.

src/scenarios/authorization-server/authorization-code-grant.ts
Michito-Okai: I added a comment.

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Thank you for the clarification.

src/scenarios/authorization-server/authorization-code-grant.ts
tnorimat: Thank you.

src/scenarios/authorization-server/auth/helpers/createCallbackServer.ts
tnorimat: Thank you.

src/scenarios/client/auth/spec-references.ts
tnorimat: Thank you.

tnorimat (approved):
LGTM

Issue #223, opened 2026-04-10
Workload Identity Federation (SEP-1933) - Client Conformance


PieterKas:
## Overview

Add conformance tests for MCP clients that authenticate using Workload 
Identity Federation as described in SEP-1933 
(https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1933).

This mechanism allows autonomous workloads (e.g. running on Kubernetes, 
SPIFFE/SPIRE, or cloud runtimes) to exchange a platform-issued JWT for an 
MCP access token, using the JWT Bearer grant defined in RFC 7523 
(https://www.rfc-editor.org/rfc/rfc7523).

No client registration is required. Trust is established at the issuer 
level — the authorization server maintains a trusted issuer allowlist and 
resolves signing keys dynamically via OIDC Discovery. Key resolution is 
an authorization server responsibility and is out of scope for client 
conformance testing.

## Specification References

- SEP-1933 — Workload Identity Federation for MCP:
  https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1933
- RFC 7523 — JWT Profile for OAuth 2.0 Authorization Grants:
  https://www.rfc-editor.org/rfc/rfc7523
- RFC 7521 — Assertion Framework for OAuth 2.0 Client Authentication and 
  Authorization Grants:
  https://www.rfc-editor.org/rfc/rfc7521
- RFC 6749 — The OAuth 2.0 Authorization Framework (Section 5.2 error 
  codes):
  https://www.rfc-editor.org/rfc/rfc6749#section-5.2

## Scope

This issue covers **client conformance** only. From the client's 
perspective, it receives a pre-issued JWT assertion (via scenario context) 
and must present it correctly to the token endpoint. The test authorization 
server accepts the assertion at face value and issues an access token — 
testing the authorization server's ability to perform OIDC Discovery, key 
validation, and enforcing specific policies is explicitly out of scope 
and belongs in authorization server conformance tests.

From the MCP server (resource server) perspective, the grant type used to 
obtain an access token does not affect how the resource server processes 
requests — no changes to `createServer` are required.

### 1. `createAuthServer` updates
- Accept `urn:ietf:params:oauth:grant-type:jwt-bearer` at the token 
  endpoint
- Advertise the grant type in `grant_types_supported` on the 
  `/.well-known/oauth-authorization-server` metadata endpoint
- Support configurable rejection rules (e.g. `rejectedSubjects`, 
[...]

PR #268, opened 2026-05-12
feat(auth): client conformance for Workload Identity Federation (SEP-1933)

coding-bobo:
#### Motivation

As autonomous workloads increasingly interact with MCP servers, the existing authorization mechanisms - primarily designed around human-centric OAuth flows or statically registered client credentials - impose operational friction and do not align with how modern platforms provision identity. Workloads running under Kubernetes, SPIFFE/SPIRE, or cloud-native runtimes already receive short-lived, cryptographically verifiable JWTs that reflect their runtime identity. Requiring those workloads to additionally register as OAuth clients or manage long-lived secrets adds complexity and weakens security posture.

SEP-1933 addresses this by letting MCP directly leverage platform-provided credentials via the JWT Bearer grant (RFC 7523), eliminating a separate client identity lifecycle while preserving strong authentication guarantees. This PR adds **client conformance coverage** for that flow, so any MCP client that claims SEP-1933 support can be checked against the same behavioural contract regardless of the underlying workload identity platform.

#### Scope

Conformance tests are scoped to **client behaviour**. From the client's perspective, it receives a pre-issued JWT assertion (via scenario context) and must present it correctly at the token endpoint. OIDC discovery, signing-key resolution, issuer-allowlist policy, and `jti` replay detection are authorization-server concerns and are deliberately deferred to AS conformance work - they don't exercise any new client code path.

#### What's included

**Signing helper**

- `helpers/createWorkloadJwt.ts`: `generateWorkloadKeypair(alg?)` and `createWorkloadJwt(opts)`, plus `JWT_BEARER_GRANT_TYPE` and `DEFAULT_WORKLOAD_JWT_ALG` constants. Keypair generation is intentionally separated from signing so the scenario can sign with a key the AS does not trust without API gymnastics. `expiresIn` accepts either a duration string or an absolute epoch number, so already-expired assertions can be constructed deterministically for negative tests.
- `helpers/createWorkloadJwt.test.ts`: 13 unit tests (round-trip sign+verify, default lifetime, expired-token construction via numeric epoch, unique `jti`, array audience preservation per RFC 7519 §4.1.3, reserved-claim protection, caller-supplied `jti`/`notBefore`, algorithm override).
- `cross-app-access.ts`: two bare URN literals replaced with `JWT_BEARER_GRANT_TYPE`. No behaviour change.

API shape mirrors existing patterns in `client-credentials.ts` (keypair + PEM export) and `cross-app-access.ts` (SignJWT chain).

**Scenario and clients**

- `src/scenarios/client/auth/wif-jwt-bearer.ts`: generates an ES256 keypair per `start()`, pre-signs three JWT variants (`valid_jwt`, `wrong_audience_jwt`, `expired_jwt`) with neutral fixture values, passes them via context, and configures the conformance AS with `grant_types: [jwt-bearer]` and `token_endpoint_auth_method: none`. DCR is disabled and the client is pre-seeded with `WIF_CLIENT_ID`, matching the SEP-1933 model where trust is established at the issuer level rather than per-client.
- `src/scenarios/client/auth/spec-references.ts`: `RFC_7523_JWT_BEARER` and `SEP_1933_WIF` references.
- `src/seps/sep-1933.yaml`: requirement traceability.
- `src/schemas/context.ts`: `auth/wif-jwt-bearer` context schema.
- `examples/clients/typescript/everything-client.ts`: `WifJwtBearerProvider`, `runWifJwtBearer`, and six broken-client runners.
- `examples/clients/typescript/auth-test-wif-*.ts`: CLI entrypoints for each broken variant.
- `src/scenarios/client/auth/index.ts`: scenario registration in `draftScenariosList`.
- `src/scenarios/client/auth/index.test.ts`: happy-path via draft loop + 6 negative tests.

#### Checks implemented (9)

| Check ID | What it verifies | Severity |
|---|---|---|
| `wif-grant-type` | Client sends `grant_type=urn:ietf:params:oauth:grant-type:jwt-bearer` | FAILURE |
| `wif-assertion-missing` | Client includes the `assertion` parameter | FAILURE |
| `wif-assertion-verified` | JWT signature, audience, and expiry are valid | FAILURE |
| `wif-assertion-expired` | Client surfaces `invalid_grant` for an expired assertion | FAILURE |
| `wif-assertion-audience` | Client surfaces `invalid_grant` for a wrong-audience assertion | FAILURE |
| `wif-assertion-malformed` | Client surfaces `invalid_grant` for a bad-signature assertion | FAILURE |
| `wif-no-retry` | Client does not retry after a JWT-bearer failure | WARNING |
| `wif-assertion-scope-rejected` | Client surfaces `invalid_scope` and does not retry | WARNING |
| `wif-grant-fallback` | Client does not fall back to `authorization_code` after `unauthorized_client` | WARNING |

`wif-no-retry`, `wif-grant-fallback`, and `wif-assertion-scope-rejected` are WARNING because RFC 7523 is silent on client retry and grant-type switching behaviour; they will be upgraded to FAILURE once SEP-1933 lands normative spec text.

Three requirements are explicitly deferred in `src/seps/sep-1933.yaml` (`iss`, `sub`, `jti`); all are AS policy decisions. The scenario does not expose `iss` or `sub` in context because the client presents a pre-signed token and has no mechanism to vary those claims.

#### Coverage vs. issue #223

Issue #223 lists seven negative checks. This PR covers all seven, plus additional client-observable checks:

- Directly from the issue: `wif-assertion-expired` (expired `exp`), `wif-assertion-audience` (wrong `aud`), `wif-assertion-missing` (missing assertion), `wif-assertion-scope-rejected` (`invalid_scope`), `wif-grant-fallback` (`unauthorized_client` without grant-type fallback).
- Added beyond the issue: `wif-grant-type`, `wif-assertion-malformed`, `wif-no-retry`, `wif-assertion-verified`.

The remaining three issue checks (`invalid_grant`/unauthorised `sub`, `invalid_grant`/replayed `jti`, issuer allowlist) exercise AS policy rather than client code paths and are deferred to AS conformance work.

#### Testing

- [x] `npx vitest run src/scenarios/client/auth/helpers/createWorkloadJwt.test.ts` - 13 tests pass
- [x] `npx vitest run src/scenarios/client/auth/index.test.ts` - 49 tests pass (6 WIF negative tests)
- [x] `npx vitest run` - full suite passes (194 tests, 16 files, no regressions)
- [x] `npm run check` - typecheck and lint clean

#### Closes / relates

- Closes `modelcontextprotocol/conformance#223`
- Implements client-side reference test coverage for `modelcontextprotocol/modelcontextprotocol#1933`


examples/clients/typescript/auth-test-wif-expired-assertion.ts
```diff
@@ -0,0 +1,15 @@
...
+/**
+ * Broken WIF client: presents a JWT that is already expired.
+ * BUG: Uses expired_jwt instead of valid_jwt — server rejects with invalid_grant.
+ */
+
+import { runWifJwtBearerExpiredAssertion } from './everything-client.js';
+import { runAsCli } from './helpers/cliRunner.js';
+
+runAsCli(
+  runWifJwtBearerExpiredAssertion,
```
pcarleton, line 12: here and below: can you put the broken implementation in the negative test file? we're keeping `everything-client` as the "good" client, and broken implementations (mostly) self-contained (though it's okay to share helpers)

src/scenarios/client/auth/wif-jwt-bearer.ts
```diff
@@ -0,0 +1,316 @@
...
+const WIF_ISSUER = 'https://idp.conformance-test.local';
+const WIF_SUBJECT = 'conformance-workload';
+const WIF_CLIENT_ID = 'conformance-wif-workload';
+
+export class WifJwtBearerScenario implements Scenario {
+  name = 'auth/wif-jwt-bearer';
+  specVersions: SpecVersion[] = [DRAFT_PROTOCOL_VERSION];
+  readonly source = { introducedIn: DRAFT_PROTOCOL_VERSION } as const;
+  description =
+    'Tests OAuth JWT-bearer grant (RFC 7523 §2.1) for workload identity federation (SEP-1933)';
```
pcarleton: please add enough description here so that a client implementer can tell what they have to implement in order to pass this test.

src/scenarios/client/auth/wif-jwt-bearer.ts
```diff
@@ -0,0 +1,316 @@
...
+      }
+    };
+  }
+
+  async stop() {
+    await this.authServer.stop();
+    await this.server.stop();
+  }
+
+  getChecks(): ConformanceCheck[] {
```
pcarleton, line 305: i think you'll want to add a sentinel failed check here for `!tokenRequestReceived` in order to prevent passing if the client never sends a request. there should be other examples of that pattern in here.

src/scenarios/client/auth/wif-jwt-bearer.ts
```diff
@@ -0,0 +1,316 @@
...
+          if (scopeList.includes(WIF_TRIGGER_UNAUTHORIZED_SCOPE)) {
+            this.failedOnce = true;
+            return {
+              error: 'unauthorized_client',
+              errorDescription: 'Client not authorized for JWT-bearer grant'
+            };
+          }
+          if (scopeList.includes(WIF_REJECTED_SCOPE)) {
+            this.checks.push({
+              id: 'wif-assertion-scope-rejected',
```
pcarleton: i think this is redundant with `wif-no-retry` -- this triggers when the bad scope is rejected, but the behavior you're checking is that the client doesn't retry.  it'd be fine to have `wif-no-retry` referenced in 2 rows in the sep-1933.yaml

src/schemas/context.ts
```diff
@@ -31,6 +31,16 @@ export const ClientConformanceContextSchema = z.discriminatedUnion('name', [
     idp_id_token: z.string(),
     idp_issuer: z.string(),
     idp_token_endpoint: z.string()
+  }),
+  z.object({
+    name: z.literal('auth/wif-jwt-bearer'),
+    client_id: z.string(),
+    // RFC 7523 does not require aud to be a URL, but this scenario's test AS
+    // is always addressed by URL, so the constraint is intentional here.
+    audience: z.string().url(),
```
pcarleton: i don't think this is read anywhere, we just use auth server URL instead, suggest removing this

src/scenarios/client/auth/helpers/createWorkloadJwt.ts
```diff
@@ -0,0 +1,86 @@
...
+export const WIF_TRIGGER_UNAUTHORIZED_SCOPE = 'wif.trigger-unauthorized';
+export const WIF_REJECTED_SCOPE = 'wif.rejected';
+
+export interface CreateWorkloadJwtOptions {
+  issuer: string;
+  subject: string;
+  audience: string | string[];
+  privateKey: jose.CryptoKey;
+  /** Jose duration string (e.g. '5m') or absolute epoch seconds. Use a number to construct already-expired tokens for negative tests. */
+  expiresIn?: string | number;
```
pcarleton: i think removing the number type here would be clearer, if you want an expired one, pass a string for a date in the past.

src/scenarios/client/auth/wif-jwt-bearer.ts
```diff
@@ -0,0 +1,316 @@
...
+      disableDynamicRegistration: true,
+      onTokenRequest: async ({ grantType, body, timestamp, authBaseUrl }) => {
+        // wif-no-retry and wif-grant-fallback fire on any second request after
+        // any first failure, not only after unauthorized_client specifically.
+        if (this.tokenRequestReceived && this.failedOnce) {
+          if (grantType !== JWT_BEARER_GRANT_TYPE) {
+            this.checks.push({
+              id: 'wif-grant-fallback',
+              name: 'WifGrantFallback',
+              description: `Client fell back to ${grantType} grant after receiving unauthorized_client; clients should not switch grant types after a JWT-bearer failure`,
```
pcarleton:
```suggestion
              description: `Client fell back to ${grantType} grant after a JWT-bearer token request was rejected; clients should not switch grant types after a JWT-bearer failure`,
```
we don't know if it got unauthorized_client for sure here

pcarleton (changes requested):
thanks for sending this

coding-bobo:
Thank you so much for the review Paul!
I fixed all comments.

pcarleton:
LGTM! merging so we have a WIF surface to iterate on.

One thing to clean up in a follow-up: the scope-rejected path is currently a no-op. The AS branch returns invalid_scope without pushing a check:

https://github.com/modelcontextprotocol/conformance/blob/620b211f4030ffb62f8c866f91356b38145bc4b8/src/scenarios/client/auth/wif-jwt-bearer.ts#L173-L180

…and its negative test has no expectedFailureSlugs, so it passes trivially:

https://github.com/modelcontextprotocol/conformance/blob/620b211f4030ffb62f8c866f91356b38145bc4b8/src/scenarios/client/auth/index.test.ts#L319-L324

The SDK doesn't retry on invalid_scope, so it can't fold into wif-no-retry either. Net effect: the PR ships 8 checks, not 9  either drop WifScopeRejectedProvider / auth-test-wif-scope-rejected.ts / the test, or wire an actual check if there's a client-observable behaviour to pin it to.

Also FYI: the ext-auth WIF draft says aud MUST be the token-endpoint URL, but this scenario (correctly) uses the AS issuer identifier per draft-ietf-oauth-rfc7523bis. We can fix SEP-1933 to match, no change needed here.

pcarleton (approved):

PR #270, opened 2026-05-12
feat: add client conformance tests for SEP-2575

anubhav756:
Adds client-side conformance tests for stateless MCP [SEP-2575](https://modelcontextprotocol.io/seps/2575-stateless-mcp#sep-2575-make-mcp-stateless).

Tests are fully aligned with the official traceability requirements in [PR #273](https://github.com/modelcontextprotocol/conformance/pull/273) (`src/seps/sep-2575.yaml`).

> [!NOTE]
> This is a companion PR to the server-side tests in [#271](https://github.com/modelcontextprotocol/conformance/pull/271).

---

## Context
SEP-2575 removes the stateful `initialize` handshake and requires carrying protocol version, implementation info, and capabilities on a per-request basis in the `_meta` parameters and HTTP headers. 

This PR introduces the **`request-metadata`** client-side conformance scenario to verify that client implementations correctly follow these new stateless rules, incorporating all feedback from review #270.

---

### 1. The `request-metadata` Client Scenario
Adds the new `request-metadata` scenario to verify per-request version headers, matching `_meta` payload specifications, and client capability obligations:
* **Optional Capability Conditional Checks**:
  * Roots, sampling, and elicitation are optional client capabilities. A specialized client is 100% conformant even if it only implements the `tools` capability and omits the others.
  * **Mock Server (`src/scenarios/client/request-metadata.ts`)**:
    * **If present**: verifies they are formatted as valid JSON objects `{}` (`SUCCESS`/`FAILURE`).
    * **If absent**: records **`SKIPPED`** (vacuously conformant).
* **Targeted Reference Client Assertions**:
  * Inside the unit tests, we explicitly assert that the comprehensive reference client (`everything-client.ts`) declarations resolve to **`SUCCESS`** (not `SKIPPED`). This guarantees that we actively test the happy path of all optional capabilities in our reference implementation while keeping optionals skipped-friendly for specialized SDKs.

### 2. Simulated Version Negotiation Retry
The mock server simulates a version negotiation flow to assert the `ClientRetrySupportedVersion` check (severity `WARNING` / `expectedFailureSlugs` compatible):
* On the first request, the mock server deliberately rejects it with `400 Bad Request` and `UnsupportedProtocolVersionError` (code `-32001`) carrying `data: { supported: ['DRAFT-2026-v1'] }`.
* The client must successfully negotiate and retry using the mutually supported version, updating the check status to `SUCCESS`.

---

## How Has This Been Tested?

All 133 unit tests are fully green, warning-free, and compile cleanly using `npm test` and `npm run check`.

#### 1. TypeScript Reference Client
```bash
npm start -- client --spec-version draft --scenario request-metadata --command "npx tsx examples/clients/typescript/everything-client.ts"
```
**Result**: **Passed: 7/7**, 0 failed, 0 warnings (**✅ OVERALL: PASSED**)
* Correctly asserts `SUCCESS` for all optional capabilities and version negotiation retry checks.

### 32 Negative Unit Tests (`request-metadata.test.ts`)
Added extensive negative cases verifying compliance checks correctly trap clients that:
* Omit the `_meta` parameter (`sep-2575-client-populates-meta` $\rightarrow$ `FAILURE`).
* Omit the `MCP-Protocol-Version` HTTP header (`sep-2575-http-client-sends-version-header` $\rightarrow$ `FAILURE`).
* Disagree between headers and `_meta.protocolVersion` (`sep-2575-http-version-header-matches-meta` $\rightarrow$ `FAILURE`).
* Fail to negotiate or retry on `400` rejections (`sep-2575-client-retry-supported-version` $\rightarrow$ `WARNING`).
* Exit cleanly without looping/hanging on empty version intersections.
* Send invalid (non-object) capabilities (`roots: "string"` $\rightarrow$ `FAILURE`).

---

## Breaking Changes
None.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context
None.

### DIFF
--- a/src/seps/sep-2575.yaml
+++ b/src/seps/sep-2575.yaml
@@ -1,0 +1,104 @@
+sep: 2575
+spec_url: https://modelcontextprotocol.io/specification/draft/basic/lifecycle
+requirements:
+  - check: sep-2575-client-populates-meta
+    text: 'Every client request MUST include the following io.modelcontextprotocol/* fields in _meta: protocolVersion, clientInfo, clientCapabilities.'
+    url: https://modelcontextprotocol.io/specification/draft/basic/index#meta
+  - check: sep-2575-server-rejects-undeclared-capability
+    text: 'A server MUST NOT rely on capabilities the client has not declared. If processing a request requires a capability the client did not include in io.modelcontextprotocol/clientCapabilities, the server MUST return a MissingRequiredClientCapabilityError (-32003).'
+    url: https://modelcontextprotocol.io/specification/draft/basic/index#meta
+  - check: sep-2575-missing-capability-http-400
+    text: 'On HTTP, the response status MUST be 400 Bad Request [for MissingRequiredClientCapabilityError].'
+    url: https://modelcontextprotocol.io/specification/draft/basic/index#meta
+  - check: sep-2575-server-tags-subscription-id
+    text: 'On notifications delivered via a subscriptions/listen stream, the server MUST include io.modelcontextprotocol/subscriptionId in _meta so the client can correlate the notification with the originating subscription request.'
+    url: https://modelcontextprotocol.io/specification/draft/basic/index#meta
+  - check: sep-2575-server-stateless-no-prior-context
+    text: 'A server MUST NOT treat connection or process identity as a proxy for conversation or session continuity. / Servers MUST NOT rely on prior requests over the same connection to establish context (e.g., capabilities, protocol version, client identity).'
+  - check: sep-2575-server-stateless-no-connection-reuse-required
+    text: 'Servers MUST NOT require that a client reuse the same connection to perform related operations.'
+  - check: sep-2575-server-unsupported-version-error
+    text: 'If the server does not implement the requested version (whether the version is unknown to the server, or is a known version the server has chosen not to support), it MUST respond with an UnsupportedProtocolVersionError listing the versions it does support.'
+    url: https://modelcontextprotocol.io/specification/draft/basic/lifecycle#protocol-version-negotiation
+  - check: sep-2575-client-retry-supported-version
+    text: 'The client SHOULD select a mutually supported version from the supported list and retry the request, or surface an error to the user if no compatible version exists.'
+    url: https://modelcontextprotocol.io/specification/draft/basic/lifecycle#protocol-version-negotiation
+  - check: sep-2575-server-implements-discover
+    text: 'Servers MUST implement server/discover.'
+    url: https://modelcontextprotocol.io/specification/draft/server/discover
+  - check: sep-2575-http-server-no-independent-requests-on-stream
+    text: 'The server MUST NOT send independent JSON-RPC requests on this stream. Server-to-client interactions are embedded as input requests inside an IncompleteResult.'
+    url: https://modelcontextprotocol.io/specification/draft/basic/transports#receiving-messages-1
+  - check: sep-2575-http-server-disconnect-is-cancel
+    text: 'Closing the SSE response stream MUST be treated by the server as cancellation of that request.'
+    url: https://modelcontextprotocol.io/specification/draft/basic/transports#cancellation-1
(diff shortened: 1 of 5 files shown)


pcarleton (commented):
Thanks for tackling this. SEP-2575 is a big one. I ran `/new-sep 2575` to build the full requirement-traceability YAML and opened it as **#273** so we have a shared reference for both this PR and #271.

**Coverage:** the YAML has 12 client-side `check:` rows; this PR currently covers 3 (`client-populates-meta`, the version-header pair). The 5 stdio-only rows can be a follow-up scenario, but the 3 `client-declares-{elicitation,roots,sampling}-capability` MUSTs and the `client-retry-supported-version` SHOULD are HTTP-reachable and would fit nicely here.

**Correctness bits:**

- **`client-consistent-version`** — I couldn't find spec backing for this. SEP-2575 makes each request self-contained; nothing forbids a client from changing `protocolVersion` between requests. The `flippingVersionClient` negative test would fail a conformant client. Suggest dropping.
- **`client-cancels-by-notification`** — spec scopes `notifications/cancelled` to **stdio only** ("Streamable HTTP: …No `notifications/cancelled` message is required or expected"). This scenario is HTTP, so the check can't legitimately fire here.
- **`server/discover` gap** — the example's discover call (everything-client.ts:98) sends no `_meta` and no `MCP-Protocol-Version` header, but the harness marks it as passing. `client-sends-version-header` is guarded by `if (currentVersion)` and `client-populates-meta` runs after the discover early-return. I think we want the example to send the header and the check order to catch it if an example doesn't do it.
- **Check IDs** — we're shifting to check ids like `sep-<NNNN>-<slug>` (see `sep-2164-error-code` in resources.ts); renaming to match #273's slugs will let the traceability tooling line them up automatically.
- **`specReferences` URLs** are empty — #273 has the per-row anchors you can lift.

pcarleton (changes requested):
(See above)

pcarleton:
Pushed the review fixes as a stacked PR you can merge into this branch: https://github.com/anubhav756/conformance/pull/1 (one commit per item above, plus the `sep-2575.yaml` traceability file and a rename to `request-metadata`).

src/seps/sep-2575.yaml
```diff
@@ -0,0 +1,104 @@
...
+  - check: sep-2575-server-sends-tools-list-changed-on-subscription
+    text: '[A server with the listChanged] capability SHOULD send a notification to clients that have opened a subscriptions/listen stream with toolsListChanged: true.'
+    url: https://modelcontextprotocol.io/specification/draft/server/tools#list-changed-notification
+  - check: sep-2575-server-no-log-without-loglevel
+    text: 'The server MUST NOT emit notifications/message for a request that does not include [io.modelcontextprotocol/logLevel in _meta].'
+    url: https://modelcontextprotocol.io/specification/draft/server/utilities/logging#per-request-log-level
+
+  - text: 'State that needs to span multiple requests (e.g., long-running tasks, application-level handles) MUST be referenced by an explicit identifier the client passes on each request.'
+    excluded: 'architectural guidance, observable only via subscriptionId/task-id rows already listed'
+  - text: 'To distinguish notifications belonging to different concurrent subscriptions, clients MUST correlate notifications using the io.modelcontextprotocol/subscriptionId field carried in _meta.'
```
pcarleton, line 83: (for future) reading this one again, we may want to have a concurrent subscription test, or a test for what to do if you get back a subscriptionId for something you didn't subscribe to

pcarleton (approved):
This looks great thanks.

I've been thinking about how we navigate these "stateless" tests that theoretically should be true on *every* request, not just a standalone scenario.  However, that will require us to revisit all the existing tests.

The strategy I think makes the most sense is for us to get our traceability coverage in for the full RC, and then refactor tests / revisit older tests to make them branch-able on version id, and then make sure we still have check coverage.

No action needed on this PR, just wanted to mention that there will likely be a refactor coming in the near future that may change this shape a bit.

Issue #266, opened 2026-05-12
Add SEP-2575 conformance tests


Yuan325:
Add conformance tests for SEP-2575 to validate the stateless-MCP behavior. This includes conformance tests for both MCP server and client.

# Server
1. Per-request _meta
    * Rejects request with missing _meta → -32602 Invalid params.
    * Rejects request with _meta missing io.modelcontextprotocol/protocolVersion.
    * Rejects request with _meta missing io.modelcontextprotocol/clientInfo.
    * Rejects request with _meta missing io.modelcontextprotocol/clientCapabilities.
2. server/discover
    * Responds to server/discover
    * Response includes required fields: supportedVersions (non-empty), capabilities, serverInfo
    * capabilities matches what the server honors on real RPC calls
3. Version negotiation
    * For a _meta.protocolVersion the server does not support, returns UnsupportedProtocolVersionError (not Invalid params and not silent rejection).
    * UnsupportedProtocolVersionError carries non-empty data.supportedVersions matching the set advertised by server/discover (no drift).
    * server/discover's supportedVersions matches the set the server actually accepts on real RPC calls (no drift).
    * Returns the error even when the unsupported version is on the server/discover request itself (and still advertises supportedVersions so the client can recover).
    * [HTTP] UnsupportedProtocolVersionError → HTTP 400 Bad Request.
    * [HTTP] Rejects request when MCP-Protocol-Version header is absent or does not match _meta.protocolVersion (HTTP 400).
4. Errors
    * MissingRequiredClientCapabilityError (-32003) carries data.requiredCapabilities listing missing keys.
    * All error responses carry the request's JSON-RPC id.
5. Removed RPCs
    * ping → -32601.
    * logging/setLevel → -32601.
    * resources/subscribe → -32601.
    * resources/unsubscribe → -32601.
6. HTTP transport
    * Unknown method → HTTP 404 Not Found + JSON-RPC -32601.
    * MissingRequiredClientCapabilityError → HTTP 400 Bad Request.

# Client
1. Per-request _meta
    * Populates _meta on every request with all three required fields.
    * Accurately reports clientCapabilities per request (does not claim unsupported features).
2. server/discover
    * Should be able to call 
3. Version negotiation
    * Does version negotiation correctly 
    * Includes io.modelcontextprotocol/protocolVersion in _meta on every request.
    * Sends a consistent protocolVersion once chosen (does not flip between versions request-to-request).
[...]

PR #271, opened 2026-05-12
feat: add server conformance tests for SEP-2575

Yuan325:
Add conformance tests for [SEP-2575](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575) to validate the stateless-MCP behavior. The checks was checked against https://github.com/modelcontextprotocol/conformance/pull/273

## Motivation and Context
Ensure that servers adhere to the SEP.

## How Has This Been Tested?
Tested locally

## Breaking Changes
n/a. This tests are for the draft specs.

## Types of changes

- [ ] Documentation update

## Checklist

server conformance tests for #266

### DIFF
--- a/src/scenarios/server/stateless.ts
+++ b/src/scenarios/server/stateless.ts
@@ -1,0 +1,749 @@
+/**
+ * Stateless MCP test scenarios for MCP servers (SEP-2575)
+ */
+
+import {
+  ClientScenario,
+  ConformanceCheck,
+  SpecVersion,
+  DRAFT_PROTOCOL_VERSION
+} from '../../types';
+
+const SPEC_REF = [
+  {
+    id: 'SEP-2575',
+    url: 'https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2575'
+  }
+];
+
+export class ServerStatelessScenario implements ClientScenario {
+  name = 'server-stateless';
+  specVersions: SpecVersion[] = [DRAFT_PROTOCOL_VERSION];
+  description = `Test stateless MCP server architecture (SEP-2575).
+
+**Server Implementation Requirements:**
+
+**Endpoints**:
+- \`server/discover\`: Returns supportedVersions, capabilities, and serverInfo.
+- \`tools/call\`: Implement tool \`test_missing_capability\` requiring \`sampling\` capability in \`_meta\`.
+
+**Requirements**:
+1. **Per-request _meta**: Rejects requests missing \`_meta\` or its required subfields (\`io.modelcontextprotocol/protocolVersion\`, \`io.modelcontextprotocol/clientInfo\`, \`io.modelcontextprotocol/clientCapabilities\`) with -32602 Invalid params.
+2. **server/discover**: Returns valid discovery metadata matching real RPC capabilities.
+3. **Version negotiation**: For unsupported versions, returns UnsupportedProtocolVersionError (HTTP 400, non-empty supportedVersions matching discover). Validates \`MCP-Protocol-Version\` header.
+4. **Errors**: MissingRequiredClientCapabilityError (-32003, HTTP 400) lists missing capabilities. All error responses carry matching JSON-RPC id.
+5. **Removed RPCs**: Rejects \`initialize\`, \`ping\`, \`logging/setLevel\`, \`resources/subscribe\`, \`resources/unsubscribe\` with -32601.
+6. **HTTP transport**: Unknown methods return HTTP 404 + JSON-RPC -32601.`;
+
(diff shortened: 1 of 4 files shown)


pcarleton (commented):
Thanks for this — covers a good chunk of the version-negotiation and discover surface. (reminder the full SEP-2575 traceability YAML in **#273** as a shared reference for this and #270.)

**Coverage:** the YAML has 20 server-side `check:` rows; this PR covers 7 (discover, version-error, 400-on-unsupported, header-mismatch, 404-on-unknown, missing-capability ×2). The big uncovered areas are `subscriptions/listen` (4 rows), HTTP cancellation (3 rows - 2 HTTP, 1 stdio), the core statelessness invariants (no-prior-context / no-connection-reuse), and `server-no-log-without-loglevel`. Fine to land those as follow-ups, but worth flagging which are intentionally deferred.

**Correctness:**

- **MUST→WARNING severity** — `stateless-error-missing-capability` and `stateless-http-missing-capability` use `WARNING` on failure, but the spec text is **MUST** for both `-32003` and HTTP 400. If the worry is the test tool not existing, skip the check rather than downgrade.
- **`stateless-version-response-header` is dead** — only ever pushed in a `catch` block (stateless.ts:360), never in the happy path. Also no spec backing: there's no server-MUST-return-header. Suggest dropping it.
- **Registered twice** in `index.ts` (both `pendingClientScenariosList` and `allClientScenariosList`). We should clean this up, but you can put it in `all` and we filter it by spec version tag.
- **"Absent header" test isn't absent** — `{'MCP-Protocol-Version': ''}` sets an empty-string header; the helper always injects the default, so the header is never truly omitted. Needs a delete-key path in `sendRpc`.
- **`stateless-version-header-mismatch`** only checks HTTP 400; spec says it SHOULD include `-32001 HeaderMismatch` in the body.
- **`stateless-removed-*` (5 checks)** — there's no normative MUST forbidding `ping`/`initialize`; the only relevant text is the generic *"unknown method → 404 + -32601"*. These checks don't assert the 404, and the example returns **200** for them. Suggest folding into one `sep-2575-http-server-method-not-found-404` slug parametrized by method.
- **`stateless-meta-missing*` (4 checks)** — the draft spec pages only put the MUST on the client (the SEP doc has a server-MUST-reject, but it didn't make it into basic/index.mdx — possibly a spec bug to fix). Either way, one check slug with the field name in details is probably better here.
- **`noDrift`** requires identical array order between discover and error; spec doesn't mandate order — set comparison.
- **Check IDs** — should be `sep-2575-*` to match #273 (same note left on #270); `specReferences` should point to spec anchors not the SEP PR.

**Structure:** ~60% of the 749 lines is duplicated catch-block boilerplate re-pushing `id`/`name`/`description`. A `runCheck(id, desc, fn)` helper would cut it to ~250 lines and prevent the try/catch description drift (e.g. `stateless-version-unsupported` has different descriptions in try vs catch).

Negative test could also be tightened — `expect(failures.length).toBeGreaterThan(0)` doesn't pin which checks fail; #270 uses `expectedFailureSlugs` which is the pattern we want.

Yuan325:
Thankyou for the thorough review @pcarleton , I've updated the test cases based on your review and added new checks according to #273. For the client capability checks, it should only check for payload structure and http status if it detects -32003. These tests were also checked against Toolbox's implementation of SEP-2575 (https://github.com/googleapis/mcp-toolbox/pull/3211). Thank you!

src/scenarios/server/stateless.ts
```diff
@@ -347,56 +339,8 @@ export class ServerStatelessScenario implements ClientScenario {
       }
     );
 
-    await runCheck(
```
pcarleton, line 350: @Yuan325 I deleted these placeholder checks. I want to either have the checks verify things or be marked as excluded if they can't be tested

pcarleton:
Going to merge this in, but note that several of the SEP-2575 server requirements still need tests (The blank always-pass tests didn't seem useful).  I think many of those are actually testable so I don't think we need to mark them as excluded, but let's look at that separately from this PR.

Yuan325:
@pcarleton Can you please clarify what are the blank always-pass tests? I can take a look at those test cases that we want to add to SEP-2575. Is it the ones that are removed from https://github.com/modelcontextprotocol/conformance/pull/271/changes/568718ab98f9fbaf86488e57adeb1a42be09f883?

Thank you!

pcarleton:
@Yuan325 Yes, those tests did not check anything and returned `PASS`

Yuan325:
@pcarleton Got it, i will submit another PR with the updated tests for those cases. Meanwhile, can you PTAL at this: https://github.com/modelcontextprotocol/conformance/pull/293? Thank you again!

Issue #258, opened 2026-04-29
Add support for a stdio testing


remyleone:
**Is your feature request related to a problem? Please describe.**

It would be helpful when I'm running a MCP server in stdio mode to have the option to check that the conformance tests are passing.

**Describe the solution you'd like**

npx @modelcontextprotocol/conformance server --stdio ./my_server

**Describe alternatives you've considered**

I've tried the run a server in a SSE fashion using the examples of official go-sdk but it didn't worked: 
- https://github.com/modelcontextprotocol/go-sdk/tree/main/examples/server/sse

**Additional context**

It would be helpful to have consistant and conformance friendly examples in all SDKs. Could you ensure that the examples in the official documentation are passing the conformance test and show casing report for them?

yagna-1:
+1 on this — stdio is the dominant transport for Python and Go MCP servers in the wild, and right now the only path to running the conformance suite against them is to wrap the stdio server in an HTTP shim, which most maintainers won't do just for testing.

A few notes from a Python-side perspective that might help shape the design, since I've been working in this area:

**1. The CLI surface in #258 is the right one.** `--stdio "<command>"` (matching the existing `--command` for client testing) keeps the runner ergonomically symmetric:

```
npx @modelcontextprotocol/conformance server --stdio "uv run my-server"  --suite active
npx @modelcontextprotocol/conformance server --url    http://localhost:3000/mcp --suite active
```

The two flags are mutually exclusive. The runner spawns the subprocess, frames JSON-RPC over its stdin/stdout, and otherwise reuses the same scenario set.

**2. Watch out for the SSE/Streamable-HTTP-only checks.** A handful of current scenarios assume HTTP semantics (status codes, headers, multi-session). Those would need a `transports: ['http']` filter, or the suite would emit spurious failures for stdio servers that are otherwise spec-conformant. Probably worth tagging at the scenario level rather than gating in the runner.

**3. Stderr is signal.** Python servers commonly log diagnostics to stderr. Capturing and surfacing that in `stderr.txt` (the way client testing already does) is much more useful than discarding it — when a stdio scenario fails, the server's stderr is usually the first thing a maintainer wants to see.

[...]

PR #273, opened 2026-05-13
feat: add sep-2575.yaml requirement traceability

pcarleton:
Adds the SEP-2484 requirement-traceability YAML for **SEP-2575 (stateless MCP)**.

Maps the **32 new/changed normative requirements** introduced by SEP-2575 to conformance check IDs, plus 3 `excluded:` rows (client-internal demux, ack-filter handling, architectural state-identifier guidance).

## Filtering note

SEP-2575 substantially rewrote `transports.mdx` and `lifecycle.mdx`, so the raw added-lines diff contains ~60 RFC-2119 sentences. ~28 of those are pre-existing text that was rewrapped (Origin validation, `x-mcp-header`, Accept header, X-Accel-Buffering, custom transports, stdio stdout/stdin, shutdown sequence) — confirmed by matching against the deletion side of the spec diff. Those are **not** in this YAML since they aren't introduced by this SEP.

## Breakdown

- **12 client-side** check rows → reference for #270
- **20 server-side** check rows → reference for #271
- **3 excluded** rows (signed off during generation)

Generated via `/new-sep 2575`.

### DIFF
--- a/src/seps/sep-2575.yaml
+++ b/src/seps/sep-2575.yaml
@@ -1,0 +1,104 @@
+sep: 2575
+spec_url: https://modelcontextprotocol.io/specification/draft/basic/lifecycle
+requirements:
+  - check: sep-2575-client-populates-meta
+    text: 'Every client request MUST include the following io.modelcontextprotocol/* fields in _meta: protocolVersion, clientInfo, clientCapabilities.'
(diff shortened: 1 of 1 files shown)


pcarleton:
Superseded — the YAML now ships as part of the #270 stack via https://github.com/anubhav756/conformance/pull/1 (with stdio rows moved to `excluded:` → #258).

PR #275, opened 2026-05-14
feat: add conformance Tests for SEP-2549: TTL for List Results

CaitieM20:
Conformance Tests for [SEP-2549: TTL for List Results](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2549/)

## Motivation and Context
Conformance tests for new feature

## How Has This Been Tested?
Local tests related to this change pass, there is an existing test in main related to SEP-2164 that is failing, but unrelated to this change.

## Breaking Changes
no

## Types of changes

- [ ] Documentation update

## Checklist


examples/servers/typescript/everything-server.ts
```diff
@@ -1023,11 +1027,138 @@ function createMcpServer() {
...
+        ttlMs: 300000,
+        cacheScope: 'public' as const
+      };
+    }
+  );
+
+  // ===== SEP-2549: Override list/read handlers to include caching hints =====
+
+  mcpServer.server.setRequestHandler(ListPromptsRequestSchema, async () => {
+    const registeredPrompts = (mcpServer as any)._registeredPrompts as Record<
```
felixweinberger: Can we add `// TODO: replace with native TS SDK Code when SEP-2549 is supported" here and to the other handlers? As we're working around the fact that the SDK doesn't support it yet I guess.

Ideally this should all be deletable once the SDK supports it.

src/seps/sep-2549.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - text: 'If ttlMs is absent, clients SHOULD assume a default of 0'
+    excluded: 'Client-side caching behavior; not observable at the protocol level'
+  - text: 'If ttlMs is negative, clients SHOULD ignore it and treat it as 0'
+    excluded: 'Client-side caching behavior; not observable at the protocol level'
+  - text: 'Once the TTL expires, the response is stale and the client SHOULD re-fetch on next access'
+    excluded: 'Client-side caching behavior; not observable at the protocol level'
+  - text: 'Clients SHOULD NOT treat TTL as a polling interval that triggers automatic background refetches'
+    excluded: 'Client-side caching behavior; not observable at the protocol level'
+  - text: 'Implementations that do choose to poll MUST apply jitter and backoff'
+    excluded: 'Client-side polling behavior; not observable at the protocol level'
```
felixweinberger: Considered asking for a client-side conformance test for this given the MUST, but it's actually conditional on the above SHOULD NOT so the value seems marginal at best.

src/seps/sep-2549.yaml
```diff
@@ -0,0 +1,42 @@
...
+  - text: 'Cached responses MAY be reused for the same authorization context. Caches MUST NOT be shared across authorization contexts'
+    excluded: 'Client/cache-side behavior; not observable at the protocol level'
```
felixweinberger: I wonder if it's worth adding one client side scenario here, something like:

```
sep-2549-client-no-background-poll

Spec: "Clients SHOULD NOT treat TTL as a polling interval that triggers automatic background refetches."

- Harness-as-server responds to tools/list with ttlMs: 500
- Reference client lists once, then idles with the connection open for ~3s
- Harness counts tools/list requests received in that window
- 1 request → SUCCESS. >1 → WARNING (client is background-polling on TTL)
```

Given there's a lot of SHOULD guidance here for the client, feels odd to not have _any_ client scenario, but given none of it is a MUST it's defensible I guess.

But the polling seems to be a specific anti-pattern we'd like to push SDKs away from so might be worth adding as a scenario.

src/scenarios/authorization-server/authorization-server-metadata.ts
```diff
@@ -9,7 +9,9 @@ import { request } from 'undici';
 
 type Status = 'SUCCESS' | 'FAILURE';
 
-export class AuthorizationServerMetadataEndpointScenario implements ClientScenarioForAuthorizationServer {
+export class AuthorizationServerMetadataEndpointScenario
+  implements ClientScenarioForAuthorizationServer
+{
```
felixweinberger: looks like lint noise

felixweinberger (dismissed):
LGTM overall, just needs a rebase / CI fix I think.

Could consider adding a client scenario to specifically catch polling as a WARNING (rather than a FAILURE) to discourage SDKs from writing polling as a default which we seem to want to deliberately avoid, but wouldn't block on it.

src/scenarios/server/caching.ts
```diff
@@ -0,0 +1,341 @@
...
+            'resources/read response includes ttlMs and cacheScope caching hints',
+          status: 'FAILURE',
+          timestamp: new Date().toISOString(),
+          errorMessage: `resources/read request failed: ${error instanceof Error ? error.message : String(error)}`,
+          specReferences: SPEC_REFS
+        });
+      }
+
+      // 6. Aggregate: ttlMs must be a non-negative integer
+      const ttlErrors: string[] = [];
```
panyam, line 270: Is it worth considering folding the type and range validation into each per-endpoint presence check, so one check per endpoint covers presence plus type plus range, and dropping the separate aggregates.

src/scenarios/server/caching.ts
```diff
@@ -0,0 +1,341 @@
...
+          specReferences: SPEC_REFS
+        });
+      }
+
+      // 5. resources/read — use test://static-text from resources/list
+      try {
+        const readResult = await connection.client.request(
+          {
+            method: 'resources/read',
+            params: { uri: 'test://static-text' }
```
panyam: This 'test://static-text' hardcode this works against the everything-server and the negative fixture because both register that exact URI.  Given it is desirable to have the conformance suite run against arbitrary SDK servers, and none of those will expose test://static-text.  This might error and  sep-2549-resources-read-caching-hints would fail with "request failed" rather than an actual caching-hints verdict. Deriving the URI from the first entry of the resources/list result you already fetch a few lines up would make the check portable?

src/scenarios/server/caching.ts
```diff
@@ -0,0 +1,341 @@
...
+            endpoint: f.endpoint,
+            cacheScope: f.fields.cacheScope
+          }))
+        }
+      });
+
+      await connection.close();
+    } catch (error) {
+      // Connection-level failure — push a single failure check
+      checks.push({
```
panyam, line 343: In this connection-failure path we reuse a real check ID.  Looks like the outer catch emits a check with id 'sep-2549-tools-list-caching-hints'. 

If the connection.close() above throws after the seven checks are already pushed, you might get a duplicate ID?   So either way a connection or teardown failure is mislabeled as a tools-list result.  A distinct ID like sep-2549-caching-connection would be cleaner.

examples/servers/typescript/everything-server.ts
```diff
@@ -1023,11 +1027,138 @@ function createMcpServer() {
...
+        description?: string;
+        argsSchema?: any;
+      }
+    >;
+
+    return {
+      prompts: Object.entries(registeredPrompts)
+        .filter(([, prompt]) => prompt.enabled)
+        .map(([name, prompt]) => ({
+          name,
```
panyam: This everything-server prompts override is dropping arguments. The new ListPromptsRequestSchema handler maps { name, title, description } only.  The default McpServer prompt-list handler derives arguments from argsSchema, so this override silently drops them for any prompt that has arguments.  Is this intended?   Will existing prompts scenarios still see what they expect?

examples/servers/typescript/everything-server.ts
```diff
@@ -1023,11 +1027,138 @@ function createMcpServer() {
...
+          const result = await template.readCallback(uri, variables);
+          return {
+            ...result,
+            ttlMs: 300000,
+            cacheScope: 'private' as const
+          };
+        }
+      }
+
+      throw new Error(`Resource not found: ${uri}`);
```
tarekgh: Potentionally surface as -32603 (Internal Error) break SEP-2164? Should this be the following instead? 

```ts
throw new McpError(ErrorCode.InvalidParams, `Resource not found: ${uri}`);
```

Would avoide reimplementing the SDK's resource dispatch entirely be better fix? The SDK's built-in ReadResourceRequestSchema handler already does URI matching, template matching, and throws the correct McpError. If there were a way to intercept the result and attach ttlMs/cacheScope after the SDK dispatche.

src/scenarios/server/caching.ts
```diff
@@ -0,0 +1,341 @@
...
+          timestamp: new Date().toISOString(),
+          errorMessage: `resources/read request failed: ${error instanceof Error ? error.message : String(error)}`,
+          specReferences: SPEC_REFS
+        });
+      }
+
+      // 6. Aggregate: ttlMs must be a non-negative integer
+      const ttlErrors: string[] = [];
+      for (const { endpoint, fields } of allFields) {
+        if (fields.hasTtlMs) {
```
tarekgh: what happen when the server returns valid responses but without caching field? ttlErrors will stay empty even should be failures. `sep-2549-ttl-non-negative` and `sep-2549-cache-scope-valid` will report success I guess. 

Should we count how many fields were actually validated; if zero, emit FAILURE with a message like "no endpoints returned ttlMs", or gate on whether any presence check passed before running the aggregate?

examples/servers/typescript/everything-server.ts
CaitieM20: added and agree.

examples/servers/typescript/everything-server.ts
CaitieM20: All the tests currently pass, but good call out as I think the current testing around prompts is light. but the right long term solution is just to use the SDK.

src/seps/sep-2549.yaml
CaitieM20: This is pretty hard to implement a meaningful test, since you are measuring timing which is not deterministic. I think its more effort than its worth, and the servers still need to safe-guard themselves and shed load anyway.

src/seps/sep-2549.yaml
CaitieM20: These kinds of tests are pretty hard to implement reliably, and there are valid scenarios where a client might call tools/list again -> i.e. error from client etc... I'm worried this test will be more trouble than it's worth:

- now have a longer running test which slows down the test suite
- unexpected false positives, flagging client caching logic that is reasonable.

Plus servers still have to safe guard themselves against thundering herd scenarios anyway.

src/scenarios/server/caching.ts
CaitieM20: I didn't fix this just added a guard open to refactoring later.

src/scenarios/authorization-server/authorization-server-metadata.ts
CaitieM20: yup but ci.yml fails for me if i don't update

pcarleton (approved):
LGTM, thanks!

PR #278, opened 2026-05-15
feat: add conformance test yaml for sep-2260

CaitieM20:
Conformance Tests for SEP-2260, no actual tests as this isn't observable but including yaml documentation. https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2260 - Require Server requests to be associated with Client Requests


felixweinberger (dismissed):
A quick check of https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2260/changes suggests this isn't an exhaustive collection of MUST (NOT) / SHOULD (NOT). 

E.g. I think this one should be probably be reflected?

```
docs/specification/draft/client/elicitation.mdx, added at L712:

▎ Clients MUST return standard JSON-RPC errors for common failure cases:
▎ - Server sends an elicitation/create request with no associated client-to-server request: -32602 (Invalid params)
```

Could we do a pass to make sure all the new clauses from 2260 are captured?

pcarleton:
the `/new-sep` skill should do it for you

CaitieM20:
Added all the ones and excluded them as these are all super-ceded by MRTR. We have a work item to go remove some of these association language from the spec as they are unnecessary now that MRTR enforces this.

pcarleton (approved):

PR #282, opened 2026-05-18
SEP-2350: scope-accumulation check in auth/scope-step-up

### LINKED ISSUE #281: Conformance coverage for SEP-2350 (client-side scope accumulation in step-up auth)
[SEP-2350](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350) clarifies that when a client re-authorizes after an `insufficient_scope` challenge, it **SHOULD compute the union** of previously-granted scopes and newly-challenged scopes so it doesn't lose permissions for other operations.

The existing `auth/scope-step-up` scenario challenges with the *full* required scope set, so a client that just echoes the challenge passes by accident. To make union-accumulation observable, the harness should challenge with only the *delta* (`mcp:write`) and assert the second authorization request still includes the previously-granted `mcp:basic`.

**Spec:** [authorization §runtime-insufficient-scope-errors](https://modelcontextprotocol.io/specification/draft/basic/authorization#runtime-insufficient-scope-errors)
**Host scenario:** `auth/scope-step-up` (`ScopeStepUpAuthScenario`, `scope-handling.ts`) — adds checks, no new scenario per the fewer-scenarios rule.

localden:
Closes #281.

Adds the `sep-2350-scope-union-on-reauth` check to the existing `auth/scope-step-up` scenario per [SEP-2350](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350): on re-authorization after a 403 `insufficient_scope`, clients **SHOULD** compute the union of previously-granted and newly-challenged scopes.

**What changed**
- `src/seps/sep-2350.yaml`: traceability — 1 client `check:`, 1 server `check:` (`sep-2350-server-single-challenge`, tracked for a future server-side scenario), 1 `excluded:` (no-normative-delta reword)
- `ScopeStepUpAuthScenario`: the 403 challenge now lists only the **missing** scope (`mcp:write`), and `tools/call` gates on `mcp:write` alone so the scenario completes for non-accumulating clients (which then surface as a WARNING, not a hang). New check asserts the second authorization request still includes the previously-granted `mcp:basic`.
- Passing example: `examples/.../helpers/withOAuthRetry.ts` now unions the prior token's `scope` with the challenge scope before re-auth.
- Negative: new `auth-test-echo-scope.ts` (re-authorizes with the challenge scope verbatim) + vitest case asserting the WARNING fires.

**Output** — `node dist/index.js client --scenario auth/scope-step-up`

| client | result |
|---|---|
| `everything-client` (accumulates) | 27/27, 0 failed, 0 warnings — `sep-2350-scope-union-on-reauth` SUCCESS |
| `auth-test-echo-scope` (echoes challenge) | 26/26, 0 failed, **1 warning** — `sep-2350-scope-union-on-reauth` WARNING |

`npm test`: auth suite 28/28 every run. `server/all-scenarios.test.ts` flakes ~2/3 in the full suite (port contention with the auth servers; passes in isolation) — pre-existing, unrelated to this change.


pcarleton:
**Follow-up to track, not for this PR:** `ScopeStepUpAuthScenario` is `introducedIn: '2025-11-25'`, so this check also runs under `--spec-version 2025-11-25` even though the union requirement is draft-only (SEP-2350 merged March 2026). A 2025-11-25-compliant client that doesn't accumulate gets a `sep-2350-…` WARNING for behavior its spec never asked for. We don't have per-check version tagging — only per-scenario (`source.introducedIn`/`removedIn` from #265). The "fewer scenarios, more checks" pattern from #281 is the right tradeoff for now, but as more SEPs land mid-cycle on existing scenarios this will keep coming up. Worth an issue to design per-check `introducedIn` (or some other mechanism) so a check inherits a tighter version window than its host scenario.

pcarleton (dismissed):
LGTM, thanks!

pcarleton (dismissed):
LGTM, thanks!

PR #287, opened 2026-05-19
fix(sep-2243): collapse check IDs to one-per-requirement

pcarleton:
## What

SEP-2243's conformance check IDs drifted from its requirement-traceability yaml in #259: the scenario code emits one check ID per *test case*, while `src/seps/sep-2243.yaml` declares one per *normative requirement*. The two naming schemes diverged, so the declared `check:` slugs no longer match any emitted ID.

This renames the emitted IDs so a **check ID maps to a single MUST/SHOULD**, emitted once per case (the per-case detail moves into `name`/`description`). That matches the repo's existing "same id, flip status/message" convention.

| before (per-case) | after (per-requirement) |
|---|---|
| `mcp-method-header-*`, `mcp-name-header-*` | `client-includes-standard-headers` |
| `reject-invalid-tool-*`, `keep-valid-tool` | `client-reject-invalid-tool` |
| `server-accepts-{lower,upper}case-header-name` | `header-name-case-insensitive` |
| reject status checks (mismatch/missing/case × method/name) | `server-reject-invalid-headers` |
| their `-error-code` variants | `server-reject-error-code` |

In the yaml, `server-reject-mismatch` + `server-reject-status` merge into one `server-reject-invalid-headers` MUST ("reject a header-validation failure with HTTP 400"); `server-reject-error-code` stays as the SHOULD.

## Out of scope (intentionally unchanged)
- Base64 / custom-header server rejection checks keep their own param-validation IDs.
- Scenario gates and the whitespace-acceptance check keep their scenario-matching names.

## Why
A follow-up traceability tool maps declared requirements to emitted check IDs by exact match; this realignment lets that work without per-ID override hacks. No behavior change — only check IDs, descriptions, and the yaml row merge.

## Test plan
`npm run typecheck`, `npm run lint`, `npm test` all pass (126 tests). The negative tests run the renamed scenarios against broken impls and assert the collapsed IDs emit FAILURE.

### DIFF
--- a/src/scenarios/client/http-custom-headers.ts
+++ b/src/scenarios/client/http-custom-headers.ts
@@ -626,7 +626,7 @@ export class HttpInvalidToolHeadersScenario extends BaseHttpScenario {
     // Check that valid_tool WAS called — proves client kept valid tools
     const validToolCalled = this.calledTools.has('valid_tool');
     this.checks.push({
-      id: 'sep-2243-keep-valid-tool',
+      id: 'sep-2243-client-reject-invalid-tool',
       name: 'ClientKeepsValidTool',
       description: 'Client MUST keep valid tools while excluding invalid ones',
       status: validToolCalled ? 'SUCCESS' : 'FAILURE',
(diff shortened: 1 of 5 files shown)


pcarleton:
cc @mikekistler @tarekgh @halter73 

I'm reshuffling some check names around for easier traceability, going to merge for expedience, but please feel free to post-hoc review.

PR #293, opened 2026-05-19
fix: unsupported header takes precedence over header mismatch

Yuan325:
The `sep-2575-http-server-header-mismatch-400` should check with a supported header and mismatch protocol version in `_meta`.

According to the SEP, 
```
* Requirement: This header is MANDATORY. Servers should reject requests with a missing or unsupported version.
* This header MUST match the value provided in the Request as specified below.
```
hence, unsupported header takes precedence over header mismatch. In order to run this check, we must provide a supported header version with a mismatch `_meta.protocolVersion` value.


Yuan325:
@pcarleton If you can take a look at this that would be great~ Thank you!

pcarleton (approved):
Lgtm, ty!

PR #299, opened 2026-05-21
fix(initialize scenario): return 202 No Content for notifications (closes #274)

### LINKED ISSUE #274: initialize scenario test server violates specification causing client failure
**Describe the bug**
When the client sends POST `notifications/initialized` after the negotiation round trip, the test scenario server returns HTTP 200 with body `{"jsonrpc": "2.0", "result": {}}` in the last condition branch without `id` as it's a response to id-less notification from the client:

```
if (request.method === 'initialize') {
  this.handleInitialize(request, res);
} else if (request.method === 'tools/list') {
  this.handleToolsList(request, res);
} else {
  res.writeHead(200, { 'Content-Type': 'application/json' });
  res.end(
    JSON.stringify({
      jsonrpc: '2.0',
      id: request.id,
      result: {}
    })
  );
}
```

As per specification [2.1 ​Sending Messages to the Server](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#sending-messages-to-the-server) point 4: 

> 4. If the input is a JSON-RPC response or notification:
> * If the server accepts the input, the server MUST return HTTP status code 202 Accepted with no body.
> * If the server cannot accept the input, it MUST return an HTTP error status code (e.g., 400 Bad Request). The HTTP response body MAY comprise a JSON-RPC error response that has no id.

This causes strict client implementation to fail, as it never expects a JSON response without `id` unless it has HTTP error status as per quoted spec. Client implementation needs a nasty workaround to don't throw error for this scenario which could cause issues with real world MCP server. 

**Expected behavior**
Test server returns 202 Accepted with no body as response for `notifications/initialized`. 

I am happy to contribute if you agree with what I've found 🤔

adityachilka1:
Closes #274.

### The bug

The initialize scenario test server had a fall-through branch for post-negotiation
requests:

```ts
// src/scenarios/client/initialize.ts:78~90
if (request.method === 'initialize') {
  this.handleInitialize(request, res);
} else if (request.method === 'tools/list') {
  this.handleToolsList(request, res);
} else {
  res.writeHead(200, { 'Content-Type': 'application/json' });
  res.end(JSON.stringify({ jsonrpc: '2.0', id: request.id, result: {} }));
}
```

When the client POSTs `notifications/initialized` (no `id` field), that
falls into the else branch and the server replied with
`{"jsonrpc":"2.0","id":undefined,"result":{}}`. That violates the transport
spec (2025-11-25 § 2.1, point 4):

> If the input is a JSON-RPC response or notification:
> - If the server accepts the input, the server MUST return HTTP status
>   code **202 Accepted** with no body.

Strict clients that reject 200-with-body-for-notification were failing against
the scenario.

### The fix

Insert a branch for `request.id == null` before the 200-placeholder else:

```diff
         } else if (request.method === 'tools/list') {
           this.handleToolsList(request, res);
+        } else if (request.id == null) {
+          res.writeHead(202);
+          res.end();
         } else {
           res.writeHead(200, { 'Content-Type': 'application/json' });
```

`== null` catches both `id: null` (explicitly passed) and `id`-omitted (the
JSON.parse result for a notification payload).

### Verification

```
npm run typecheck  # tsgo --noEmit
npm run lint       # eslint + prettier
```

Both pass clean.

### No other behavioral changes

- Initialize and tools/list flows are untouched.
- Requests with a present `id` still hit the 200-placeholder else branch.
- Only the true notification case (e.g. `notifications/initialized`) changes -
  from 200+JSON to 202+empty.


src/scenarios/client/initialize.ts
```diff
@@ -78,6 +78,13 @@ export class InitializeScenario implements Scenario {
           this.handleInitialize(request, res);
         } else if (request.method === 'tools/list') {
           this.handleToolsList(request, res);
+        } else if (request.id == null) {
```
nbarbettini: Can this condition be made more explicit: check that it is in fact a `notifications/initialized` notification and not something else?

adityachilka1:
Good call. Narrowed the condition to check the exact method (`notifications/initialized`) rather than any id-less payload. Anything else - stray notifications, malformed requests - falls through to the existing else branch, so the prior behavior for those cases is preserved.

Force-pushed - single-commit PR, one-line diff from the previous revision. Typecheck and lint clean.

nbarbettini:
This LGTM 👍

pcarleton (approved):

PR #307, opened 2026-05-22
feat: add conformance tests for SEP-2106 (JSON Schema 2020-12 loosening)

olaservo:
## Summary

Adds conformance support for [SEP-2106](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2106) (merged 2026-05-18), which loosens the tool-schema restrictions:

- `inputSchema` may now carry any JSON Schema 2020-12 keyword alongside `type: "object"` (composition: `oneOf`/`anyOf`/`allOf`/`not`, conditional: `if`/`then`/`else`, references: `$ref`/`$defs`/`$anchor`)
- `outputSchema` drops the `type: "object"` requirement entirely — arrays, primitives, and compositions are now valid at the root
- `structuredContent` widens from `{[key: string]: unknown}` to `unknown`

SEP-2106's spec diff in `docs/specification/draft/server/tools.mdx` is **descriptive-only** — it adds no new RFC 2119 sentences. This PR therefore follows the **SEP-1613 capability-test pattern**: a SEP-2106-compliant SDK can advertise the new schema shapes and round-trip non-object `structuredContent` without flattening or rewrapping.

### What's in the PR

- **`src/scenarios/server/sep-2106.ts`** — `Sep2106Scenario` (single class, 9 checks per AGENTS.md "fewer scenarios, more checks"):
  | check ID | severity | what it asserts |
  |---|---|---|
  | `sep-2106-array-output-tool-found` | FAILURE | tool advertised |
  | `sep-2106-array-output-schema-preserved` | FAILURE | `outputSchema.type === 'array'` survives `tools/list` |
  | `sep-2106-array-structured-content` | FAILURE | `tools/call` returns array directly in `structuredContent` |
  | `sep-2106-array-text-fallback` | **WARNING** | the elevated SHOULD: TextContent block carries serialized JSON for pre-SEP clients |
  | `sep-2106-oneof-input-tool-found` | FAILURE | tool advertised |
  | `sep-2106-oneof-input-schema-preserved` | FAILURE | `inputSchema.oneOf` survives `tools/list` |
  | `sep-2106-primitive-output-tool-found` | FAILURE | tool advertised |
  | `sep-2106-primitive-output-schema-preserved` | FAILURE | `outputSchema.type === 'number'` survives |
  | `sep-2106-primitive-structured-content` | FAILURE | `tools/call` returns raw number in `structuredContent` |
- **`src/seps/sep-2106.yaml`** — traceability with one `check:` row (the elevated TextContent-fallback SHOULD) and one `excluded:` row noting that the eight capability check IDs surface in the manifest under `untracked` per the SEP-2484 contract for descriptive-only SEPs.
- **`examples/servers/typescript/everything-server.ts`** — three test tools (`sep_2106_array_output_tool`, `sep_2106_oneof_input_tool`, `sep_2106_primitive_output_tool`) plus a `ListToolsRequestSchema` override that re-injects raw schemas the Zod converter would strip — **gated behind `SEP_2106_ENABLED=1`**.
- **`examples/servers/typescript/sep-2106-broken-schema.ts`** — negative-test server (wraps array in object, omits TextContent fallback).
- **`examples/servers/typescript/sep-2106-compliant-server.ts`** — raw-Express positive-test server that bypasses the SDK to emit fully-compliant wire output.
- **`src/scenarios/server/negative.test.ts`** — two new `describe` blocks pinning specific failing/passing check IDs (per AGENTS.md §"prove it passes and fails").
- **`src/scenarios/index.ts`** — registered in both `pendingClientScenariosList` and `allClientScenariosList` mirroring SEP-1613.

### Why raw HTTP (and the env-var gate)

The currently-pinned SDK validates both incoming and outgoing `structuredContent` as `Record<string, unknown>`. That means:

1. The **SDK Client** rejects any `tools/list` response containing a non-object `outputSchema` — breaking unrelated scenarios that share the everything-server. → tools gated behind `SEP_2106_ENABLED=1`.
2. The **SDK Server** refuses to send a `tools/call` result with array/primitive `structuredContent`, returning JSON-RPC -32602 — so the only way to demonstrate a fully-compliant green-path server today is to bypass the SDK. → `sep-2106-compliant-server.ts` uses raw Express.
3. `Sep2106Scenario` itself uses raw HTTP for the same reason — it needs to see what the server actually put on the wire, not what the SDK validator allows through.

Once the SDK ships SEP-2106 support, the pending entry in `src/scenarios/index.ts` can be removed and the env-var gate can stay (or go) depending on whether the everything-server's existing tools also get migrated to the new shapes.

### Spec-vs-keyword note for reviewers

Per AGENTS.md §"Reviewing PRs > SEP scenarios", severity must come from the spec text. The spec diff for SEP-2106 in `docs/specification/draft/server/tools.mdx` is:

> **Structured** content is returned as a JSON value in the `structuredContent` field of a result. This can be any JSON value (object, array, string, number, boolean, or null) that conforms to the tool's `outputSchema` if one is defined.

Plus two example blocks. No new RFC 2119 keywords. The only normative tie-in is the existing-but-unchanged "SHOULD also return the serialized JSON in a TextContent block" line, which is newly load-bearing under SEP-2106 — captured as the one `check:` row in the yaml and emitted at WARNING severity.

## Test plan

- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean
- [x] `npm test` — 200/200 passing, including:
  - `sep-2106-broken-schema` negative case (pins five expected `FAILURE`s + one `WARNING`)
  - `sep-2106-compliant-positive` (all 9 checks `SUCCESS` against the raw-Express server)
- [ ] Reviewer to verify spec-keyword severity mapping against `docs/specification/draft/server/tools.mdx` (no new RFC 2119 sentences → capability-test framing is the right call)
- [ ] Once a reference SDK ships SEP-2106 support: `npm start -- sdk <sdk>@<ref> --mode client --scenario sep-2106-json-schema-loosening` should go green end-to-end; at that point the `pendingClientScenariosList` entry can be removed in a follow-up.

---

*Note*: implementation assisted by Claude Code 🦉 (full transcript available on request); design/review/spec-keyword judgments are mine.

### DIFF
--- a/examples/servers/typescript/everything-server.ts
+++ b/examples/servers/typescript/everything-server.ts
@@ -137,6 +137,51 @@ const JSON_SCHEMA_2020_12_INPUT_SCHEMA = {
   additionalProperties: false
 };
 
+// SEP-2106: array-at-root outputSchema for conformance testing.
+// Tests that SDKs preserve non-object outputSchema instead of wrapping it.
+const SEP_2106_ARRAY_OUTPUT_SCHEMA = {
+  type: 'array' as const,
+  items: {
+    type: 'object',
+    properties: {
+      hour: { type: 'string' },
+      temp: { type: 'number' },
+      conditions: { type: 'string' }
+    },
+    required: ['hour', 'temp', 'conditions']
+  }
+};
+
+// SEP-2106: oneOf composition inputSchema. type: "object" is required
+// (tool args are always objects), but the oneOf alternates between two
+// required keys — exactly the pattern SEP-2106's Specification §1 enables.
+const SEP_2106_ONEOF_INPUT_SCHEMA = {
+  type: 'object' as const,
+  oneOf: [
+    {
+      properties: { id: { type: 'string', format: 'uuid' } },
+      required: ['id']
+    },
+    {
+      properties: { name: { type: 'string', minLength: 1 } },
+      required: ['name']
+    }
+  ]
+};
+
+// SEP-2106: primitive-at-root outputSchema (raw number).
+const SEP_2106_PRIMITIVE_OUTPUT_SCHEMA = { type: 'number' as const };
+
+// Fixed payload for the array-output tool so the conformance check can
+// compare structuredContent against the TextContent fallback verbatim.
+const SEP_2106_FORECAST_PAYLOAD = [
+  { hour: '09:00', temp: 68, conditions: 'sunny' },
+  { hour: '10:00', temp: 72, conditions: 'partly cloudy' },
+  { hour: '11:00', temp: 75, conditions: 'cloudy' }
+];
+
+const SEP_2106_COUNT_PAYLOAD = 42;
+
 // Function to create a new MCP server instance (one per session)
 function createMcpServer() {
   const mcpServer = new McpServer(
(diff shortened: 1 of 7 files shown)


olaservo:
Closing in favor of #295

After comparing, our two approaches overlap on the schema-preservation surface (composition keywords surviving `tools/list`) but diverge:
- #295 also adds a client-side check for the SEP-2106 security MUST (no auto-deref of network `$ref`) — solid addition, not something I had.
- This branch tested the `tools/call` wire shape end-to-end (array / primitive `structuredContent` actually surviving onto the wire) with a raw-Express compliant reference server, which #295 doesn't cover. SEP-2106's motivation section is largely about that wire shape, so it seems like a real gap worth filling.

I'll open a follow-up draft PR with just the wire-shape scenario + the compliant server, framed as complementing #295 rather than replacing it.

(edited  - Claude was on auto pilot here and decided to post a couple things without asking, sorry for the noise!)

PR #316, opened 2026-05-25
fix: send HTTP DELETE to terminate server sessions after each scenario

### LINKED ISSUE #79: Clean up streamable session after test
**Describe the bug**
The conformance test harness does not issue an HTTP DELETE cleaning up its session after it runs.

**To Reproduce**
1. Run the conformance test against a server.
2. Check to see if that server receives a DELETE.

**Expected behavior**
As much as possible, the conformance test should endeavor to be hermetic, and should clean up when it is done.

Additionally, there should probably be a conformance test for session lifecycle!

**Logs**
If applicable, add logs to help explain your problem.

**Additional context**

pugsatoshi:
This pull request makes server-targeting scenarios issue an HTTP DELETE to terminate the MCP session before closing the
client, so that each scenario leaves the server under test in a hermetic state. The fix lands in the shared
`connectToServer` helper plus a new `terminateSessionRaw` for scenarios that bypass the SDK transport. A follow-up PR will
add a dedicated `session-lifecycle` scenario that exercises the second half of the issue (servers MUST return 404 for
requests on a terminated session ID).

## Motivation and Context

Fixes #79.

The [Streamable HTTP transport spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management) says clients SHOULD send an HTTP DELETE with the `Mcp-Session-Id` header to explicitly terminate a session when they no longer need it. 
The conformance harness never did this — every scenario opened a session via `initialize`, ran its checks, and let the server hold onto the dangling session.
For real servers that gate resources by active session (in-memory state, connection slots, etc.), this is observable as drift between conformance runs.

## How Has This Been Tested?

- `npm run check` — typecheck + lint pass.
- `npm run test` — 173/173 tests pass. `lifecycle.test.ts` gains two new cases asserting that DELETE is sent with the issued
- Manual smoke against `examples/servers/typescript/everything-server.ts`:

$ PORT=3010 npx tsx examples/servers/typescript/everything-server.ts
$ npx tsx src/index.ts server --url http://localhost:3010/mcp --scenario tools-call-simple-text
Running client scenario 'tools-call-simple-text' against server: http://localhost:3010/mcp
Checks:
2026-05-21T10:23:18.394Z [tools-call-simple-text] SUCCESS Tool returns simple text content

Test Results:
Passed: 1/1, 0 failed, 0 warnings

The everything-server now logs `Received session termination request for session <id>` for every scenario, which was previously absent.

## Breaking Changes

None

## Types of changes

- [ ] Documentation update

## Checklist


pugsatoshi:
http-standard-headers.ts is excluded because it was made stateless in #319

pugsatoshi:
Rebased onto latest `main` (25fd443) to resolve conflicts introduced by #318 (version-aware Connection abstraction), which moved `connectToServer` from `src/scenarios/server/client-helper.ts` to `src/connection/sdk-client.ts`.

Changes from the previous revision:

- Import path: `./client-helper` → `../../connection/sdk-client` in `lifecycle.ts` / `lifecycle.test.ts`.
- `MCPClientConnection.transport` is no longer exposed on the public interface; the SDK transport is captured by the `close()` closure since no caller needs direct access. Smaller API surface, same behavior.
- `lifecycle.test.ts` uses `vi.mock(..., importOriginal)` partial mocking so `terminateSessionRaw` remains real and is observable via the global `fetch` stub used by the new DELETE assertions.

All 270 tests pass locally; build + lint clean.

pcarleton:
Thanks @pugsatoshi, this is a correct read of the spec's session-termination SHOULD and worth landing. To save a timezone round trip I pushed two commits onto your branch rather than requesting changes:

1. Merge with current main. The `connectToServer`/`close()` rewrites from #399/#421 conflicted with the docstring hunk; your two payloads (`terminateSession` in `close()`, the `terminateSessionRaw` helper) applied cleanly on top.
2. Review follow-ups: the teardown DELETE now sends the negotiated protocol version instead of a hardcoded 2025-11-25 (a strict server 400s a version it didn't negotiate, the helper swallows the error, and the session survives, which is the exact failure #79 describes). It is also bounded with `AbortSignal.timeout(5s)` so a wedged server can't stall the suite, and there's a regression test asserting `close()` sends the DELETE.

Verified locally: full test suite green, and the typescript, python, go, rust, and csharp everything-servers all pass the touched scenarios with the new teardown.

Follow-ups we'll take separately (not blockers here): a runner-level per-scenario timeout, and bounding the pre-existing raw initialize fetch in lifecycle.ts.

pcarleton (approved):
Reviewed, fixes applied on-branch, SDK matrix green (ts/py/go/rust/csharp).

PR #318, opened 2026-05-26
feat: version-aware Connection abstraction for server scenarios

felixweinberger:
Hoists the connection preamble out of individual server scenarios and into the runner, so the same scenario code runs under both the 2025 stateful lifecycle and the 2026 stateless lifecycle (SEP-2575).

## Motivation and Context

Carry-forward behaviors (`tools/list`, `resources/read`, `prompts/get`, etc.) are spec-invariant, but every server scenario hard-coded `connectToServer()` → SDK `Client.connect()` → `initialize`. That preamble is 2025-specific; in the 2026 draft it's replaced by per-request `_meta` + `MCP-Protocol-Version` header. So a scenario tagged `{introducedIn: '2025-06-18'}` with no `removedIn` was selected under `--spec-version draft` but couldn't actually exercise a pure-2026 server — it would fail at connect, or pass only because the fixture is dual-stack.

This PR makes the connect preamble a function of `--spec-version`, not something each scenario owns.

## What changes

- **`src/spec-types/{version}.ts`** — vendored verbatim from `modelcontextprotocol/schema/{version}/schema.ts` so the suite can type against draft spec versions before any SDK ships them. `npm run sync-schema -- <ref>` refreshes; `SOURCE` records the pin.
- **`src/connection/`** — `Connection` interface (`request<R>(method, params, opts?)`, `notifications`, `close`) with two impls:
  - `connectStateful` — thin adapter over the SDK `Client` (don't reimplement the 2025 handshake/session/SSE)
  - `connectStateless` — raw fetch with `_meta` injection + `MCP-Protocol-Version` header, decoupled from the SDK
  - `connectFor(specVersion)` picks the impl. Both throw `JsonRpcError` on JSON-RPC error responses.
- **`ClientScenario.run(serverUrl)` → `run(ctx: RunContext)`** where `RunContext = {serverUrl, specVersion, connect()}`. Runner builds it from `--spec-version`.
- **22 server scenarios** migrated from `connectToServer()` + SDK convenience methods to `ctx.connect()` + `conn.request<ResultType>('method', params)`. Result types come from `spec-types/{introducedIn}`.
- **12 scenarios tagged `removedIn: DRAFT`** — they test methods or mechanics removed in the 2026 draft.
- **`dns-rebinding-protection`** now picks its probe body from `ctx.specVersion` (initialize vs server/discover) so the "valid Host accepted" check works under both.
- **`everything-server`** stateless path now dispatches carry-forward methods (`tools/call`, `resources/*`, `prompts/get`, `completion/complete`) to the same `McpServer` instance the stateful path uses, via an in-memory client. `tools/call` is served as SSE so progress notifications reach the conformance client.

## How Has This Been Tested?

214/214 unit tests, typecheck, lint, build clean.

Full `--suite all` against `everything-server`:

| `--spec-version` | scenarios | result |
|---|---|---|
| `2025-11-25` | 32 | **32/32** |
| `draft` | 40 | **39/40** |

The one ✗ under draft is `http-header-validation`, which is already in `pendingClientScenariosList` on main ("Pending until everything-server fully implements SEP-2243 header validation") and is unrelated to this change — the scenario itself sends a raw `initialize` to obtain a session, which doesn't work under the stateless lifecycle. That's part of the broader DRAFT-scenario coherence pass.

<details>
<summary><b>Per-scenario matrix (51 scenarios)</b></summary>

### Applicable under both — 21

| Scenario | 2025-11-25 | draft |
|---|:-:|:-:|
| completion-complete | ✓ | ✓ |
| dns-rebinding-protection | ✓ | ✓ |
| json-schema-2020-12 | ✓ | ✓ |
| prompts-get-embedded-resource | ✓ | ✓ |
| prompts-get-simple | ✓ | ✓ |
| prompts-get-with-args | ✓ | ✓ |
| prompts-get-with-image | ✓ | ✓ |
| prompts-list | ✓ | ✓ |
| resources-list | ✓ | ✓ |
| resources-read-binary | ✓ | ✓ |
| resources-read-text | ✓ | ✓ |
| resources-templates-read | ✓ | ✓ |
| tools-call-audio | ✓ | ✓ |
| tools-call-embedded-resource | ✓ | ✓ |
| tools-call-error | ✓ | ✓ |
| tools-call-image | ✓ | ✓ |
| tools-call-mixed-content | ✓ | ✓ |
| tools-call-simple-text | ✓ | ✓ |
| tools-call-with-progress | ✓ | ✓ |
| tools-list | ✓ | ✓ |
| server-sse-multiple-streams | ✓ | ✓ |

### 2025-only (`removedIn: DRAFT`) — 11

| Scenario | 2025-11-25 | draft |
|---|:-:|:-:|
| server-initialize | ✓ | — |
| ping | ✓ | — |
| logging-set-level | ✓ | — |
| resources-subscribe | ✓ | — |
| resources-unsubscribe | ✓ | — |
| server-sse-polling | ✓ | — |
| tools-call-with-logging | ✓ | — |
| tools-call-sampling | ✓ | — |
| tools-call-elicitation | ✓ | — |
| elicitation-sep1034-defaults | ✓ | — |
| elicitation-sep1330-enums | ✓ | — |

### draft-only (`introducedIn: DRAFT`) — 19

| Scenario | 2025-11-25 | draft |
|---|:-:|:-:|
| server-stateless | — | ✓ |
| sep-2164-resource-not-found | — | ✓ |
| caching | — | ✓ |
| http-custom-header-server-validation | — | ✓ |
| http-header-validation | — | ✗ (pre-existing pending) |
| input-required-result-basic-elicitation | — | ✓ |
| input-required-result-basic-sampling | — | ✓ |
| input-required-result-basic-list-roots | — | ✓ |
| input-required-result-request-state | — | ✓ |
| input-required-result-multiple-input-requests | — | ✓ |
| input-required-result-multi-round | — | ✓ |
| input-required-result-missing-input-response | — | ✓ |
| input-required-result-non-tool-request | — | ✓ |
| input-required-result-result-type | — | ✓ |
| input-required-result-unsupported-methods | — | ✓ |
| input-required-result-tampered-state | — | ✓ |
| input-required-result-capability-check | — | ✓ |
| input-required-result-ignore-extra-params | — | ✓ |
| input-required-result-validate-input | — | ✓ |

</details>

<details>
<summary><b>2026 coverage gaps created by <code>removedIn: DRAFT</code> tagging</b></summary>

Some 2025-only scenarios test behavior that still exists in 2026 via a different mechanism. The `removedIn` tag is correct (the wire mechanic changed), but a 2026 sibling is needed to cover the same spec requirement:

| `removedIn` scenario | What it tested | 2026 mechanism | 2026 scenario |
|---|---|---|---|
| `server-initialize` | init handshake, session-id format | `server/discover` + `_meta` | covered — `server-stateless` |
| `ping` | ping roundtrip | (removed entirely) | n/a |
| `logging-set-level` | `logging/setLevel` accepted | per-request `_meta.logLevel` | partial — see below |
| `tools-call-with-logging` | tool emits `notifications/message` after setLevel | tool emits when `_meta.logLevel` present | partial — see below |
| `resources-subscribe` / `-unsubscribe` | `resources/subscribe` accepted | `subscriptions/listen` | partial — see below |
| `tools-call-sampling` | server→client `sampling/createMessage` via SSE | MRTR `inputRequests` | covered — `input-required-result-basic-sampling` |
| `tools-call-elicitation` | server→client `elicitation/create` via SSE | MRTR `inputRequests` | covered — `input-required-result-basic-elicitation` |
| `elicitation-sep1034-defaults` | elicitation schema carries `default` (SEP-1034) | same, via MRTR | **gap** |
| `elicitation-sep1330-enums` | elicitation schema carries `enum` (SEP-1330) | same, via MRTR | **gap** |
| `server-sse-polling` | SSE resumability via `Last-Event-ID` (priming event, `id:`, `retry:`) | (removed entirely; resumable streams not supported) | n/a |

### Scenarios to write for full 2026 parity

| Proposed scenario | What it should assert | Why not already covered | Suggested check IDs |
|---|---|---|---|
| `tools-call-with-logging-meta` | `tools/call` with `_meta['io.modelcontextprotocol/logLevel']: 'debug'` set → server emits `notifications/message` on the response stream at that level | `server-stateless` covers the negative (`sep-2575-server-no-log-without-loglevel`) but not the positive; `sep-2575.yaml` has no `check:` row for "emits at requested level" | `sep-2575-server-emits-log-at-requested-level` (new row in `sep-2575.yaml`) |
| `subscriptions-resources-updated` | `subscriptions/listen` with `resourcesUpdated: {uris: [...]}` → server sends `notifications/resources/updated` on the stream when the resource changes | `server-stateless` covers `sep-2575-server-tags-subscription-id` and `sep-2575-server-sends-{tools,prompts}-list-changed-on-subscription`, but not the resource-content-update path | `sep-2575-server-sends-resources-updated-on-subscription` (new row in `sep-2575.yaml`) |
| `mrtr-elicitation-sep1034-defaults` | `tools/call` → `InputRequiredResult.inputRequests[k].params.requestedSchema` carries `default` values per SEP-1034 | `input-required-result-basic-elicitation` checks the roundtrip but not schema-shape details | `sep-1034-defaults-via-mrtr` (no `sep-1034.yaml` exists yet) |
| `mrtr-elicitation-sep1330-enums` | same, for `enum` constraints per SEP-1330 | same | `sep-1330-enums-via-mrtr` (no `sep-1330.yaml` exists yet) |

Adding the `check:` rows to the SEP YAMLs would make these show as `untested` in `traceability.json` (the existing TODO mechanism); not done in this PR.

</details>

## Breaking Changes

`ClientScenario.run(serverUrl: string)` → `run(ctx: RunContext)`. All in-tree scenarios are updated; out-of-tree scenarios (none known) would need a one-line shim.

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)
- [ ] Documentation update

## Checklist

## Additional context

**Deferred** (out of scope, will conflict with the DRAFT-scenario coherence pass):
- Migrating `stateless.ts` / `input-required-result.ts` off their `sendRpc` helper onto `Connection` — those scenarios assert on `error.code` / HTTP status for nearly every call; would need an `expectError` helper or `httpStatus` on `JsonRpcError`.
- `http-header-validation` coherence under draft (it sends raw `initialize` to get a session).
- Client-conformance side (`src/scenarios/client/*`) — symmetric `ctx.createServer()` abstraction.

**Type import rule:** a scenario imports result types from `spec-types/{its source.introducedIn}`. That's the contract it asserts; if a later spec adds optional fields, the carry-forward scenario doesn't see them (a separate scenario covers the addition).


felixweinberger:
Note: vast majority of this PR is vendoring spec types from `modelcontextprotocol` rather than implementation code:

CleanShot 2026-05-26 at 15 38 33

Actual implementation is ~ +800/-300

src/connection/select.ts
```diff
@@ -0,0 +1,23 @@
...
+  '2025-11-25'
+]);
+
+export function connectFor(
+  specVersion: SpecVersion
+): (serverUrl: string) => Promise<Connection> {
+  return STATEFUL_VERSIONS.has(specVersion)
+    ? connectStateful
+    : connectStateless;
+}
```
felixweinberger, line 25: For reviewers: this is the primary abstraction to allow testing scenarios over both old and new protocol version.

examples/servers/typescript/everything-server.ts
```diff
@@ -72,6 +77,49 @@ function getMrtInputText(inputResponse: unknown, field: string): string {
...
+// so this bridges via the in-memory transport after a one-time initialize.
+//
+// A fresh server+client pair is built per request so concurrent requests
+// can't observe each other's notifications.
+type DispatchClient = {
+  client: Client;
+  drainNotifications: () => unknown[];
+  close: () => Promise<void>;
+};
+async function getStatelessDispatchClient(): Promise<DispatchClient> {
```
felixweinberger, line 93: Shim for the fact that the TS SDK doesn't support stateless servers yet, and we don't want to duplicate every tool/resource/prompt registration in the stateless HTTP branch.

For each stateless request, we spin up an SDK Client + McpServer pair connected over InMemoryTransport. The in-memory client.connect() does the initialize handshake the SDK requires, then we forward the incoming method/params through it. The express handler writes whatever comes back to the HTTP response, so from the outside the fixture looks like a native stateless server.

Goes away once the SDK has stateless server support.

examples/servers/typescript/everything-server.ts
```diff
@@ -1304,11 +1352,19 @@ app.post('/mcp', async (req, res) => {
     }
 
     if (method === 'tools/list') {
+      const dispatch = await getStatelessDispatchClient();
```
felixweinberger, line 1358: Note: the other changes in everything-server here sit inside the `if (!session && (reqVersion || meta)) { ... }` block which **only** applies in stateless mode.

A stateful request falls straight through to `transport.handleRequest()` unchanged.

I think it would be nice to refactor everything-server to have just 2 paths though, `handleStateful` and `handleStateless`, but didn't want to mix that in here.

src/connection/index.ts
```diff
@@ -0,0 +1,88 @@
...
+export interface RunContext {
+  serverUrl: string;
+  specVersion: SpecVersion;
+  /**
+   * Open a version-appropriate connection to the server-under-test.
+   * Scenarios that test the connection mechanics themselves (initialize,
+   * GET-SSE, DNS rebinding) bypass this and use raw fetch.
+   */
+  connect(): Promise<Connection>;
+}
```
felixweinberger, line 52: instead of just a server URL, this is now the context for running a test in so the version can be specififed.

src/scenarios/server/dns-rebinding.ts
```diff
@@ -42,32 +47,61 @@ function getHostFromUrl(serverUrl: string): string {
...
+function probeBody(specVersion: string): unknown {
+  const clientInfo = {
+    name: 'conformance-dns-rebinding-test',
+    version: '1.0.0'
+  };
+  if (specVersion === DRAFT_PROTOCOL_VERSION) {
+    return {
+      jsonrpc: '2.0',
+      id: 1,
+      method: 'server/discover',
```
felixweinberger, line 68: small change to make the dns rebinding check compatible with stateless

src/spec-types/README.md
```diff
@@ -0,0 +1,29 @@
...
+Vendored copies of `schema/{version}/schema.ts` from the
+[modelcontextprotocol](https://github.com/modelcontextprotocol/modelcontextprotocol)
+spec repository.
+
+These are the canonical TypeScript types for each protocol version. The
+conformance suite imports types from here rather than from
+`@modelcontextprotocol/sdk` so that it can test draft spec versions before any
+SDK has implemented them.
+
+**Do not edit these files by hand.** To refresh:
```
felixweinberger, line 12: This one we could argue about - I chose to import spec types directly here instead of relying on the typescript-sdk types to decouple the two.

Because the conformance tests really need to front-run SDKs, coupling to the typescript-sdk feels like the wrong approach here - it also matches conceptually more closely that SDKs are _downstream_ of conformance which itself is downstram of the protocol.

Also given we now have _different_ types we potentially need to handle (as some are removed / changed between 2025-11-25 and DRAFT) I'm thinking we probably can't get away with just having a single import anymore like we did?

Open to discussion on this one though.

felixweinberger:
Audited every `removedIn: DRAFT` tag on this branch against the current draft spec (modelcontextprotocol@6abdad65), prompted by the question about multiple SSE streams. Result: 11 of the 12 tags hold up, one was wrong and is now reverted and reworked instead.

### Brought back: `server-sse-multiple-streams`

The tag was wrong. The scenario asserts that several concurrent POST requests are all accepted and that any per-request SSE streams are readable. That is core transport behavior in the draft too ([transports.mdx](https://modelcontextprotocol.io/specification/draft/basic/transports): "Every JSON-RPC message sent from the client MUST be a new HTTP POST", and the server MUST answer each with `application/json` or `text/event-stream`). Only the request scaffolding was 2025-specific (initialize handshake, `Mcp-Session-Id`). Fixed in 396c055: the scenario keeps both checks and picks its scaffolding by spec version, session-id under 2025-x, `_meta` + `MCP-Protocol-Version` header under draft. It now runs and passes under both `--spec-version 2025-11-25` and `draft`.

### Tags that hold up

| Scenario | Why `removedIn` is correct | Where the behavior went |
|---|---|---|
| `server-initialize` | `initialize` is gone from the draft schema; lifecycle.mdx defines a stateless protocol, and protocol-level sessions are removed with it | `server-stateless` covers `server/discover` + per-request `_meta` |
| `ping` | `ping` is gone from the draft schema; the ping utility page no longer exists in draft | no replacement |
| `logging-set-level` | `logging/setLevel` is gone from the draft schema; logging.mdx specifies per-request `_meta` log level instead | positive `_meta.logLevel` scenario still to write (gap table) |
| `resources-subscribe` / `resources-unsubscribe` | both RPCs gone from the draft schema; subscriptions.mdx: `subscriptions/listen` "replaces the former `resources/subscribe` RPC and the HTTP GET" | `server-stateless` covers listen/acknowledge/list-changed; resource-updated path still to write (gap table) |
| `tools-call-with-logging` | depends on `logging/setLevel`, which is gone | same `_meta.logLevel` gap as above |
| `tools-call-sampling` | asserts the server sends `sampling/createMessage` as a JSON-RPC request on the response stream; transports.mdx: the server "MUST NOT send independent JSON-RPC requests on this stream", server-to-client interactions are embedded per SEP-2322 | `input-required-result-basic-sampling` |
| `tools-call-elicitation` | same MUST NOT, for `elicitation/create` | `input-required-result-basic-elicitation` |
| `elicitation-sep1034-defaults` | same MUST NOT (delivery mechanism); the schema content (`default` on requested schemas) survives unchanged | MRTR sibling asserting defaults on `inputRequests[*].params.requestedSchema` still to write (gap table) |
| `elicitation-sep1330-enums` | same, for `enum` constraints | MRTR sibling still to write (gap table) |
| `server-sse-polling` | asserts SEP-1699 resumability (priming event, `id:`/`retry:` fields, `Last-Event-ID` reconnect); transports.mdx: "Resumable SSE streams via `Last-Event-ID` are not supported" | no replacement |

The distinction I used throughout: a tag is only correct when the *asserted behavior* is gone from the draft, not just the test's setup. `dns-rebinding-protection` and now `server-sse-multiple-streams` are the two cases where the behavior survives and only the setup needed to become version-aware; both stay selected under draft.

pcarleton:
## Pre-merge verification — PR #318 @ e5ec4c9

### Build & test integrity
- [x] `npm ci && npm run check` → clean
- [x] `npm test` → 21 files / 233 tests pass (main: 20 files / 223 — branch adds 10, deletes none)

### E2E — stateful path (2025-11-25, must match main)
- [x] `npx tsx examples/servers/typescript/everything-server.ts`
- [x] `npm start -- server --url http://localhost:3000/mcp --suite active`
      → 30 scenarios, 41 passed, 0 failed — identical to main

### E2E — stateless path (the new code in this PR)
- [x] `npm start -- server --url http://localhost:3000/mcp --suite active --spec-version draft`
      → 20 scenarios, 22 passed, 0 failed (dual-version scenarios running over connectStateless — the parity check)
- [x] `npm start -- server --url http://localhost:3000/mcp --suite draft`
      → 19 scenarios, 61 passed, **5 failed — all in http-header-validation** (pending scenario;
      everything-server doesn't enforce SEP-2243 rejection; need to fix this in a follow up)
- [x] `npm start -- server --url http://localhost:3000/mcp --scenario server-stateless` → 26/26 (draft inferred)
- [x] same + `--spec-version 2025-11-25` → SKIPPED (not applicable), exit 0
- [x] same + `--spec-version 2025-11-25 --force` → runs, 8/20 (server rejects the version mismatch — expected)
- [x] `npm start -- server --url http://localhost:3000/mcp --scenario caching` → passes over stateless connection (spec-version inference for ctx.connect() scenarios)
- [x] `npm start -- server --url http://localhost:3000/mcp --scenario caching --spec-version 2025-11-25 --force` → still passes since caching headers are additive to Result type ( ℹ️ we may want to test for this in a backcompat suite)


### Scenario inventory
- [x] `npm start -- list --server` → 51 scenarios, names identical to main
- [x] `npm start -- list --server --spec-version draft` → main 51 → **branch 40**:
      11 stateful-only scenarios now tagged removedIn:draft (server-initialize, ping, logging-set-level,
      resources-subscribe/-unsubscribe, server-sse-polling, elicitation-sep1034/-1330,
      tools-call-elicitation/-sampling/-with-logging), a few of these need porting, but dropping their stateful ones now.

### Packaged CLI (what npm actually ships)
- [x] `npm run build && node dist/index.js server --url http://localhost:3000/mcp --scenario tools-list` → 2/2

pcarleton (approved):

PR #322, opened 2026-05-28
feat: add session-lifecycle conformance scenario for streamable HTTP

pugsatoshi:
This pull request adds a new `server-session-lifecycle` conformance scenario that verifies the two server-side guarantees the [Streamable HTTP transport spec](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports#session-management) places on session termination: (a) the server accepts (or 405s) an HTTP DELETE bearing the issued `Mcp-Session-Id`, and (b) subsequent requests bearing the terminated session ID get HTTP 404 Not Found. 
It also corrects a related spec deviation in the example `everything-server.ts`, where stale/unknown session IDs were returning HTTP 400 instead of 404.

## Motivation and Context

Refs #79.

Issue #79 asked for two things. The first (the conformance harness should clean up its sessions after a run) is handled by PR 316. This PR addresses the second half — `there should probably be a conformance test for session lifecycle!` — by adding a dedicated scenario that exercises DELETE + 404-on-terminated-session end-to-end. 
While developing the scenario, I noticed that the example server (`examples/servers/typescript/everything-server.ts`) returned HTTP 400 for requests bearing a stale or unknown `Mcp-Session-Id`, which violates the spec's "MUST 404" requirement. 
Without that fix, the new scenario would have failed in `all-scenarios.test.ts`, so the two changes are bundled in the same PR.

## How Has This Been Tested?

- `npm run check` — typecheck + lint pass.
- `npm run test` — 219/219 tests pass. The new `session-lifecycle.test.ts` covers four cases: stateless server (INFO + SKIPPED), happy path (DELETE 200 → POST 404), DELETE-not-supported (405 → both SKIPPED), and the regression case (server returns 200 for a terminated session — Check B FAILURE).
- Manual smoke against the patched `examples/servers/typescript/everything-server.ts`:

$ PORT=3011 npx tsx examples/servers/typescript/everything-server.ts &
$ npx tsx src/index.ts server --url http://localhost:3011/mcp --scenario server-session-lifecycle
Running client scenario 'server-session-lifecycle' against server: http://localhost:3011/mcp
Checks:
2026-05-25T02:25:16.795Z [server-session-delete-accepted       ] SUCCESS Server accepts HTTP DELETE on the issued session ID with a 2xx response
2026-05-25T02:25:16.798Z [server-session-terminated-returns-404] SUCCESS Server returns HTTP 404 for requests bearing a
terminated session ID

Test Results:
Passed: 2/2, 0 failed, 0 warnings

The server logs the expected `Session initialized with ID: ...` / `Received session termination request for session ...` / `Session ... closed` sequence.

## Breaking Changes

Downstream SDK conformance suites may surface failures on the new `server-session-delete-accepted` or `server-session-terminated-returns-404` checks if their server returns 400 (or some other status) for a stale/unknown `Mcp-Session-Id`. 
The scenario is intentionally lenient compared to the spec (servers MAY refuse DELETE outright with 405, in which case both checks are SKIPPED), so the only way to fail is the specific “DELETE accepted, but subsequent POST does not return 404” misbehavior.
SDKs that hit this can add the check IDs to their `expected-failures` baseline as a temporary measure until they implement the spec.

## Types of changes

- [ ] Documentation update

## Checklist


pcarleton:
Thanks @pugsatoshi. Solid scenario design and worth landing; main moved a lot under this PR, so rather than a change-request round trip I pushed the rebase and review fixes onto your branch:

1. Merge with current main, plus the `run(ctx: RunContext)` signature from #318.
2. Follow-up requests now carry the protocol version the server negotiated in the initialize result (parsed from JSON or SSE, bounded at 5s), falling back to the requested spec version, matching the #415/#416 convention. Previously a compliant server on a 2025-03-26/06-18 run would 400 the hardcoded 2025-11-25 follow-ups and false-fail.
3. False-pass guards: a failed initialize (401/500/dead endpoint) is now a FAILURE instead of a green "stateless" INFO, and the initialized notification response is checked.
4. DELETE status handling: the spec never pins a success status for session termination (filed #429 to chase that upstream), so unexpected statuses are WARNING rather than FAILURE, and 404 counts as already-terminated only when the follow-up probe confirms the session is gone.
5. `removedIn`: draft — sessions were removed entirely by SEP-2575.

Verified: 494/494 unit tests, scenario green end-to-end at 2025-03-26/06-18/11-25 against the everything-server, and all five SDK everything-servers (typescript, python, go, rust, csharp) pass 3/3.

pcarleton (approved):
Reviewed, fixes applied on-branch, SDK matrix green (ts/py/go/rust/csharp).

PR #329, opened 2026-06-03
feat: add python-sdk as a built-in SDK runner target

waprin:
Adds `python-sdk` to `KNOWN_SDKS` so `conformance sdk python-sdk` can clone, build, and run it like typescript-sdk and go-sdk. python-sdk already ships conformance fixtures; this is just the runner wiring.

- **build**: `uv sync --frozen --all-extras --all-packages` — one sync over the uv workspace covers both the `mcp` (client) and `mcp-everything-server` (server) packages.
- **client/server**: reuse the SDK's own fixtures (`.github/actions/conformance/client.py` and the everything-server console script). `--port 3000` matches the url and the convention used by the other entries.

Validated: `--mode server --scenario server-initialize` (2/2) and `--mode client --scenario initialize` (1/1).

### DIFF
--- a/src/sdk-runner/index.ts
+++ b/src/sdk-runner/index.ts
@@ -273,7 +273,10 @@ export function createSdkCommand(): Command {
             serverUrl,
             ...passThrough({
               scenario: options.scenario,
-              suite: options.suite,
+              // Default to the `active` suite (excludes pending/draft) — the same
+              // suite tiering runs, and the most reasonable default to avoid
+              // surfacing intentionally-deferred `pending` scenarios.
+              suite: options.suite ?? 'active',
               verbose: options.verbose,
               output,
               specVersion
--- a/src/sdk-runner/known-sdks.ts
+++ b/src/sdk-runner/known-sdks.ts
@@ -43,6 +43,21 @@ export const KNOWN_SDKS: Record<string, SdkConfig> = {
       command: './.conformance-server -http=:3000',
       url: 'http://localhost:3000'
     }
+  },
+  // uv workspace: the `mcp` (client) and `mcp-everything-server` (server)
+  // packages are both members, so one `uv sync --all-packages` covers both
+  // modes. Fixtures live in the python-sdk repo (.github/actions/conformance/
+  // and examples/servers/everything-server). `--port 3000` matches the url and
+  // the 3000 convention used above; the server's own default is 3001.
+  'python-sdk': {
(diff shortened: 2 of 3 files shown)


pcarleton:
Thanks @waprin! 

Can you check the full output? (from discord, seems like there are some failures).

python on the v1 branch should pass 100% at 2025-11-25 (that's what was submitted for tiering).  That might mean a few things:
* i added a "v1" typescript sdk runner, which should pass there
* if you do run on the v1 branch, and it still fails, let's investigate the failures because it may be a harness regression due to some of the transport refactors

dulcianekiffergoncalves-ops (approved):

waprin:
@pcarleton ok Im targeting Python v1 now. Following issues:

  - json-schema-2020-12 (server) — json-schema-2020-12-tool-found FAILURE: no
  json_schema_2020_12_tool advertised.
  - server-sse-polling (server) — server-sse-priming-event + server-sse-retry-field
  WARNINGs: no SSE priming event / retry field.
  - auth/scope-step-up (client) — sep-2350-scope-union-on-reauth WARNING: drops
  previously-granted scope on re-auth.

* For json-schema tool and server-sse-polling found, looks likes its  its cause the python sdk's own confrmance suite runs `active` instead of `all` . The server doesnt implement it. 

So, should we be running `active` or all? 

* for auth/scope-step-up seems the SEP-2350 check is newer than what the python-sdk pins , but it's missing on SDK side. Should we try to add it there or is it out of scope?

pcarleton:
awesome, ty!

I'll approve this to merge in, however we should change it to run just `active` for servers (if you want to do that in this one, i'm happy to re-review).  That's the suite we run for tiering, and so is the most reasonable default here to avoid confusion.

Here's where we set that in tiering:
https://github.com/modelcontextprotocol/conformance/blob/main/src/tier-check/checks/test-conformance-results.ts#L214

You're right about step-up... we should probably add a spec version branch in the step-up behavior so we don't fail on the "non accumulating" behavior from 2025-11-25 (we added "set-wise union" as a requirement in 2026-07-28).

pcarleton (dismissed):

waprin:
@pcarleton ok I changed it to suite active please re-review . I will look into spec-version branch thing.

pcarleton (approved):

pcarleton:
thanks @waprin !

PR #330, opened 2026-06-04
feat(sep-2640): skills server conformance against the Accepted SEP, enumeration + manifest + directory

panyam:
Adds `src/seps/sep-2640.yaml`, the Skills Extension requirement-traceability file, together with eight runnable scenarios: three that exercise a SEP-2640 server, and five that exercise a SEP-2640 client.

**SEP-2640 passed the core-maintainer vote on 2026-09-01 and is Accepted.** This branch is extracted against spec branch head `a3e147ca2710` (2026-08-25), which is the commit the vote ran on, so the requirements here track the voted text rather than an earlier draft.

## Coverage, stated up front

The SEP carries 97 sentences with an RFC 2119 term. This file declares **96 rows: 89 checks and 7 excluded**, each exclusion carrying its reason inline. Of the 89 checks, **40 run on the wire** across the three scenarios below.

The other 49 are client and host obligations, invisible to a harness that is acting as a client. They split into two groups.

**Five are now covered by client scenarios in this PR.** `Scenario` runs the harness in the opposite direction, standing up a deliberately hostile server and grading what the client under test does: retrieval policy (a client MUST NOT fetch a skill's files ahead of need), reads of URIs absent from `resources`, digest mismatch, size mismatch, and frontmatter that disagrees with the entry. Roughly seven more rows are reachable the same way and are not written yet, including the two-server cross-origin pair.

**The remainder is genuinely out of reach of any harness.** What a host displays, whether a user was actually prompted, disk-cache isolation, and the "acting on a skill" context window are application-internal. Those stay traceability-only by nature rather than by omission.

The 7 excluded rows are things the wire genuinely cannot show: DNS resolution of a URI authority component, what a host displays to a user, and cache hits, which are the absence of a request and indistinguishable from a host that simply did not need the file again.

## What the scenarios cover

- **`sep-2640-skills-enumeration`** (`enumeration.ts`, 29 checks) - `skills/list` and `skills/get`, which replaced the retired `skill://index.json` resource. Covers pagination and the atomic-entry rule, the `ttlMs`/`cacheScope` attributes on 2026-07-28 and later, the `{uri, frontmatter, resources}` entry schema, `resources` completeness and the `"dynamic"` sentinel, per-resource `digest` and `size`, the 512-entry and 16 MiB limits, URI scheme and naming rules, and the capability declaration.
- **`sep-2640-skills-manifest`** (`manifest.ts`, 5 checks) - the SKILL.md resource itself: markdown mime type, frontmatter `name` and `description` surfacing on the resource, the final skill-path segment equalling the frontmatter `name`, and the `io.modelcontextprotocol.skills/` reverse-domain prefix on `_meta` keys.
- **`sep-2640-skills-directory`** (`directory.ts`, 6 checks) - `resources/directory/read`, which survived the rewrite behind the same `directoryRead` capability gate. Reads the capability from `server/discover` rather than inferring it from a `-32601`, SKIPs when `directoryRead` is undeclared, and discovers the directory to exercise dynamically.

Five client scenarios run the harness as the server, with the client under test driven through `MCP_CONFORMANCE_SCENARIO`:

- **`sep-2640-client-no-prefetch`** - serves a listing naming a `SKILL.md` and a supporting file, then fails if either is read. The client's contract is to connect, enumerate, and stop.
- **`sep-2640-client-verify-digest`**, **`-verify-size`**, **`-verify-frontmatter`** - each serves an internally honest listing and tampers with exactly one thing on the wire. Rejection is detected by absence: the client's contract is "load the skill, then read its supporting file", and a client that verifies aborts on the load and never reaches the second read.
- **`sep-2640-client-verify-unlisted`** - asks for a URI the manifest does not contain, which a conformant client refuses locally without putting the read on the wire.

Each was verified in both directions, passing against a conformant client and failing against a driver rewritten to skip verification. A check nobody has watched fail is not evidence.

Discovery is dynamic throughout, so the scenarios certify any conformant SEP-2640 implementation rather than one vendor's fixture.

## Motivation and Context

Conformance tests are one of the three deliverables the WG named as gating SEP-2640's move from Accepted to Final, alongside a reference implementation in a Tier 1 SDK and spec-language docs in the extension repo. This PR is that deliverable.

The traceability YAML went up first to float the requirement model for WG review ahead of scenario design. With the model settled and the SEP voted, this lands the scenarios that assert the server-observable half.

## How Has This Been Tested?

Three independent implementations, all green.

| Implementation | Source | Wire | Result | Notes |
| --- | --- | --- | --- | --- |
| `modelcontextprotocol/go-sdk` | PRs 1238 + 1240 | 2026-07-28 stateless | **43, 0 failures** | |
| `modelcontextprotocol/go-sdk` | PRs 1238 + 1240 | 2025-11-25 stateful | **42, 0 failures** | |
| `modelcontextprotocol/csharp-sdk` | PR 1856 | 2026-07-28 stateless | **37, 0 failures** | `directoryRead` not declared, so its six checks SKIP. Optional capability |
| `modelcontextprotocol/csharp-sdk` | PR 1856 | 2025-11-25 stateful | **36, 0 failures** | as above |
| mcpkit `ext/skills` | released | 2026-07-28 stateless | **43, 0 failures** | |
| mcpkit `ext/skills` | released | 2025-11-25 stateful | **42, 0 failures** | |

Run 2026-09-07. Per-scenario at 2026-07-28: enumeration 30, manifest 6, directory 7. Each total includes the framework's `wire-schema-valid`, so subtract one per scenario for SEP-2640 check IDs.

Every 2025-11-25 row is exactly one lower than its 2026-07-28 counterpart, for the same reason in all three cases: `ttlMs` and `cacheScope` are not defined by that schema, so the cache-attributes check reports SKIPPED rather than grading a server for omitting a field the negotiated version does not have. A SKIP is not a gap, it means the requirement does not apply.

The scenarios hardcode no fixture URIs, so pointing them at a new implementation is a URL change. As other SDKs and servers are ready, happy to run the suite against them and add a row. The full loop, including the per-SDK setup, is in `RUNNING_SEP2640.md` on this branch.

**What the pagination checks do and do not prove.** `skills/list` and `resources/directory/read` are both followed to exhaustion, with a page bound and a repeated-cursor guard, and the round-trip is asserted whenever a server offers a cursor. A harness cannot make a server paginate, though, so against a server that returns everything in one page these checks pass without a multi-page flow ever running. Read them as "cursors are honoured correctly when present" rather than as pagination coverage.

**These checks fail when they should.** Three defects have been found through them, each invisible to the implementation that hosted it:

- a server nesting its extension settings one level too deep, burying `directoryRead` where a spec-following client could not see it, so the feature was implemented correctly and unreachable
- a scenario reading only the first page of a paginated `resources/directory/read`, which reported a conformant server as failing
- a check grading `ttlMs` and `cacheScope` without the negotiated version, warning falsely below 2026-07-28 and under-reporting above it

The second and third were found by pointing the suite at an implementation other than the author's, which is the argument for the table above having more than one row in it.

**Running against a server below 2026-07-28.** SEP-2640 states it "has no dependencies beyond the base MCP Resources primitive", and only `ttlMs` and `cacheScope` are conditioned on 2026-07-28 and later. The enumeration scenario now gates that check on the negotiated version, reporting SKIPPED below the floor and SUCCESS or FAILURE above it. The runner defaults to the stateless draft wire and extension scenarios do not match `--spec-version` alone, so an older server needs both flags:

```
--scenario sep-2640-skills-enumeration --spec-version 2025-11-25 --force
```

Without them a 2025-x server refuses with `-32022` and the failure reads as a missing method rather than a wire mismatch. Worth a clearer message, though that belongs in the runner rather than this PR.

`npm run check` (typecheck, eslint, prettier) passes. The vitest suite passes. lefthook pre-push passes.

## Breaking Changes

None. The PR is purely additive: six new files under `src/scenarios/`, one under `src/seps/`, and `RUNNING_SEP2640.md`, plus `io.modelcontextprotocol/skills` appended to `EXTENSION_IDS` in `src/types.ts` and the scenarios registered in `src/scenarios/index.ts`.

The server scenarios sit in the pending list because the everything-server does not implement the skills extension. Targeted runs point at a conformant fixture with `npm start -- server --scenario sep-2640-skills-* --url <fixture>`. Each scenario SKIPs cleanly against a server that does not declare the extension.

## Types of changes

- [x] Documentation update

## Checklist

## Additional context

**The capability-nesting question from the earlier revision is resolved, and the check no longer accepts both forms.** That revision flagged a disagreement between SEP-2640's capability example, which places `directoryRead` directly on the extension object, and an SEP-2133 reading that wrapped settings in a `{specVersion, stability, config}` envelope. It asked which nesting was canonical and accepted either.

SEP-2133 is Final and settles it. It defines the field as a map of extension identifiers to per-extension settings objects, with no envelope, and SEP-2640's declaration block matches that exactly. There is no disagreement between the two SEPs. The envelope was an implementation's invention. `sep-2640-capability-declaration-inline` now asserts the inline form only, which is what caught the bug described above.

**A note on `resultType`**, which appears in all three result examples in the SEP but is deliberately not declared here. It is a base-protocol field on the common `Result` interface rather than a SEP-2640 requirement, and its caching-hint obligations belong to `sep-2549.yaml`. Declaring it here would double-count a base-protocol rule against this extension.

The `sep-2640.yaml` header carries the full provenance: what the 2026-08-21 rewrite changed, which rows were removed and why, which were reworded, and the scenario-to-row mapping.

**Reproducing the mcpkit rows.** mcpkit's server is its skills example:

- Code: https://github.com/panyam/mcpkit/tree/main/examples/skills
- Walkthrough: https://panyam.github.io/mcpkit/examples/skills/

It is listed in the WG's `related-work.md`, though that entry still describes the pre-rewrite `skill://index.json` shape and is due an update. Setup for the go-sdk and csharp-sdk rows is in `RUNNING_SEP2640.md`.


sambhav:
I ran these scenarios against modelcontextprotocol/go-sdk#1238 plus its split filesystem-helper branch and found one harness false negative around directory pagination. Repro: configure the server directory page size to 1, with root children `SKILL.md` and `references/`. `resources/directory/read` correctly returns `SKILL.md` plus `nextCursor` on page 1 and `references` with `mimeType: inode/directory` on page 2. The scenario reports `sep-2640-directory-read-subdir-mimetype` as FAILURE because it searches only `happy.resources` from the first page (directory.ts around the `subdirChild` lookup) rather than following `nextCursor`. With page size 100, the same server passes directory 7/7. The subdirectory check should aggregate directory pages, or search subsequent pages before declaring the prerequisite untestable. This matters because SEP-2640 explicitly permits pagination and a conforming page boundary should not change the outcome.

The run also confirmed the already-documented runner diagnostic: omitting `--spec-version 2025-11-25 --force` against the stateful fixture produces `-32022` but is reported as missing skills methods. I agree with the PR text that this is a runner-message issue rather than a Skills scenario or SDK failure.

Final pinned-wire results after fixing a YAML normalization bug caught by the suite: enumeration 29/29, manifest 4/4 with the two expected metadata SHOULD warnings, directory 7/7, no failures.

panyam:
Confirmed and now fixed. @sambhav - spot on with the diagnosis.

directoryReadAll now aggregates pages with a page bound and a repeated-cursor guard, the same shape listAllResources already used for resources/list.  The pattern was in the file and had simply never been applied to directory reads.

Verified against your SUT rebuilt with DirectoryOptions{PageSize: 1}:

| SUT                 | Before        | After |
|---------------------|---------------|-------|
| go-sdk, PageSize: 1 | 6/7, 1 failed | 7/7   |
| go-sdk, default     | 7/7           | 7/7   |
| mcpkit, default     | 7/7           | 7/7   |

Agreed on the -32022 diagnostic being a runner-message problem rather than a scenario or SDK one. I would rather fix that upstream than massage over it here.

I love that the suite caught a bug in the go-sdk (yaml norm) and vice versa (pagination bug in the suite).   Neither was findable from one side.  

In mcpkit I hardcoded pagesize to 0 with no override, so it never emited a cursor and every run was against a server with exactly one page. By changing page boundary in a conforming way, this failure mode can be caught.

Once go-sdk#1238 and the filesystem-helper branch settle happy to add a go-sdk row to the results table.  I can run both later too.

sambhav:
The latest Go SDK runs reproduce a false-positive warning in `sep-2640-skills-list-cache-attributes`.

At [`fa8aaec`](https://github.com/panyam/mcpconformance/blob/fa8aaec3611e387249da6adc826b02dd77372a3e/src/scenarios/server/skills/enumeration.ts#L292-L311), this check reports missing `ttlMs` or `cacheScope` as `WARNING` without checking the effective protocol version. The SDK intentionally omits those fields before `2026-07-28`, where the requirement does not apply.

Results against modelcontextprotocol/go-sdk#1238 + modelcontextprotocol/go-sdk#1240 at `a33c199`, with `DirectoryOptions{PageSize: 1}`:

| Protocol / transport | Enumeration | Manifest | Directory | Total |
| --- | --- | --- | --- | --- |
| 2025-11-25 / stateful | 29 pass, 1 warning | 6 pass | 7 pass | 42 pass, 0 failures, 1 warning |
| 2026-07-28 / stateless | 30 pass | 6 pass | 7 pass | 43 pass, 0 failures, 0 warnings |

Totals include each scenario’s `wire-schema-valid` check. The sole legacy warning is cache attributes. The filesystem helper now exposes SKILL.md metadata through `resources/list`, so both manifest metadata checks are exercised.

Could we gate the cache-attributes check on the effective protocol version?

- Before `2026-07-28`: report `SKIPPED`, with a version-inapplicability reason.
- From `2026-07-28`: require both fields and report `FAILURE` when either is missing.

This would address both the legacy false positive and the current under-reporting of missing required fields on newer versions. Regression coverage should include legacy omission, modern valid fields, and each modern missing field.

The PR description currently says enumeration already checks this conditionally, so that should be aligned with the implementation too.

For the legacy run, I used `--spec-version 2025-11-25 --force`; this warning is separate from the previously discussed transport diagnostic.

panyam:
@girishkvs - Ran the conf suite against the [C# SDk](https://github.com/modelcontextprotocol/csharp-sdk/pull/1856), and it passes clean on both stateless/stateful wires.

| Wire                  | Enumeration | Manifest | Directory | Total               |
|-----------------------|-------------|----------|-----------|---------------------|
| stateless, 2026-07-28 | 30          | 6        | 1         | 37 pass, 0 failures |
| stateful, 2025-11-25  | 29          | 6        | 1         | 36 pass, 0 failures |

Lookgin good and no fixes needed. Three notes on reading those numbers:

Directory is 1 rather than 7 because the SDK does not declare directoryRead, so all six resources/directory/read checks skip. IMO this is correct, the capability is optional. If you do want them exercised later, declaring the flag will turn them on.

Enumeration is 29 on the older wire because ttlMs and cacheScope are not defined by that schema, so the cache-attributes check skips rather than warning.  I added that gating only today, after @Sam (Bloomberg) reported it warning falsely against the go-sdk.

Each total includes the framework's own wire-schema-valid, so subtract one per scenario for SEP-2640 check IDs.

The shipped - ModelContextProtocol.ConformanceServer made this easy actuall coz it serves stateful at / and stateless at /stateless off one port.    Happy to share details on how I ran this so it can be self serve.

couple of things to watch out for - 
* extension scenarios need --force or they skip as inapplicable
* --spec-version picks the wire lifecycle rather than filtering, so leaving it out it against a pre-2026-07-28 server gives you a -32022 that reads like a missing method.

For comparison, the go-sdk is at 43/43 and mcpkit at 43/43, both with directory declared.

panyam:
@sambhav Thanks mate you are right - it is wrong in both dirs.   Now fixed.

Below 2026-07-28 the attributes are not merely optional, they are undefined by the negotiated schema, so a server omitting them is correct and the WARNING was a false positive. From 2026-07-28 they are required, so a missing one is a FAILURE rather than something to shrug at. The check now reports SKIPPED with a version reason below the floor, and SUCCESS or FAILURE above it, which is exactly the gating you proposed.

Now verified across both Go implementations and in both directions.   "SKIPPED" against the go-sdk at 2025-11-25 where it previously warned, and FAILURE against a deliberately stripped mcpkit build at 2026-07-28.

Also reproduced your numbers against 1238 + 1240 independently, and they match:

| Protocol / transport | Enumeration | Manifest | Directory | Total |
| --- | --- | --- | --- | --- |
| 2026-07-28 / stateless | 30 | 6 | 7 | **43 pass, 0 failures** |
| 2025-11-25 / stateful | 29 | 6 | 7 | **42 pass, 0 failures** |

Manifest is 6 rather than 4 now, so 1240 did what its description claims: both metadata SHOULDs are exercised where they were untestable before. The PR description is updated with this and the stale 41 row is gone.

One thing to watch out for is that the go-sdk's streamable handler serves 2026-07-28 only with "Stateless: true". A default handler negotiates down to 2025-11-25, correctly per SEP-2575, though it might look like a defect at first.

Awesome work with the findings!

src/scenarios/client/skills/verification.ts
```diff
@@ -0,0 +1,246 @@
...
+  },
+  unlisted: {
+    scenario: 'sep-2640-client-verify-unlisted',
+    checkId: 'sep-2640-host-unlisted-read-failure',
+    description:
+      "While acting on a skill, a host MUST resolve reads of the skill's files only to URIs listed in that entry's resources, and MUST treat a read of an unlisted file as a verification failure."
+  }
+};
+
+export class SkillsVerificationScenario extends BaseHttpScenario {
```
pja-ant, line 86: I think in here an `no-prefetch.ts` we need to override this:

https://github.com/modelcontextprotocol/conformance/blob/a983ba93c91e0bb31d0b6849eeb52f0ad1083107/src/scenarios/client/http-base.ts#L164-L166

so that we provide the skills extension capability (and resources), otherwise the server cannot advertise skills.

src/scenarios/server/skills/enumeration.ts
```diff
@@ -0,0 +1,1102 @@
...
+
+      // === capability-empty-object ===
+      // "An empty object indicates support for the extension with no optional
+      // features." Observable as: the declared value is a JSON object, and an
+      // empty one is a valid declaration rather than a malformed capability.
+      checks.push(
+        skillsCheck(
+          'sep-2640-capability-empty-object',
+          'An empty object indicates support for the extension with no optional features.',
+          'SUCCESS',
```
pja-ant: If this is always SUCCESS is it just a tautology that is checking nothing?

src/scenarios/server/skills/enumeration.ts
```diff
@@ -0,0 +1,1102 @@
...
+  ...ENTRY_IDS,
+  ...GET_IDS,
+  ...READBACK_IDS
+];
+
+/**
+ * Agent Skills naming rules as the SEP defers to them: 1-64 characters,
+ * lowercase alphanumeric and hyphens.
+ */
+const SKILL_NAME_PATTERN = /^[a-z0-9]([a-z0-9-]{0,62}[a-z0-9])?$/;
```
pja-ant: Agent skills must not contain consecutive hyphens.
```suggestion
const SKILL_NAME_PATTERN = /^(?!.*--)[a-z0-9]([a-z0-9-]{0,62}[a-z0-9])?$/;
```
maybe this? (untested...)

src/scenarios/server/skills/manifest.ts
```diff
@@ -0,0 +1,358 @@
...
+ * Each check's verbatim spec quote lives next to its check ID in
+ * src/seps/sep-2640.yaml.
+ *
+ * Discovery is dynamic and brand-neutral: the scenario finds a `skill-md`
+ * skill's `SKILL.md` resource from `resources/list` (preferred — it carries the
+ * Resource `name`/`description` metadata) or falls back to the first `skill-md`
+ * entry in `skill://index.json`, hardcoding no fixture skill. Undeclared
+ * extension SKIPs; a declared extension with no discoverable `SKILL.md` reports
+ * the missing prerequisite via untestableCheck (issue #248), never a silent
+ * green.
```
pja-ant, line 14: This is from a stale version of the spec and no longer correct.

src/scenarios/server/skills/enumeration.ts
```diff
@@ -0,0 +1,1102 @@
...
+  'sep-2640-name-naming-rules',
+  'sep-2640-authority-reg-name',
+  'sep-2640-names-should-be-unique'
+] as const;
+
+const GET_IDS = [
+  'sep-2640-skills-get-implemented',
+  'sep-2640-skills-get-entry-shape',
+  'sep-2640-skills-get-no-cursor',
+  'sep-2640-skills-get-unknown-uri-invalid-params'
```
pja-ant, line 92: I've added cache attributes for skills/get, so should add them here

pja-ant (changes requested):
Thanks for putting this together! Just a few things after taking it for a spin

pja-ant (dismissed):

src/scenarios/server/skills/enumeration.ts
```diff
@@ -0,0 +1,1134 @@
...
+      fmErrs.push(
+        `${entryLabel(e, i)}: frontmatter is missing or not an object`
+      );
+      return;
+    }
+    const obj = fm as Record<string, unknown>;
+    if (typeof obj.name !== 'string' || obj.name.length === 0) {
+      fmErrs.push(`${entryLabel(e, i)}: frontmatter.name is missing or empty`);
+    }
+    if (typeof obj.description !== 'string' || obj.description.length === 0) {
```
vijaydeepsinha, line 465: @panyam As per Agent Skills Spec the max length of description can be 1024. Do we need this check here or below  at // === skillmd-frontmatter ===.

src/scenarios/client/skills/verification.ts
panyam: Ah nice catch.    Now fixed.   One thing though.  Because it changes what the regression test has to cover.  Both scenarios *did* advertise `resources` and the skills extension, but this only happened by passing them explicitly to `sendInitialize`.  

The base class intercepts `server/discover` at `http-base.ts:128`, before `handlePost` runs, so that path never reached them and fell through to the `{tools: {}}` default.  So the hole was the 2026-07-28 discover-first lifecycle specifically, not both.

Found that by overriding `discoverCapabilities()` fixed both at once, since `sendInitialize` already defaults to it, and the explicit argument is now gone.

previously they passed anyway as `server/discover` is optional for clients and it was run by neither client.
So there is now a `capabilities.test.ts` asserting both lifecycles for all five scenarios.

src/scenarios/server/skills/enumeration.ts
panyam: Yep it was.   Now fixed.    Found a worse one behind this though.    `skillsCapability` folded any non-object declaration into `undefined`, which downstream read as "extension not declared", so a server declaring `"io.modelcontextprotocol/skills": true` skipped the entire suite and reported a clean run.

This check now ensures the observable bit, that the declared value is an object at all, and `declaredSkillsCapability` separates absent from malformed.  The inline/envelope check SKIPs on a malformed value rather than reporting a misleading SUCCESS, since there are no settings to place either way.

Note - The base `DiscoverResult` schema rejects a non-object here too, so `wire-schema-valid` catches it independently.  The bug was that our scenario was *masking* it with a SKIP.   Now we assert both in
`server/skills/negative.test.ts`, over `true`, a string, an array and `null`.

src/scenarios/server/skills/enumeration.ts
panyam: Holy mother of god - who needs claude when we have you !!  Or is every version of Claude's training just hooking a probe into your brain @pja-ant ????

Applied this now.

src/scenarios/server/skills/manifest.ts
panyam: Ah yes.  Yep fixed now

src/scenarios/server/skills/enumeration.ts
panyam: Makes sense. one thing though - can i do that as a follow up once your 139

139 has a couple of things that help:

1. A new `skills/get` cache-attributes check, mirroring the existing `sep-2640-skills-list-cache-attributes`.
2. The "Protocol Revision" section.  Our YAML row currently carries the SEP's
   "in protocol versions 2026-07-28 and later" conditional.  This is why I went with the version gating in the previous commit.  If the stable page is defined as 2026-07-28-or-later and drops the conditional, the gate may not matter any more for those reading the stable page?  But it still might matter for 11/25 wire that can be targetted?

src/scenarios/server/skills/enumeration.ts
panyam: Ah nice caatch.  Today we only check non-empty at both sites.  Though SEP-2640 delegates the format entirely to Agent Skills and only puts fidelity MUSTs on the server.   So a server relaying an over-long description is still conformant, which is why `name` is checkable (it has its own MUST via the URI final segment) and this is not.  

If you read "as defined by the Agent Skills specification" as importing the field constraints, happy to fold it into`sep-2640-skillmd-frontmatter` as a WARNING and evaluate it here, where it covers every listed skill.

wdyt?

panyam:
Hey @pja-ant heads up.  Sorry my push was after your approval.  There iS a behaviour change.   The `unlisted` client check matched one literal fixture URI, so it would have reported SUCCESS for a client that reads unlisted files freely as long as it probed a different path.  It now grades any read landing outside the entry's `resources`, which is the requirement.  This actually came out of my driver hardcoding the fixture paths where deriving the probe would have made the old checks pass easily.   Also ive now recorded a -ve control.   A client reading a synthesized probe is now FAILURE where it previously passed.  Client scenarios are still 5/5 with 614 suite tests green.

@sambhav @vijaydeepsinha @girishkvs FYI

src/scenarios/server/skills/enumeration.ts
vijaydeepsinha: Yes, that makes sense to me and we can proceed this way.

panyam:
Now that `ext-skills#139` is merged, 73ac2c4 adds two checks:
* `sep-2640-skills-get-cache-attributes` (gated on 2026-07-28 like the list one, so a conformant 2025-11-25 server SKIPs rather than fails) and
* `sep-2640-capability-requires-resources`.


Also Rows 88 -> 90, which more than covers the one retired with `unlisted`.   And mcpkit is 32/32 on 2026-07-28 and 30/30 on 2025-11-25. Negative control also recorded: a build with the attributes stripped reports 31/32.

pcarleton:
Thanks Sri. The manifest-read gate and the two new checks look right, and retiring `unlisted` makes sense.

I've pushed a few small commits so we can land this today:

1. `entry-frontmatter-identical` compares content instead of JSON serialisation, so a server whose encoder sorts map keys (Go's `encoding/json`) no longer fails it.
2. The client-scenario mocks now answer like a conformant server: `ttlMs`/`cacheScope` on `skills/list` and `resources/read`, and `initialize` at the version the client asked for. The runner hands these clients 2025-11-25, so a strict client was disconnecting and every verify scenario reported SKIPPED.
3. Dropped `RUNNING_SEP2640.md`: it pointed at your fork, and the run notes and results table are better kept in the PR description.

(Fixups drafted with Claude Code and reviewed by me.)

pcarleton (approved):

panyam:
Hey @pcarleton thanks heaps man - really appreciate it.

PR #337, opened 2026-06-11
fix: gate SEP-2350 scope-union check to 2026-07-28+

waprin:
## What

The `auth/scope-step-up` scenario emitted the `sep-2350-scope-union-on-reauth` check at **every** spec version. But the set-wise union requirement (SEP-2350) was only introduced in **2026-07-28** — at **2025-11-25** a client that re-requests *only* the challenged scope (dropping its previously-granted scope) is conformant.

So the harness wrongly WARNed conformant 2025-11-25 clients. This surfaced on **python-sdk v1** (pinned to 2025-11-25) while wiring up its runner in #329 — see [Paul's review comment there](https://github.com/modelcontextprotocol/conformance/pull/329):

> we should probably add a spec version branch in the step-up behavior so we don't fail on the "non accumulating" behavior from 2025-11-25 (we added "set-wise union" as a requirement in 2026-07-28).

## How

- Add `specVersionAtLeast(v, threshold)` to `types.ts` — a small timeline-ordering helper (same `indexOf`-over-the-ordered-timeline approach scenario applicability already uses for `introducedIn`/`removedIn`).
- Gate both the live check and its `getChecks()` fallback in `ScopeStepUpAuthScenario` behind `specVersionAtLeast(ctx.specVersion, DRAFT_PROTOCOL_VERSION)` (i.e. 2026-07-28+).
- Tests pin **both sides** of the branch:
  - **enforced at draft** — the echo-scope client (drops prior grant) fails `sep-2350-scope-union-on-reauth`;
  - **silent at 2025-11-25** — the same client passes cleanly (check not emitted).
  - The existing "responds to 401, not 403" negative test moves to draft so its expected `sep-2350` failure is still emitted.

Mirrors the existing draft-gating pattern used for SEP-837 (`application_type`).

## Verification

- `npm test`, `npm run typecheck`, `npm run lint` all green.
- `conformance sdk python-sdk-v1 --mode client` no longer WARNs on `auth/scope-step-up`.

Follow-up to #329.

### DIFF
--- a/src/scenarios/client/auth/index.test.ts
+++ b/src/scenarios/client/auth/index.test.ts
@@ -130,8 +130,12 @@ describe('Negative tests', () => {
   });
 
   test('client only responds to 401, not 403', async () => {
+    // Run at draft so the SEP-2350 union check is also emitted: a client that
+    // never makes the second authorization request fails both the escalation
+    // check and the union check.
     const runner = new InlineClientRunner(ignore403Client);
     await runClientAgainstScenario(runner, 'auth/scope-step-up', {
+      specVersion: DRAFT_PROTOCOL_VERSION,
       expectedFailureSlugs: [
         'scope-step-up-escalation',
         'sep-2350-scope-union-on-reauth'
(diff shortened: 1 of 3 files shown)


waprin:
@pcarleton followup to #329

pcarleton (approved):
this is great, thank you!

PR #350, opened 2026-06-18
fix(sep-2243): scope Mcp-Method to requests only; add Base64 Mcp-Name checks

pja-ant:
Aligns the SEP-2243 standard-header checks with the spec change in modelcontextprotocol/modelcontextprotocol#2891.

## Summary

- **`Mcp-Method` scope (correctness fix)**: the spec now requires `Mcp-Method` on **requests only**; header rules for notification POSTs are explicitly undefined. `HttpStandardHeadersScenario` no longer emits a check row for notification POSTs and drops `notifications/initialized` from the expected-method list.
- **New client check `sep-2243-client-base64-mcp-name`**: the mock server advertises a tool named `tööl_unicode` (non-ASCII). When the client under test calls it, the scenario asserts `Mcp-Name` is wrapped as `=?base64?{b64(utf8(name))}?=`. Emits SKIPPED if the client never calls the tool.
- **New server check `sep-2243-server-decode-base64-mcp-name`**: `HttpHeaderValidationScenario` sends a `tools/call` whose `Mcp-Name` header is the Base64-sentinel encoding of the body's `params.name` and asserts the server accepts it (no 400, no JSON-RPC error).
- Adds both new check IDs to `src/seps/sep-2243.yaml` with the MUST sentences from the #2891 spec diff.

## New check IDs

- `sep-2243-client-base64-mcp-name`
- `sep-2243-server-decode-base64-mcp-name`

## Testing

- `npx tsc --noEmit` clean
- `npm test` — 296 passed (including 3 new negative tests in `http-standard-headers.test.ts`)
- `npm run lint` clean

https://claude.ai/code/session_01Gbqx5LodtAghT5fRYFxeja

### DIFF
--- a/src/scenarios/client/http-standard-headers.ts
+++ b/src/scenarios/client/http-standard-headers.ts
@@ -20,6 +20,18 @@ const SPEC_REFERENCE = {
   url: 'https://modelcontextprotocol.io/specification/draft/basic/transports#standard-mcp-request-headers'
 };
 
+const SPEC_REFERENCE_ENCODING = {
+  id: 'SEP-2243-Value-Encoding',
+  url: 'https://modelcontextprotocol.io/specification/draft/basic/transports#value-encoding'
+};
+
+/**
+ * Tool name containing non-ASCII characters. Tool/prompt names are only
+ * SHOULD-constrained to header-safe characters, so a name outside the safe
+ * set MUST be carried in `Mcp-Name` via the Base64 sentinel encoding.
+ */
+const UNICODE_TOOL_NAME = 'tööl_unicode';
+
 export class HttpStandardHeadersScenario extends BaseHttpScenario {
   name = 'http-standard-headers';
   description =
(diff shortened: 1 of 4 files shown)


src/scenarios/client/http-standard-headers.ts
```diff
@@ -79,6 +93,19 @@ export class HttpStandardHeadersScenario extends BaseHttpScenario {
...
+    if (!this.unicodeToolCallReceived) {
+      result.push({
+        id: 'sep-2243-client-base64-mcp-name',
+        name: 'ClientMcpNameHeaderBase64',
+        description:
+          'Client Base64-encodes Mcp-Name when the source value is not header-safe',
+        status: 'SKIPPED',
+        timestamp: new Date().toISOString(),
+        errorMessage: `Client did not call '${UNICODE_TOOL_NAME}'; Base64 Mcp-Name encoding was not exercised.`,
+        specReferences: [SPEC_REFERENCE_ENCODING]
```
LucaButBoring, line 105: Should this be an `untestableCheck`? This is testing a `MUST` requirement, so it shouldn't be marked as a skip if it's not actually tested.

src/scenarios/client/http-standard-headers.ts
```diff
@@ -202,6 +241,57 @@ export class HttpStandardHeadersScenario extends BaseHttpScenario {
     });
   }
 
+  private checkMcpNameBase64Header(
```
LucaButBoring, line 244: This function should also update `nameHeaderChecks`, shouldn't it?

src/scenarios/client/http-standard-headers.ts
```diff
@@ -202,6 +241,57 @@ export class HttpStandardHeadersScenario extends BaseHttpScenario {
...
+    const mcpNameHeader = req.headers['mcp-name'] as string | undefined;
+    const expectedHeader = `=?base64?${Buffer.from(bodyName, 'utf-8').toString('base64')}?=`;
+
+    const errors: string[] = [];
+    if (!mcpNameHeader) {
+      errors.push(
+        `Missing Mcp-Name header on tools/call for '${bodyName}'. Clients MUST include the Mcp-Name header for tools/call requests.`
+      );
+    } else {
+      const base64Match = mcpNameHeader.match(/^=\?base64\?(.*)\?=$/);
```
LucaButBoring, line 263: Node's base64 decoder is a bit permissive, for example `dMO2!w7ZsX3VuaWNvZGU=` would pass this test even though the `!` is invalid:

```bash
$ node -e 'console.log(Buffer.from("dMO2!w7ZsX3VuaWNvZGU=", "base64").toString("utf8"))' 
tööl_unicode
```

Its encoder is fine though (produces `dMO2w7ZsX3VuaWNvZGU=`), so we should be doing two separate checks here - one for the base64 sentinel and then a second equality check against `expectedHeader`.

LucaButBoring (changes requested):

PR #356, opened 2026-06-19
fix(sep-2663): use renumbered draft error-code constants in tasks scenarios

panyam:
Fast-follow to the tasks-mrtr extension (`#262`), addressing post-merge review feedback.

## Problem

The tasks extension scenarios hardcoded the JSON-RPC error codes `-32003` (Missing Required Client Capability) and `-32001` (Header Mismatch). The draft spec types were since renumbered to `-32021` and `-32020` per modelcontextprotocol#2907 (landed in `#353`). A server built against the current `draft.ts` now returns the new codes, so these checks would produce false failures against a conformant server.

`#353` updated `draft.ts`, `sep-2575.yaml`, and `sep-2243.yaml`, but `#262` had not merged yet, so its tasks files kept the old literals.

## Change

- `capability.ts`, `headers.ts`, `required-task-error.ts` now import `MISSING_REQUIRED_CLIENT_CAPABILITY` and `HEADER_MISMATCH` from `spec-types/draft` instead of hardcoding, matching the existing `http-standard-headers` scenario. The checks assert whatever the schema currently defines, so a future renumber propagates automatically.
- Dropped the now-stale code number from the affected check-ID slugs, e.g. `sep-2663-server-returns-32003-when-required` becomes `sep-2663-server-returns-missing-capability-when-required`. This keeps the slug from going stale on any future renumber. Renamed in both the scenario `.ts` and `sep-2663.yaml`.
- Updated the `sep-2663.yaml` requirement text to `-32021`, consistent with the already-updated `sep-2575` and `sep-2243` entries.
- Refreshed doc-comments, check descriptions, and error messages to the new codes (interpolating the constant where the string is emitted at runtime).

## Notes

- The tasks scenarios remain in `pendingClientScenariosList`, so default CI is unaffected.
- `traceability.json` is not touched here. It already predates the upstream renumber (it still shows the old `-32001`/`-32003` in the `sep-2575`/`sep-2243` rows) and is regenerated separately from a full suite run with `--results`. `sep-2663` is not yet represented in the manifest.

## Verification

- `npm run typecheck` clean
- `npm run lint` clean
- `npm test` 292 passed

### DIFF
--- a/src/scenarios/server/tasks/capability.ts
+++ b/src/scenarios/server/tasks/capability.ts
@@ -12,6 +12,7 @@
 
 import { ClientScenario, ConformanceCheck } from '../../../types';
 import type { Connection, RunContext } from '../../../connection';
+import { MISSING_REQUIRED_CLIENT_CAPABILITY } from '../../../spec-types/draft';
 import { SEP_2575_REF, SEP_2663_REF } from './mrtr-helpers';
 import { errMsg, failureCheck } from './mrtr-helpers';
 import { TASKS_EXTENSION_ID } from './helpers';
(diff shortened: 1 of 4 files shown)


panyam:
@pcarleton - Please find the the follow up to 262 here (the upgrades on the error codes based on #353).  

The tasks scenarios now import the renumbered constants (`MISSING_REQUIRED_CLIENT_CAPABILITY` = -32021, `HEADER_MISMATCH` = -32020) from `spec-types/draft` instead of the hardcoded `-32003`/`-32001`, matching the existing `http-standard-headers` scenario.

I also renamed the `*-32003` check-ID slugs to drop the embedded number (e.g.  `sep-2663-server-returns-32003-when-required` becomes `...-returns-missing-capability-when-required`) so they can't go stale on the next renumber.

Ive gone ahead and sanity-checked end to end against my conformant Go SDK server - before the SDK matched the renumber, these scenarios produced 6 failures (`tasks/get MUST return -32021; got -32003` across the capability and required-task-error checks, and `expected -32020 HeaderMismatch; got -32001` on the header checks).   After the fix they are back to green

Published conformance report (SEP-2663 row):  https://panyam.github.io/mcpkit/conformance/

pcarleton (approved):

PR #358, opened 2026-06-19
Fix custom-header schema freshness

### LINKED ISSUE #354: http-custom-headers requires a SEP-2243/SEP-2549 interpretation that is not implied by either SEP
## Issue

The `http-custom-headers` scenario serves `tools/list` with `"ttlMs": 0` (immediately stale per SEP-2549) and then asserts the very next `tools/call` carries `Mcp-Param-*` headers derived from the just-returned schema. SEP-2243 explicitly tells the client to **omit** `Mcp-Param-*` *"if the cached schema is stale (e.g., its TTL has expired)"*, so a strict reading of the two SEPs together makes this scenario unpassable.

## Spec text

**SEP-2243** (`docs/specification/draft/basic/transports.mdx`, "Custom headers from tool parameters"):

> If the client does not have the tool's `inputSchema` (e.g., `tools/list` has not yet been called) **or the cached schema is stale (e.g., its TTL has expired)**, the client SHOULD send the request without custom `Mcp-Param-*` headers. If the server rejects the request because required custom headers are missing, the client SHOULD call `tools/list` to obtain the current `inputSchema`, then retry the original request with the appropriate headers.

**SEP-2549** (`docs/specification/draft/server/utilities/caching.mdx`):

| `ttlMs` value | Client behavior |
|---|---|
| `0` | "The response SHOULD be considered immediately stale, The Client MAY re-fetch every time the result is needed." |
| missing | "clients SHOULD assume a default `ttlMs` of 0 (immediately stale)" |

Composing the two: a server that publishes `tools/list` with `ttlMs: 0` is telling the client *"my schema is stale the moment you receive it,"* and SEP-2243 then tells the client to *omit* `Mcp-Param-*` until either a fresh refetch arrives or the server rejects with `-32001 HeaderMismatch` and the client retries.

## What the scenario does

`@modelcontextprotocol/conformance@0.2.0-alpha.4`, scenario `http-custom-headers`:

1. Serves `tools/list` → `{ resultType:"complete", ttlMs:0, cacheScope:"private", tools:[…test_custom_headers…] }`.
2. Has the harness immediately call `tools/call test_custom_headers` with a fixed argument set seeded via `configCtx.toolCalls`.
3. Asserts (via `ClientSupportsCustomHeaders` etc.) that the call carried `Mcp-Param-Region`, `Mcp-Param-Verbose`, …

Step 1 declares the schema immediately stale; step 3 requires headers derived from that same stale schema. No `-32001 HeaderMismatch` rejection path is exercised — the harness's `handleToolsCall` always returns success regardless of headers, so a SEP-2243-compliant "retry on `-32001`" loop never fires either.

(Note: there is a separate issue, #344, where the same scenario's `test_custom_headers.priority` and `.float_val` declare `x-mcp-header` on `type:"number"`, which SEP-2243 forbids. A SEP-2243-conformant client correctly excludes the whole tool from `tools/list`, which produces the same observable failure for an additional reason. **Even after #344 is fixed**, the SEP-2243/SEP-2549 stale-schema interaction reported here remains and continues to fail the scenario for SDKs that take both SEPs strictly.)

## Related

- #344 — `type:"number"` annotations in the same scenario fixture (independent blocker; fixing #344 alone does not fix this issue for strict SDKs).

jstar0:
## Summary

Make the positive `http-custom-headers` mock advertise a fresh `tools/list` schema before requiring clients to mirror `x-mcp-header` parameters into `Mcp-Param-*` request headers.

## Root Cause / Context

`http-custom-headers` returned `tools/list` with `ttlMs: 0`, which marks the tool schema immediately stale. The scenario then required the following `tools/call` to include `Mcp-Param-*` headers derived from that schema.

A strict client that composes the custom-header and caching rules can reasonably treat that schema as stale and omit schema-derived headers. That makes the positive custom-header scenario fail because of the mock's cache hint rather than the client's header support.

## Changes

- Change the positive custom-header `tools/list` response to use a positive TTL.
- Add a regression test that only sends schema-derived `Mcp-Param-*` headers when the listed schema is fresh, then asserts all custom-header checks pass.
- Update the 2026 result-field test to pin this scenario to a positive TTL while still checking the required cache fields.

## Scope / Risk

This only changes the positive `http-custom-headers` client scenario. The invalid-tool scenario and other cacheable mock responses are unchanged.

## Verification

```sh
npm test -- src/scenarios/client/http-custom-headers.test.ts -t "serves a fresh tools/list TTL"
npm run typecheck
npm run lint
npm run build
npm test
```

All commands passed.

Closes #354


pcarleton (approved):
Thanks — taking this because ttlMs: 300000 is a more realistic fixture, but wanted to note that since this PR was opened clients should not drop Mcp-Param-* on a stale schema: since modelcontextprotocol/modelcontextprotocol#2972 and #3071 the 2026-07-28 text makes header emission independent of TTL (streamable-http 'Client Behavior' steps 3–5), so a client that omitted them at ttlMs: 0 would still be non-conformant.

PR #371, opened 2026-06-26
fix(http-custom-headers): use spec-permitted x-mcp-header types

Kludex:
## Summary

The `http-custom-headers` client scenario annotates three `number`-typed properties with `x-mcp-header` (`priority` and `float_val`, in both `test_custom_headers` and `test_custom_headers_null`). [SEP-2243](https://modelcontextprotocol.io/specification/draft/server/tools#x-mcp-header) forbids this:

> **MUST** only be applied to parameters with primitive types (integer, string, boolean). Parameters with type `number` are not permitted.

The same spec says a conformant client **MUST reject** (exclude from `tools/list`) any tool whose `x-mcp-header` annotations violate the constraints:

> Clients **MUST** reject tool definitions where any `x-mcp-header` value violates these constraints. Rejection means the client **MUST** exclude the invalid tool from the result of `tools/list`.

So a strict, spec-conformant client drops both `test_custom_headers` and `test_custom_headers_null` and never calls them - failing `ClientSupportsCustomHeaders` (and `ClientCustomHeaderOmitNull`) through no fault of its own. The fixture is internally inconsistent: it requires the client both to reject the tool (per the type constraint) and to call it and mirror its params.

This surfaced while implementing SEP-2243 client emission in the Python SDK (modelcontextprotocol/python-sdk#2990): the SDK rejects `number` per the spec text, so the scenario is unpassable as written. The TypeScript SDK works around it by deliberately allow-listing `number` against the spec ([note in `mcpParamHeaders.ts`](https://github.com/modelcontextprotocol/typescript-sdk/blob/v2-2026-07-28/packages/core/src/shared/mcpParamHeaders.ts)). Fixing the fixture lets both SDKs follow the spec text.

## Changes

- `priority`: `number` -> `integer` (both tools). Its value (`42` / `1`) and the resulting `Mcp-Param-Priority` header are unchanged.
- `float_val`: drop the `x-mcp-header` annotation. A floating-point header parameter is exactly what the spec prohibits and has no valid replacement; the property stays in the body and is now asserted **not** to be mirrored, mirroring the existing negative check on the unannotated `query` parameter.
- Treat `'integer'` like `'number'` in `validateEncodedHeader` and the check-id classification, so `Priority` keeps its cross-SDK numeric tolerance (`42` vs `42.0`) and lands under `sep-2243-client-encode-values`.

## Verification

```sh
npm run typecheck
npm run lint
npm run build
npx vitest run src/scenarios/client/http-custom-headers.test.ts   # 5 passed
npm test                                                          # 315 passed
```

All pass.

### DIFF
--- a/src/scenarios/client/http-custom-headers.ts
+++ b/src/scenarios/client/http-custom-headers.ts
@@ -321,7 +321,7 @@ export class HttpCustomHeadersScenario extends BaseHttpScenario {
                 float_val: {
                   type: 'number',
                   description:
-                    'Floating point numeric value — no x-mcp-header annotation: the spec forbids x-mcp-header on `number`-typed properties, so this is sent in the body but never mirrored'
+                    'Floating point value — no x-mcp-header annotation, should not be mirrored'
                 },
                 non_ascii_val: {
                   type: 'string',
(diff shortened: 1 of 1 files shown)


pcarleton (approved):

pcarleton:
ty @Kludex! I reworded slightly. I'm also going to open an issue b/c I think rejecting a full tool entry if the header is invalid seems a bit extreme (vs. just not sending that header).

Issue #245, opened 2026-04-24
Policy: do SHOULD-level (WARNING) checks count toward Tier-1 conformance?


pcarleton:
SEP-1730 says Tier-1 requires "all conformance tests pass" / "100% compliance" but doesn't define whether SHOULD-level checks (which emit `WARNING`) count. Our own tooling disagrees with itself:

| Layer | Behavior | Location |
|---|---|---|
| SEP-1730 | undefined | [seps/1730](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/seps/1730-sdks-tiering-system.md) |
| `tier-check` | WARNING ignored in `pass_rate` (only SUCCESS/FAILURE counted) | `src/tier-check/checks/test-conformance-results.ts:55-57` |
| CI runner | `warnings > 0` → `overallFailure` | `src/runner/client.ts:202` |

Today every Tier-1 SDK (Go, TypeScript, Python) shows 0 warnings. So the de-facto policy is "WARNINGs must be fixed," but `tier-check` would still award Tier-1 if one slipped through.

**What a WARNING actually means here:** since conformance runs against a maximally-configured everything-client, a WARNING signals "the SDK cannot be configured to satisfy this SHOULD" (missing API surface), not "a typical app skips it." That's a stronger statement than a SHOULD usually carries.

**Decide and document one of:**
- (a) Tier-1 requires zero WARNINGs → align `tier-check` with `client.ts:202`; scenario authors keep using WARNING for SHOULD and it's effectively required
- (b) Tier-1 tolerates WARNINGs → drop `warnings > 0` from `client.ts:202` `overallFailure`; WARNINGs become genuinely advisory
- (c) Distinguish "SDK can't" (FAILURE) from "SDK can but everything-client didn't opt in" (WARNING) — probably impractical

Since it's roughly "0 cost" to do it today, and having SHOULD checks be required for Tier 1 makes the ecosystem stronger, I'm leaning towards (a).

Whichever we pick should be recorded in AGENTS.md and potentially proposed as a clarification to SEP-1730.

Came up reviewing #200 (SEP-2164), where the same SHOULD-vs-FAILURE question determines whether a wrong error code blocks Tier-1. Related: #243.

PR #375, opened 2026-06-30
fix(sse-retry): gate retry timing only on the early side

maxisbey:
## Problem

`sse-retry` flakes in python-sdk CI (~4–5% of runs, including recent failures on every push to main), always with the signature `✗ sse-retry: 2 passed, 0 failed, 1 warnings` — the retry-timing check landing in its (700, 1000]ms "slightly late" WARNING band, which fails the scenario under the WARNING policy (#245). Example run: https://github.com/modelcontextprotocol/python-sdk/actions/runs/28443055485

The 2025-11-25 spec sentence is "The client **MUST** respect the `retry` field, waiting the given number of milliseconds **before** attempting to reconnect" — a lower bound. Neither MCP nor WHATWG SSE bounds reconnect delay from above. The measured delay is the client's sleep (≥ retry) plus EOF detection, GET propagation, and harness dispatch — all inflated by load, and suite mode runs every scenario's client concurrently. Measured locally with the python client: 503ms idle against the 700ms gate, 591–674ms with the suite pinned to 2 cores. The band gates the environment, not the client.

## Change (one file, +11/−36)

- Delete the slightly-late WARNING and >2× FAILURE bands: the timing check is now FAILURE only when the reconnect arrives earlier than `retry − 50ms`, otherwise SUCCESS, with the measured delay kept in `details`. Load can only push a compliant client deeper into the passing region, so the flake is structurally gone.
- Demote the "could not measure timing" path from WARNING to INFO (the other spurious-warning path).
- One-line fix: the initialize response negotiated `protocolVersion: 2025-03-26`, a version whose spec had no retry MUST; now 2025-11-25 (same issue 4686e06 fixed for server-sse-polling).

Strictly relaxing — no SDK that passes today can start failing on a pin bump; retry stays at 500ms so test duration is unchanged.

Trade-off accepted: the >2× band was the only (unsound) detector for clients that ignore the retry field — a client's ~1000ms default backoff sat near that threshold by coincidence, and a loaded runner pushed compliant clients over it. With the band gone, retry-field-ignorance is untested, which seems proportionate for a scenario that is removed in draft.

## Verification

- python-sdk client, real run: 3/3 SUCCESS (`actualDelayMs: 502`).
- python-sdk client patched to ignore the retry hint (reconnects at its 1000ms fallback, previously red): now passes, delay recorded in details.
- `npm run test` 328 passed; typecheck and lint clean.

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

### DIFF
--- a/src/scenarios/client/sse-retry.ts
+++ b/src/scenarios/client/sse-retry.ts
@@ -42,10 +42,8 @@ export class SSERetryScenario implements Scenario {
   private pendingToolCallId: number | string | null = null;
   private getResponseStream: http.ServerResponse | null = null;
 
-  // Tolerances for timing validation
+  // Tolerance for timing validation (early side only; lateness is not gated)
   private readonly EARLY_TOLERANCE = 50; // Allow 50ms early for scheduler variance
-  private readonly LATE_TOLERANCE = 200; // Allow 200ms late for network/event loop
-  private readonly VERY_LATE_MULTIPLIER = 2; // If >2x retry value, client is likely ignoring it
 
   async start(_ctx: ScenarioContext): Promise<ScenarioUrls> {
     return new Promise((resolve, reject) => {
@@ -290,7 +288,8 @@ export class SSERetryScenario implements Scenario {
       jsonrpc: '2.0',
       id: request.id,
       result: {
-        protocolVersion: '2025-03-26',
+        // The retry MUST under test was introduced in 2025-11-25 (SEP-1699)
+        protocolVersion: '2025-11-25',
         serverInfo: {
           name: 'sse-retry-test-server',
           version: '1.0.0'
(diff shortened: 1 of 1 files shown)


pcarleton:
can we fix this more surgically?
* this test is irrelevant in 2026-07-28 and beyond, so i don't want to invest a bunch more into it
* SDK's have passed at this for previous tiering, so every change to an existing suite means potentially noisy failures when an SDK bumps (maybe it gets better! but we'd need to test that)
* Bumping timeouts to be super long means every test runs for at least that long.  if we are going to do this, i'd rather we categorize the test as "long running" / tiering only, so we don't run it in CI by default, and then we just run it during tiering. that's not ideal either though
* 5+ line comments are ai smell imo

maxisbey:
Descoped — the diff is now one file, +11/−36:

- deleted the two late-side bands (the 2025-11-25 MUST is "waiting the given number of milliseconds *before* attempting to reconnect" — a lower bound; neither MCP nor WHATWG SSE bounds reconnect delay from above), and demoted the can't-measure warning to INFO
- retry stays at 500ms, so test duration is unchanged — no need for a long-running/tiering-only class
- strictly relaxing: no SDK that passes today can start failing on a bump; flaky SDKs just stop flaking
- kept one one-line fix: the initialize response negotiated 2025-03-26, a version with no retry MUST (same issue 4686e06 fixed for server-sse-polling)

Trade-off accepted: the >2× band was the only (unsound) detector for clients that ignore the retry field; with it gone that's untested, which seems right for a scenario that's removed in draft. PR description updated to match.

<sub>[AI Disclaimer](https://gist.github.com/maxisbey/6123d132484e4c533eab519a2800693d)</sub>

pcarleton (approved):
thank you!

PR #381, opened 2026-07-01
Add json-rpc-batch-rejection scenario for 2025-06-18+

### LINKED ISSUE #378: Add conformance test (2025-06-18+): servers MUST reject JSON-RPC batch requests
**Describe the bug**

The conformance suite does not test whether servers reject JSON-RPC batch requests (HTTP POST bodies that are a JSON **array** of request objects). MCP removed batch support in [2025-06-18](https://modelcontextprotocol.io/specification/2025-06-18/changelog#major-changes) ([modelcontextprotocol/modelcontextprotocol#416](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/416)); the draft spec continues with single-message `JSONRPCMessage` only.

All harness helpers (`sendStatelessRequest`, `ctx.connect()` → `conn.request()`) send single JSON-RPC objects. A server or SDK that accepts batch arrays (or crashes on them) passes conformance today with no signal in tier-check results.

**This gap applies only to spec versions where batching was removed:** `2025-06-18`, `2025-11-25`, and draft (`2026-07-28`). It must **not** run for `2025-03-26`, which *requires* batch support — a rejection check there would false-fail compliant servers.

**To Reproduce**

1. Run the server conformance suite against any reference SDK, e.g. `node dist/index.js server --url '<sdk url>' --suite all`.
2. Observe that no scenario POSTs a batch body such as:
   ```json
   [
     { "jsonrpc": "2.0", "id": 1, "method": "tools/list", "params": {} },
     { "jsonrpc": "2.0", "id": 2, "method": "ping", "params": {} }
   ]
   ```
3. Search emitted checks — there is no check ID for batch rejection (e.g. nothing matching `json-rpc-batch-rejected`).
4. (Optional) Point the suite at a server that incorrectly accepts batch arrays; it still passes with no failure on this requirement.

**Expected behavior**

For `--spec-version` values ≥ `2025-06-18` (`2025-06-18`, `2025-11-25`, draft):

- A scenario should POST a JSON-RPC batch array via raw `fetch` (same pattern as `dns-rebinding.ts` / `stateless.ts` negative probes).
- The server under test should reject it (likely HTTP 400 + JSON-RPC `-32600` Invalid Request — confirm exact wording against the spec diff).
- Conformance should emit a check (e.g. `json-rpc-batch-rejected`) with `FAILURE` if the server accepts or mishandles the batch.
- The reference SDK / everything-server should `SUCCESS` on applicable versions.
- A deliberately broken fixture that accepts batches should fail the check in vitest (per AGENTS.md negative-test pattern).

For `--spec-version 2025-03-26`: the check should **not run** (gate via `introducedIn: '2025-06-18'` or `specVersionAtLeast`). Do not attach this only to draft-only scenarios like `server-stateless`, or 2025-06-18 and 2025-11-25 runs would miss it.

**Logs**

N/A — this is a missing test, not a failing run. No check is emitted today.

**Additional context**

- Wire-level negative tests with precedent: `server-stateless`, `http-standard-headers`, `dns-rebinding`; vitest broken fixtures in `src/scenarios/server/negative.test.ts`.
- SDKs already reject batches by design: [go-sdk#21](https://github.com/modelcontextprotocol/go-sdk/issues/21), [python-sdk#934](https://github.com/modelcontextprotocol/python-sdk/issues/934). [kotlin-sdk#82](https://github.com/modelcontextprotocol/kotlin-sdk/issues/82) still tracks 2025-03-26 batch *support* (backcompat only).
- No open issue in this repo currently tracks batch rejection tests.
- Optional follow-up (separate issue): positive batch-acceptance test for `--spec-version 2025-03-26` backcompat only.
- Stdio batch rejection could follow if stdio conformance lands ([#258](https://github.com/modelcontextprotocol/conformance/issues/258)).
- Spec refs to verify before implementing: [2025-06-18 changelog](https://modelcontextprotocol.io/specification/2025-06-18/changelog#major-changes), [2025-03-26 transports (contrast)](https://modelcontextprotocol.io/specification/2025-03-26/basic/transports), [draft lifecycle/transports](https://modelcontextprotocol.io/specification/draft/basic/lifecycle).

canardleteer:
Closes #378

## Summary

- Adds server scenario `json-rpc-batch-rejection` with check `json-rpc-batch-rejected` for spec versions ≥ `2025-06-18`, exercising the Streamable HTTP transport **MUST** that the POST body be a [single JSON-RPC message](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server) (not a JSON array).
- Probe establishes a stateful session, then POSTs a two-request `[ping, ping]` batch so rejection is tested on batch handling rather than unrelated session routing errors.
- Adds negative fixture `accepts-json-rpc-batch.ts`, unit tests for acceptance/rejection helpers, and an early array guard in `everything-server` so the reference server passes `all-scenarios.test.ts` (documented in code — negative proof remains in the broken fixture per AGENTS.md).

## Test plan

- [x] `npm run build`
- [x] `npm test` (338 tests, including `all-scenarios.test.ts` / `json-rpc-batch-rejection`)
- [x] `npm test -- src/scenarios/server/negative.test.ts` (`json-rpc-batch-rejected` → `FAILURE` against `accepts-json-rpc-batch.ts`)
- [x] `npm test -- src/scenarios/server/json-rpc-batch-rejection.test.ts`
- [x] CLI smoke: `node dist/index.js server --url http://127.0.0.1:<port>/mcp --scenario json-rpc-batch-rejection --spec-version 2025-11-25` against `everything-server` → `SUCCESS`
- [x] CLI run against typescript-sdk conformance server — `node dist/index.js sdk typescript-sdk-v1 --mode server --scenario json-rpc-batch-rejection --spec-version 2025-11-25` → **`json-rpc-batch-rejected` FAILURE** (typescript-sdk v1.x accepts batch arrays at 2025-11-25; intentional Tier-1 signal for that spec version until the SDK fixes batch rejection)
- [x] Confirm scenario absent for `--spec-version 2025-03-26` — CLI: `SKIPPED: … not applicable … (introduced in 2025-06-18)`, exit 0
- [x] `--spec-version draft` run against everything-server → `json-rpc-batch-rejected` **SUCCESS** (stateless draft batch probe)

## Notes for reviewers

- **Spec backing:** MUST-level requirement from transports prose, not changelog narrative alone. Changelog linked as supporting context.
- **Success criteria:** HTTP `4xx` with a single JSON-RPC error object (not strict `-32600`) until SDKs converge on error codes.
- **Tier-1 impact (2025-06-18+ / draft only):** This scenario is gated to spec versions where batch rejection is required; `--spec-version 2025-03-26` skips it entirely. 
  - Tier-1 conformance is evaluated **per target spec version** — when an SDK is scored at **2025-06-18+** (or draft) and its server still accepts batch arrays after session setup, it should **fail** until fixed. 
  - SDKs not ready to reject batches yet can stay in good standing by tier-checking at an older `--spec-version` (or otherwise explicitly opting out of claiming full conformance at 2025-06-18+). The check is active, not pending; the margin is which spec version the SDK holds itself to.
- **everything-server change:** Reference-server alignment for `all-scenarios`; failure proof is the separate negative fixture.

### DIFF
--- a/examples/servers/typescript/accepts-json-rpc-batch.ts
+++ b/examples/servers/typescript/accepts-json-rpc-batch.ts
@@ -1,0 +1,91 @@
+#!/usr/bin/env node
+
+/**
+ * Negative test server that incorrectly accepts JSON-RPC batch arrays.
+ *
+ * AGENTS.md negative-fixture pattern: deliberately broken server in
+ * examples/servers/typescript/, exercised from negative.test.ts (not
+ * everything-server). Proves json-rpc-batch-rejected emits FAILURE when a
+ * server returns 200 with a batch response array.
+ */
+
+import express from 'express';
+
+function handleSingle(body: {
+  id?: number | string | null;
+  method?: string;
+  params?: Record<string, unknown>;
+}) {
+  const id = body.id ?? null;
+  const method = body.method;
+
+  switch (method) {
+    case 'initialize':
+      return {
+        jsonrpc: '2.0' as const,
+        id,
+        result: {
+          protocolVersion:
+            (body.params?.protocolVersion as string | undefined) ??
+            '2025-11-25',
+          capabilities: {},
+          serverInfo: { name: 'accepts-json-rpc-batch', version: '1.0.0' }
+        }
+      };
+    case 'ping':
+      return { jsonrpc: '2.0' as const, id, result: {} };
+    case 'server/discover':
+      return {
+        jsonrpc: '2.0' as const,
+        id,
+        result: {
+          supportedVersions: ['2026-07-28'],
+          capabilities: {},
+          serverInfo: { name: 'accepts-json-rpc-batch', version: '1.0.0' }
+        }
+      };
+    case 'tools/list':
+      return {
+        jsonrpc: '2.0' as const,
+        id,
+        result: { tools: [] }
+      };
+    default:
+      return {
+        jsonrpc: '2.0' as const,
+        id,
+        error: { code: -32601, message: 'Method not found' }
+      };
+  }
+}
+
+const app = express();
+app.use(express.json());
+
+app.post('/mcp', (req, res) => {
+  const body = req.body;
+
+  if (Array.isArray(body)) {
+    const responses = body.map((item) =>
+      handleSingle(
+        typeof item === 'object' && item !== null
+          ? (item as {
+              id?: number | string | null;
+              method?: string;
+              params?: Record<string, unknown>;
+            })
+          : { id: null }
+      )
+    );
+    return res.status(200).json(responses);
+  }
+
+  return res.json(handleSingle(body ?? {}));
+});
+
+const PORT = parseInt(process.env.PORT || '3008', 10);
+app.listen(PORT, '127.0.0.1', () => {
+  console.log(
+    `JSON-RPC batch acceptance negative test server running on http://localhost:${PORT}/mcp`
+  );
+});
--- a/examples/servers/typescript/everything-server.ts
+++ b/examples/servers/typescript/everything-server.ts
@@ -1237,6 +1237,25 @@ const LEGACY_SESSION_PROTOCOL_VERSIONS = [
 
 // Handle POST requests - stateful mode
 app.post('/mcp', async (req, res) => {
+  // AGENTS.md: all-scenarios.test.ts runs every active scenario against
+  // everything-server as the reference "does not false-positive" fixture.
+  // Batch arrays must be rejected from 2025-06-18 onward, but this handler
+  // reads req.body.method before the SDK transport sees the POST body. Without
+  // an explicit array guard, batch probes either hit session routing (-32000)
+  // for the wrong reason or reach the transport and get processed. Failure
+  // proof lives in accepts-json-rpc-batch.ts + negative.test.ts; this guard
+  // keeps the reference server aligned with the json-rpc-batch-rejection check.
+  if (Array.isArray(req.body)) {
+    return res.status(400).json({
+      jsonrpc: '2.0',
+      error: {
+        code: -32600,
+        message: 'Invalid Request: JSON-RPC batch requests are not supported'
+      },
+      id: null
+    });
+  }
+
   const sessionId = req.headers['mcp-session-id'] as string | undefined;
   const reqVersion = req.headers['mcp-protocol-version'] as string | undefined;
   const body = req.body || {};
--- a/examples/servers/typescript/everything-server.ts
+++ b/examples/servers/typescript/everything-server.ts
@@ -1235,29 +1235,44 @@ const LEGACY_SESSION_PROTOCOL_VERSIONS = [
   '2025-11-25'
 ];
 
+// Protocol revisions that MUST support receiving JSON-RPC batches
+// (2025-03-26 base protocol §Batching). Later revisions removed that
+// requirement; the array guard below rejects batches only for those.
+const BATCHING_REQUIRED_PROTOCOL_VERSIONS = new Set([
+  '2024-11-05',
+  '2025-03-26'
+]);
+
 // Handle POST requests - stateful mode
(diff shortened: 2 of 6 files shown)


Math1987:
Thanks for tackling #378. I did a second pass because this check would create a Tier-1 `FAILURE`. At `0cf3fe0`, the full local test suite passes (33 files / 361 tests) and `npm run lint` passes. I also exercised two paths that the PR does not currently cover.

### Normative contract first

The cited [2025-06-18 transport requirement](https://modelcontextprotocol.io/specification/2025-06-18/basic/transports#sending-messages-to-the-server) says that the POST body **MUST be a single** JSON-RPC message. In context, that constrains the body sent by the client. I could not find a corresponding “server MUST reject a batch array” requirement, nor a prescribed HTTP status or JSON-RPC error shape.

The history also reflects this ambiguity: the original specification PR is titled [“Remove batching requirement” (#416)](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/416), while later clarification attempts [#828](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/828) and [#1008](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1008) raised the same “MUST reject vs MAY tolerate” question but were closed without a normative clarification.

@pcarleton, since this check would affect Tier-1 conformance, could you confirm whether a server that tolerates batch input as a superset is itself non-conformant? If the requirement is client-side only, I do not think a server `FAILURE` check is justified without first clarifying the specification.

### Implementation blockers if server-side rejection is intended

1. **Optional sessions are treated as mandatory, and the lifecycle is incomplete.** `establishStatefulSession()` throws when `initialize` returns no `Mcp-Session-Id`, although servers only **MAY** mint one. It also never sends the required `notifications/initialized` notification before the ping batch. I reproduced the first false failure against the repository’s existing SDK-backed no-session fixture at `2025-11-25`: the scenario fails with `initialize did not return Mcp-Session-Id header` before testing any array. A clean probe would complete the lifecycle, retain the session ID only when present, then send a batch of identical `ping` requests.

2. **The rejection predicate adds a response contract that the cited text does not define.** `isBatchRejected()` requires both HTTP 4xx and a numeric JSON-RPC error. That false-fails a bodyless HTTP rejection, while any unrelated 4xx JSON-RPC error can false-pass. The unit test explicitly treats `-32000 Invalid or missing session ID` as proof of batch rejection, which demonstrates the attribution problem. The accepted rejection shapes need a maintainer decision before they can be encoded safely.

3. **The draft probe is confounded by `Mcp-Method`.** It sends `Mcp-Method: server/discover` for an array containing both `server/discover` and `tools/list`. A server that incorrectly processes arrays but correctly enforces the draft header/body matching rule can reject this as `HeaderMismatch`, and the scenario records `SUCCESS`. Two identical `server/discover` requests would isolate the array dimension.

4. **The reference server regresses 2025-03-26.** The new top-level `Array.isArray(req.body)` guard is unconditional. After a valid `2025-03-26` initialization and `notifications/initialized`, I sent a ping batch to the PR’s `everything-server` and received HTTP 400 / `-32600`. However, the [2025-03-26 base protocol](https://modelcontextprotocol.io/specification/2025-03-26/basic#batching) explicitly says implementations **MUST support receiving JSON-RPC batches**. Gating the new scenario does not gate this global server change.

I would add regression coverage for a valid no-session server, a bodyless rejection, a batch-capable draft server that enforces `Mcp-Method`, and preservation of 2025-03-26 batch behavior. I’m happy to help turn the agreed contract into a focused patch.

AI-assistance disclosure: This review was prepared with OpenAI Codex, including repository/specification analysis, local test execution, and drafting. I reviewed and approved the final comment before posting.

canardleteer:
Thanks for the review @Math1987 !

All I can really do without additional specification clarification, is remove my guard for "The reference server regresses 2025-03-26." 

Turning our rejection into a ">=2025-06-18" one.

Issue #296, opened 2026-05-20
SEP-2575: remaining conformance tests


pcarleton:
Status as of the [`70b2800d` traceability refresh](https://github.com/modelcontextprotocol/conformance/pull/292) (run against `typescript-sdk@main`). 8 of 26 declared requirements are tested, 18 are untested. Of the untested 18, 7 have a scenario already written but the reference SDK can't run it yet, and 11 have no scenario at all.

### A. Scenario written — blocked on `typescript-sdk` support

[`src/scenarios/client/request-metadata.ts`](https://github.com/modelcontextprotocol/conformance/blob/main/src/scenarios/client/request-metadata.ts) (#270) emits all 7 of these, but `everythingClient.ts` on `typescript-sdk@main` has no `request-metadata` handler — it logs `Unknown scenario` and exits before any HTTP request is made, so no checks are emitted. These will flip to `tested` once the SDK has a SEP-2575-aware conformance client, or we fix the checker to be more correct.

So no scenario action needed on these, this is a tracking bug.

| Check | Requirement | Level |
|---|---|---|
| `sep-2575-client-populates-meta` | Every client request includes `protocolVersion`, `clientInfo`, `clientCapabilities` in `_meta` | MUST |
| `sep-2575-http-client-sends-version-header` | Every POST includes `MCP-Protocol-Version` header | MUST |
| `sep-2575-http-version-header-matches-meta` | Header value matches `_meta.…/protocolVersion` | MUST |
| `sep-2575-client-declares-roots-capability` | Roots-capable clients declare `roots` in `_meta.…/clientCapabilities` | MUST (cond.) |
| `sep-2575-client-declares-sampling-capability` | Sampling-capable clients declare `sampling` | MUST (cond.) |
| `sep-2575-client-declares-elicitation-capability` | Elicitation-capable clients declare `elicitation` | MUST (cond.) |
| `sep-2575-client-retry-supported-version` | Client retries with a mutually-supported version after `UnsupportedProtocolVersionError` | SHOULD |

### B. No scenario yet — seems straightforward to write

| Check | Requirement | Level | Notes |
|---|---|---|---|
| `sep-2575-server-sends-subscription-ack` | `notifications/subscriptions/acknowledged` is the first message on a `subscriptions/listen` stream | MUST | Open a listen stream, assert first frame |
[...]

PR #383, opened 2026-07-02
fix(server-stateless): declare the elicitation capability on the streaming-probe request

### LINKED ISSUE #382: server-stateless: `sep-2575-http-server-no-independent-requests-on-stream` probe never declares the elicitation capability its own fixture requires
**Describe the bug**
The `server-stateless` scenario builds one shared request `_meta` envelope with
`"io.modelcontextprotocol/clientCapabilities": {}` (src/scenarios/server/stateless.ts, `validMeta`)
and reuses it for every check. The `sep-2575-http-server-no-independent-requests-on-stream` check
drives its streaming probe by calling `tools/call` with `name: "test_streaming_elicitation"` under
that empty-capabilities envelope.

Per SEP-2575, a server MUST reject a request that exercises a capability the client did not declare
with `-32021 MissingRequiredClientCapability` — the exact behavior the sibling check
`sep-2575-server-rejects-undeclared-capability` (same scenario, same shared envelope, fixture
`test_missing_capability`) verifies and rewards two checks earlier. So a spec-correct,
capability-enforcing server necessarily answers the streaming probe with `-32021`, and the check
can never be exercised against it.

Before #372 this reported SKIPPED; since #372 (alpha.8) it scores a hard failure:

> the 'test_streaming_elicitation' call was rejected (code -32021), so the response stream could
> not be exercised

The scenario is now internally contradictory: one check requires the `-32021` rejection that breaks
the other.

**To Reproduce**
1. Run `server-stateless` at `0.2.0-alpha.9` against any server that enforces per-request capability
declaration (observed with the PHP SDK, logiscape/mcp-sdk-php v2 `main`, whose everything-server
exposes the `test_streaming_elicitation` fixture): 27/28 checks pass;
`sep-2575-http-server-no-independent-requests-on-stream` fails with the message above. The only
ways to "pass" today are to strip the capability requirement from the fixture (defeating its
purpose of demonstrating SEP-2322 elicitation-as-`InputRequiredResult`) or to stop enforcing
capability declaration (failing `sep-2575-server-rejects-undeclared-capability`).

**Expected behavior**
Declare the capability the probe needs, at the probe site only — the shared `validMeta` must stay
empty because `sep-2575-server-rejects-undeclared-capability` depends on it:

```ts
_meta: { ...validMeta, 'io.modelcontextprotocol/clientCapabilities': { elicitation: {} } }
```

This matches the intent recorded in tracking issue #296 for this check ("Call a tool that needs
sampling/elicitation, scan stream frames") and strictly widens the set of servers the check can
exercise: capability-enforcing servers become testable, and servers that ignore capability
declarations behave as before.

**Additional context**
Draft PR #351 refactors this same probe (shares the stream with a new check).

I have a fix with a self-test ready and verified against a real SDK per CONTRIBUTING (both
outcomes: the untestable failure without the change, 28/28 with it) and will open a PR referencing
this issue.

logiscapedev:
Fixes #382

## What

`sep-2575-http-server-no-independent-requests-on-stream` drives its streaming probe by calling
`tools/call` `test_streaming_elicitation` under the scenario's shared `validMeta` envelope, whose
`io.modelcontextprotocol/clientCapabilities` is intentionally empty. The fixture legitimately needs
the elicitation capability, so a capability-enforcing server MUST reject the probe with `-32021
MissingRequiredClientCapability` — the exact behavior `sep-2575-server-rejects-undeclared-capability`
requires and rewards earlier in the same scenario. Since #372 turned missing-prerequisite skips
into hard failures, one check requires the rejection that makes the other untestable.

This PR declares the capability at the probe site only:

```ts
_meta: {
  ...validMeta,
  'io.modelcontextprotocol/clientCapabilities': { elicitation: {} }
}
```

`validMeta` itself stays empty — `sep-2575-server-rejects-undeclared-capability` depends on it.
This matches the intent recorded for this check in #296 ("Call a tool that needs
sampling/elicitation, scan stream frames") and strictly widens coverage: capability-enforcing
servers become testable; servers that ignore capability declarations behave exactly as before.

The sibling probe `sep-2575-server-no-log-without-loglevel` (`test_logging_tool`) was audited for
the same latent problem: its fixture keys off `_meta.../logLevel`, not a client capability, so it
needs no change.

## Self-tests (prove it passes and fails)

Two tests added to `stateless.test.ts`, modeled on the existing mock patterns:

- **Green:** a spec-correct, capability-enforcing mock (rejects `test_missing_capability` without
  `sampling` and `test_streaming_elicitation` without `elicitation`, each with a spec-shaped
  `-32021`) now passes BOTH formerly-contradictory checks. Against `main` without the fix this test
  fails exactly as real servers do: the stream check reads
  `FAILURE — Not testable: the 'test_streaming_elicitation' call was rejected (code -32021)`.
- **Guard:** a server that rejects the streaming probe despite the declared capability still
  produces the red untestable outcome, not a vacuous SUCCESS.

`npm test` 333/333, `npm run check` clean.

## Real-SDK evidence (per CONTRIBUTING)

Run against the PHP SDK (logiscape/mcp-sdk-php v2 `main`), whose everything-server enforces
per-request capability declaration:

- **Before:** 27/28 — `sep-2575-http-server-no-independent-requests-on-stream` fails:
  `Not testable: the 'test_streaming_elicitation' call was rejected (code -32021), so the response
  stream could not be exercised`
- **After:** 28/28.

No regression against this repo's own example server (`examples/servers/typescript`): 30/30
before and after (it does not enforce capabilities on this fixture, so the added declaration is
inert there).

## Note on sequencing

Draft PR #351 hoists/shares this same probe stream between checks; this fix is written against
current `main` and is a one-expression change at the probe site, so whichever lands first, the
other rebases trivially. Happy to coordinate.

### DIFF
--- a/src/scenarios/server/stateless.test.ts
+++ b/src/scenarios/server/stateless.test.ts
@@ -628,6 +628,135 @@ describe('Stateless Server Scenario Negative Tests', () => {
     expect(independentRequestCheck?.status).toBe('FAILURE');
   });
 
+  // A spec-correct server that enforces per-request capability declaration:
+  // test_missing_capability requires sampling and test_streaming_elicitation
+  // requires elicitation; each is rejected with a spec-shaped -32021 unless
+  // the request's _meta declares the capability. A rejection reaches
+  // listenToStream as the parsed JSON error body — a single error frame.
+  function capabilityEnforcingMock(
+    streamingCallFrames: (reqBody: any) => any[]
+  ) {
+    return mockFetchTarget((reqBody) => {
+      if (reqBody.method === 'server/discover') {
+        return discoverResponse(reqBody, { tools: {} }, 'capability-enforcer');
+      }
+      if (reqBody.method === 'tools/list') {
+        return {
+          status: 200,
+          body: {
+            jsonrpc: '2.0',
+            id: reqBody.id,
+            result: {
+              tools: [
+                { name: 'test_missing_capability' },
+                { name: 'test_streaming_elicitation' }
+              ]
+            }
+          }
+        };
+      }
+      if (reqBody.method === 'tools/call') {
+        const declaredCaps =
+          reqBody.params?._meta?.[
+            'io.modelcontextprotocol/clientCapabilities'
+          ] ?? {};
+        if (reqBody.params?.name === 'test_missing_capability') {
+          if (!declaredCaps.sampling) {
+            return {
+              status: 400,
+              body: {
+                jsonrpc: '2.0',
+                id: reqBody.id,
+                ...spec32021({ sampling: {} })
+              }
+            };
+          }
+          return {
+            status: 200,
+            body: {
+              jsonrpc: '2.0',
+              id: reqBody.id,
+              result: { content: [{ type: 'text', text: 'ok' }] }
+            }
+          };
+        }
+        if (reqBody.params?.name === 'test_streaming_elicitation') {
+          if (!declaredCaps.elicitation) {
+            return {
+              isStream: true,
+              status: 400,
+              frames: [
+                {
+                  jsonrpc: '2.0',
+                  id: reqBody.id,
+                  ...spec32021({ elicitation: {} })
+                }
+              ]
+            };
+          }
+          return {
+            isStream: true,
+            status: 200,
+            frames: streamingCallFrames(reqBody)
+          };
+        }
+      }
+    });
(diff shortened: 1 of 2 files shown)


pcarleton (approved):

pcarleton:
thanks for this @logiscapedev !

PR #394, opened 2026-07-07
Client Auth: DPoP Sender-Constrained Tokens (SEP-1932)

PieterKas:
## Overview

Conformance tests for **DPoP sender-constrained tokens (SEP-1932 / RFC 9449)** on the **MCP client** side.

This is the first of three DPoP PRs (client, MCP server, authorization server) and the **shared foundation**: it introduces the DPoP-capable test authorization server and the resource-side proof "judge" that the server and authorization-server PRs reuse. **Please review this one first.**

Closes #368.

## What it tests

That an MCP client correctly obtains and presents a DPoP-bound access token:

- **Token request** — the client sends a valid DPoP proof at the token endpoint (the prerequisite for a sender-constrained token).
- **Authorization scheme** — the client presents the token with the `DPoP` scheme, not `Bearer`.
- **Fresh proof** — a fresh, well-formed proof accompanies each request (unique `jti`).

Because server-provided nonces are **optional** in RFC 9449, the scenario runs in two postures:

- **`auth/dpop`** — the common **nonce-less** baseline.
- **`auth/dpop-nonce`** — the authorization server and MCP server both require a nonce (§8/§9), additionally checking that the client retries with the server-supplied nonce.

Each check is proven both to pass (compliant example client) and to fail (deliberately-broken example clients), with an automated acceptance suite.

## Scope

Client behaviour only. The test authorization server and test MCP server are the judges — they observe and score; they don't exercise their own enforcement here (that's PR #369 / #370).

## Notes

- Registered as **draft / not-scored** (SEP-1932 is a draft extension).
- Validated against the bundled example fixtures. No reference SDK implements DPoP yet, so there is no real-SDK run to report.

Issue #369, opened 2026-06-25
Server Auth: DPoP Proof Validation (SEP-1932)


PieterKas:
## Overview

[SEP-1932](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1932)
adopts OAuth 2.0 Demonstrating Proof of Possession
([RFC 9449](https://www.rfc-editor.org/rfc/rfc9449.html)) as an optional MCP
authorization extension for sender-constrained access tokens. When a client
presents a DPoP-bound token, the **MCP server (resource server)** must verify a
DPoP proof on every request before granting access.

This issue covers **MCP server** conformance only. It validates that a server
acting as an OAuth 2.1 resource server correctly applies the RFC 9449 §4.3 proof
checking steps, enforces the ±5-minute `iat` acceptance window, returns correct
`401` / `WWW-Authenticate: DPoP` challenges, and continues to enforce the baseline
token audience requirements.

Key properties of the server role:

- The framework drives the server **as a DPoP client**, presenting valid proofs
  for positive checks and deliberately malformed proofs for negative checks.
- The proposal adopts DPoP **as defined in RFC 9449** — no MCP-specific extensions
  to proof validation are expected.
- A conformant stateless server provides replay protection via the `iat` window
  and standard claim validation; `jti` tracking and nonces are optional.

## Specification References

- SEP-1932 PR: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1932
- Detailed proposal: https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
- RFC 9449 — OAuth 2.0 Demonstrating Proof of Possession (DPoP): https://www.rfc-editor.org/rfc/rfc9449.html
  - §4.2 DPoP Proof JWT Syntax, **§4.3 Checking DPoP Proofs** (the core checklist)
  - §6 Public Key Confirmation (`cnf`/`jkt`), §7 DPoP-bound access tokens
  - §7.1 The DPoP Authentication Scheme and `WWW-Authenticate` challenge
  - §9 Resource Server-Provided Nonce, §11.1 DPoP Proof Replay
- Baseline MCP Authorization specification (resource-server token validation and
  audience binding): https://modelcontextprotocol.io/specification/draft/basic/authorization

## Scope

**In scope — what the MCP/resource server does:**

- Requiring the `DPoP` `Authorization` scheme and a `DPoP` proof header on
  protected requests.
[...]

PR #396, opened 2026-07-07
Authorization Server: DPoP Support (SEP-1932)

### LINKED ISSUE #370: Authorization Server Auth: DPoP Token Binding (SEP-1932)
## Overview

[SEP-1932](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1932)
adopts OAuth 2.0 Demonstrating Proof of Possession
([RFC 9449](https://www.rfc-editor.org/rfc/rfc9449.html)) as an optional MCP
authorization extension. For sender-constrained tokens to work, the **OAuth
authorization server** must accept a DPoP proof at the token endpoint, bind the
issued access token to the client's public key, and advertise its DPoP support.

This issue covers **authorization server** conformance only. It validates that an
authorization server advertises `dpop_signing_alg_values_supported`, validates the
token-endpoint DPoP proof, issues a DPoP-bound token (`token_type=DPoP` with a
`cnf.jkt` confirmation), honours the optional nonce mechanism, and enforces
`dpop_bound_access_tokens` client registration.

Key properties of the authorization-server role:

- The framework drives the authorization server **as a DPoP client** at the token
  endpoint, presenting valid and invalid proofs.
- DPoP is adopted **as defined in RFC 9449** — no MCP-specific extensions to the
  token-endpoint exchange.
- This is distinct from client and resource-server conformance even though all
  three involve OAuth; the authorization server is tested in isolation.

## Specification References

- SEP-1932 PR: https://github.com/modelcontextprotocol/modelcontextprotocol/pull/1932
- Detailed proposal: https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
- RFC 9449 — OAuth 2.0 Demonstrating Proof of Possession (DPoP): https://www.rfc-editor.org/rfc/rfc9449.html
  - §4.3 Checking DPoP Proofs (token-endpoint validation), §5 DPoP Access Token Request
  - §6 Public Key Confirmation (`cnf`/`jkt`)
  - §5.1 `dpop_signing_alg_values_supported`, §5.2 `dpop_bound_access_tokens`
  - §8 Authorization Server-Provided Nonce
- RFC 8414 — OAuth 2.0 Authorization Server Metadata: https://datatracker.ietf.org/doc/html/rfc8414
- Baseline MCP Authorization specification: https://modelcontextprotocol.io/specification/draft/basic/authorization

## Scope

**In scope — what the authorization server does:**

- Advertising `dpop_signing_alg_values_supported` (asymmetric algorithms only; no
  `none`) in its authorization server metadata.
- Validating the DPoP proof presented on the token request (§4.3 as applicable to
  the token endpoint: `typ`, `alg`, signature, `jwk` without private key,
  `htm=POST`, `htu`=token endpoint, `jti`, `iat` window, `nonce` when required).
- Issuing a DPoP-bound access token: `token_type` of `DPoP` and a `cnf.jkt`
  confirmation equal to the base64url JWK SHA-256 thumbprint of the proof key.
- Token-endpoint nonce mechanism: `400 use_dpop_nonce` + `DPoP-Nonce`, then
  accepting the retried request.
- Honouring `dpop_bound_access_tokens=true` client registration by rejecting token
  requests that lack a DPoP proof. _(Excluded — requires dynamic client registration, out of scope for these DPoP scenarios; see #396.)_
- Binding refresh tokens to the same key for public clients.  _(Deferred — not testable until the shared conformance test AS supports refresh tokens; tracked as a follow-up. See #396.)_

**Not in scope (covered elsewhere or by another role):**

- Resource-server proof validation and `ath`/`cnf` agreement at resource access —
  covered by the server conformance issue.
- Client proof construction and nonce-retry behaviour — covered by the client
  conformance issue.

### Conformance harness (framework acting as a DPoP client at the token endpoint)

- The framework, acting as a DPoP client, performs a token request with a
  DPoP proof against the authorization server under test, then inspects the
  metadata, the token response (`token_type`), and the issued token's `cnf.jkt`.
- Fixture generation for a valid proof keyed to a generated key pair, plus crafted
  invalid proofs (one per token-endpoint failure mode).

### Helpers (`helpers/`)

- DPoP proof builder with per-field overrides (reused from / shared with the
  server scenario helpers where practical).
- JWK SHA-256 thumbprint helper to assert `cnf.jkt` equals the proof key
  thumbprint.
- Metadata fetch/parse helper for `dpop_signing_alg_values_supported`.

### Scenario (`src/scenarios/authorization-server/`)

- A single scenario file implementing all checks below, registered in the
  authorization-server scenario list.

### Acceptance test suite

- Helper unit tests and scenario acceptance tests asserting each check passes for
  a conformant authorization server and fails for a deliberately non-conformant
  one.

### Components that do **not** change

- The resource-server (MCP server) path — the grant/token type does not change how
  a resource server validates an access token's audience.
- OIDC discovery infrastructure beyond surfacing the new metadata field.

### Positive

- [ ] Metadata advertises `dpop_signing_alg_values_supported` as a JSON array of
      asymmetric JWS `alg` values; `none` is not present.
- [ ] Token endpoint accepts a valid DPoP proof (`typ=dpop+jwt`, asymmetric `alg`,
      embedded public `jwk`, `htm=POST`, `htu`=token endpoint, `jti`, `iat` in
      window) and issues a token.
- [ ] Issued token response sets `token_type` to `DPoP`.
- [ ] Issued token carries a `cnf.jkt` equal to the base64url JWK SHA-256
      thumbprint of the proof's public key. _(Verified for JWT access tokens; opaque/reference tokens are out of scope — `cnf.jkt` isn't observable without introspection; see #396.)_
- [ ] For a public client, an issued refresh token is bound to the same key. _(Deferred to a follow-up. The shared test AS does not yet support refresh tokens; see #396.)_

### Negative (token-endpoint proof failures → `400 invalid_dpop_proof` unless noted)

- [ ] Proof signature does not verify against the embedded `jwk`.
- [ ] `typ` is not `dpop+jwt`.
- [ ] `alg` is `none` or a symmetric algorithm.
- [ ] `jwk` contains a private key.
- [ ] `htm`/`htu` do not match the token endpoint request.
- [ ] `iat` is outside the acceptable window.
- [ ] A required claim is missing (`jti`, `htm`, `htu`, `iat`).

### Nonce behaviour (if the authorization server requires nonces)

- [ ] Returns `400 use_dpop_nonce` + `DPoP-Nonce` when a nonce is required and
      absent.
- [ ] Accepts the retried token request carrying the matching `nonce` claim.
- [ ] Rejects a proof whose `nonce` does not match a recently supplied value.

### Client registration enforcement

- [ ] With `dpop_bound_access_tokens=true`, the authorization server rejects a
      token request that does not include a DPoP proof. _(Excluded — depends on dynamic client registration; see #396.)_

## Acceptance Criteria

- [ ] A single scenario file in `src/scenarios/authorization-server/` implements
      all checks above (one scenario, many checks).
- [ ] Each check has both a passing case (conformant authorization server) and a
      deliberate failing case (crafted invalid proof / missing field) proven by the
      automated acceptance test suite — no check is vacuously passing.
- [ ] Helper unit tests cover the proof builder, thumbprint, and metadata helpers.
- [ ] The acceptance test suite runs as part of `npm test`.
- [ ] Scenario runs through the standard CLI runner; no parallel entry point is
      introduced.
- [ ] Validated against at least one real authorization server implementation
      before the PR is submitted; SDK/AS baseline YAMLs updated where existing
      implementations do not yet support DPoP.

## Out of Scope

- Nonce **cryptographic construction** (e.g. AEAD-encrypted timestamps) — an
  implementation choice; only the observable `use_dpop_nonce` protocol is tested.
- Full OAuth 2.1 / token-endpoint conformance unrelated to DPoP (PKCE, resource
  indicators, etc.) — covered by existing baseline authorization conformance.
- Resource-server and client behaviour — covered by the separate server and
  client conformance issues.

## Notes
Prepared with the help of Claude (Opus 4.8)

PieterKas:
> ℹ️ **Update:** the client foundation PR (#394) is now merged and this branch has been rebased onto `main`, so the diff below contains only the authorization-server-side changes.

## Overview

Conformance tests for **DPoP support (SEP-1932 / RFC 9449)** on the **authorization server** side.

Builds on the shared DPoP test helpers introduced in the client PR (#394, merged).

Closes #370.

## What it tests

That an authorization server correctly advertises and implements DPoP. The scenario drives a real authorization-code + PKCE flow and checks the server:

- **Advertises** `dpop_signing_alg_values_supported` in its metadata (RFC 9449 §5.1).
- **Lists only asymmetric algorithms** — no `none`, no symmetric algorithms.
- **Binds the issued token** to the client's DPoP key (`cnf.jkt` matches the presented proof, `token_type: DPoP`).

The scenario is **support-gated**: an authorization server that doesn't advertise DPoP is not a DPoP server, so it's skipped rather than failed.

Each check is proven to pass (compliant example authorization server) and to fail (misbehaving configurations), with an automated acceptance suite.

## Scope

Authorization-server behaviour only — metadata and token issuance. Proof validation at the resource is covered by the MCP-server PR (#369); client presentation by the client PR (#394).

## Notes

- SEP-1932 is a draft extension.
- Validated against the bundled example fixtures. No reference SDK implements DPoP yet, so there is no real-SDK run to report.

### DIFF
--- a/src/scenarios/authorization-server/dpop.ts
+++ b/src/scenarios/authorization-server/dpop.ts
@@ -94,6 +94,37 @@ const SUPPORTED_PROOF_ALGS = [
   'EdDSA'
 ];
 
+/**
+ * Pick a proof-signing algorithm the harness can produce that the AS also
+ * advertises (RFC 9449 §5.1), given `dpop_signing_alg_values_supported`.
+ *
+ * Returns null (→ the caller SKIPs the binding check) when the value is a
+ * non-empty array with no algorithm we support, OR any present-but-non-array
+ * shape — a string, `null`, number, or object are all malformed metadata, not
+ * "unspecified", so we must not fall back to ES256 (which the AS would reject,
+ * mis-scoring binding). Only an EMPTY array falls back to ES256 as a best-effort
+ * to still exercise the binding (an empty list is itself flagged by the metadata
+ * check). An absent field never reaches here — the scenario's support gate SKIPs
+ * the whole scenario upstream — but is treated as the empty case for safety.
+ */
+export function negotiateProofAlg(
+  advertised: unknown,
+  supported: readonly string[] = SUPPORTED_PROOF_ALGS
+): string | null {
+  // Any present-but-non-array value (string / null / number / object) is
+  // malformed — SKIP rather than fall back to ES256.
+  if (advertised !== undefined && !Array.isArray(advertised)) {
+    return null;
+  }
+  if (Array.isArray(advertised) && advertised.length > 0) {
+    const match = advertised.find(
+      (a) => typeof a === 'string' && supported.includes(a)
+    );
+    return typeof match === 'string' ? match : null;
+  }
+  return 'ES256';
+}
+
 /** Strip query + fragment from a URL for use as an `htu` (RFC 9449 §4.2). */
 function stripUrlQuery(url: string): string {
   try {
@@ -393,34 +424,9 @@ browser login + callback for login-gated servers.`;
     }
   }
 
-  /**
-   * Pick a proof-signing algorithm the harness can produce that the AS also
-   * advertises (RFC 9449 §5.1). Returns null when the AS advertises a non-empty
-   * list with no algorithm we support, OR a present-but-non-array (malformed)
-   * value — the caller then SKIPs rather than sending an unadvertised alg (e.g.
-   * defaulting to ES256) that the AS would legitimately reject and we'd mis-score
-   * as a binding failure. Only an absent or empty list (itself flagged by the
-   * metadata check) falls back to ES256 as a best effort to still exercise the
-   * binding.
-   */
+  /** See the module-level {@link negotiateProofAlg}. */
   private negotiateProofAlg(metadata: Record<string, any>): string | null {
-    const advertised = metadata.dpop_signing_alg_values_supported;
-    // A present-but-non-array value (e.g. the string "RS256") is malformed
-    // metadata, not "unspecified" — SKIP rather than fall back to ES256.
-    if (
-      advertised !== undefined &&
-      advertised !== null &&
-      !Array.isArray(advertised)
-    ) {
-      return null;
-    }
-    if (Array.isArray(advertised) && advertised.length > 0) {
-      const match = advertised.find(
-        (a) => typeof a === 'string' && SUPPORTED_PROOF_ALGS.includes(a)
-      );
-      return typeof match === 'string' ? match : null;
-    }
-    return 'ES256';
+    return negotiateProofAlg(metadata.dpop_signing_alg_values_supported);
   }
 
   /**
--- a/src/scenarios/authorization-server/dpop.ts
+++ b/src/scenarios/authorization-server/dpop.ts
@@ -396,8 +396,8 @@ browser login + callback for login-gated servers.`;
   /**
    * Pick a proof-signing algorithm the harness can produce that the AS also
    * advertises (RFC 9449 §5.1). Returns null when the AS advertises a non-empty
-   * list with no algorithm we support (including a non-empty but malformed list)
-   * — the caller then SKIPs rather than sending an unadvertised alg (e.g.
+   * list with no algorithm we support, OR a present-but-non-array (malformed)
+   * value — the caller then SKIPs rather than sending an unadvertised alg (e.g.
    * defaulting to ES256) that the AS would legitimately reject and we'd mis-score
    * as a binding failure. Only an absent or empty list (itself flagged by the
    * metadata check) falls back to ES256 as a best effort to still exercise the
@@ -405,6 +405,15 @@ browser login + callback for login-gated servers.`;
    */
   private negotiateProofAlg(metadata: Record<string, any>): string | null {
     const advertised = metadata.dpop_signing_alg_values_supported;
+    // A present-but-non-array value (e.g. the string "RS256") is malformed
+    // metadata, not "unspecified" — SKIP rather than fall back to ES256.
+    if (
+      advertised !== undefined &&
+      advertised !== null &&
+      !Array.isArray(advertised)
+    ) {
+      return null;
+    }
     if (Array.isArray(advertised) && advertised.length > 0) {
       const match = advertised.find(
         (a) => typeof a === 'string' && SUPPORTED_PROOF_ALGS.includes(a)
--- a/src/scenarios/authorization-server/dpop.ts
+++ b/src/scenarios/authorization-server/dpop.ts
@@ -1,0 +1,712 @@
+/**
+ * DPoP authorization-server scenario (SEP-1932 / RFC 9449).
+ *
+ * The framework acts as a DPoP-capable OAuth client against the authorization
+ * server under test at `options.url`. It probes:
+ *
+ *  - metadata: `dpop_signing_alg_values_supported` is advertised (RFC 9449 §5.1)
+ *    and does not include the `none` or symmetric algorithms;
+ *  - token binding: a code exchanged WITH a DPoP proof yields a token bound to
+ *    the proof key (`cnf.jkt`) with `token_type: DPoP` (RFC 9449 §5–§6).
+ *
+ * An AS that does not advertise `dpop_signing_alg_values_supported` is not a
+ * DPoP authorization server (RFC 9449 §5.1 is how support is signalled), so the
+ * whole scenario SKIPs rather than failing it — the DPoP checks only apply once
+ * the AS opts in. (`dpop_bound_access_tokens` is per-client registration
+ * metadata, RFC 9449 §5.2, not an AS capability, so no enforcement check is
+ * made here.)
+ *
+ * Tokens are obtained via the authorization_code + PKCE grant (the MCP grant).
+ * The authorization step auto-follows a direct redirect to the registered
+ * redirect_uri (the headless path used by auto-approving/test ASs) and falls
+ * back to an interactive browser + callback-server wait for login-gated ASs.
+ *
+ * Emits the sep-1932-as-* check IDs declared in src/seps/sep-1932.yaml.
+ */
+
+import {
+  CheckStatus,
+  ClientScenarioForAuthorizationServer,
+  ConformanceCheck,
+  DRAFT_PROTOCOL_VERSION,
+  SpecReference
+} from '../../types';
+import { AuthorizationServerOptions } from '../../schemas';
+import { request } from 'undici';
+import { createHash, randomBytes } from 'crypto';
+import { startCallbackServer } from './auth/helpers/createCallbackServer';
+import {
+  generateDpopKeyPair,
+  buildDpopProof
+} from '../client/auth/helpers/dpopProof';
+import { readTokenBinding } from '../client/auth/helpers/dpopToken';
+import { SpecReferences } from './auth/spec-references';
+
+const REDIRECT_URI_ORIGIN = 'http://127.0.0.1';
+const REDIRECT_URI_PATH = '/callback';
+
+/** Static id → (name, description, spec references) for each emitted check. */
+const CHECK_DEFS: Record<
+  string,
+  { name: string; description: string; specReferences: SpecReference[] }
+> = {
+  'sep-1932-as-metadata-alg-values': {
+    name: 'DpopMetadataAlgValues',
+    description:
+      'Authorization server metadata advertises dpop_signing_alg_values_supported',
+    specReferences: [
+      SpecReferences.SEP_1932_DPOP,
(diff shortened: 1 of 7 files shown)


Michito-Okai:
@PieterKas 
Could you please resolve the conflicts? Also, would it be possible to share the test execution procedure, test scenarios, and the check items?

PieterKas:
@Michito-Okai - I believe the conflicts are resolved.

src/seps/sep-1932.yaml
```diff
@@ -21,8 +21,6 @@ requirements:
     text: 'Authorization servers supporting DPoP MUST include the `dpop_signing_alg_values_supported` field in their Authorization Server Metadata as defined in RFC 9449 Section 5.1. This field MUST contain a JSON array listing the JWS algorithm values supported for DPoP proof JWTs'
   - check: sep-1932-as-no-none-alg
     text: 'Only asymmetric signature algorithms are permitted; the `none` algorithm MUST NOT be included'
-  - check: sep-1932-as-dpop-bound-enforcement
-    text: 'When `dpop_bound_access_tokens` is set to `true`, the authorization server MUST reject token requests from the client that do not include a valid DPoP proof'
   - check: sep-1932-as-token-binding
```
tnorimat, line 25: This part (sep-1932-as-token-binding) was not added/modified by the PR, but its following text is not described in the "spec_url" referring the spec  https://github.com/modelcontextprotocol/ext-auth/blob/pieterkas-dpop-extension/specification/draft/dpop-extension.mdx
It might be good to show where the text is coming.

src/scenarios/authorization-server/dpop.ts
```diff
@@ -0,0 +1,727 @@
...
+            errorMessage: 'Token response was 200 but carried no access_token',
+            details: { tokenType: binding.tokenType ?? null }
+          })
+        );
+        return;
+      }
+      // Only inconclusive when the AS CLAIMS a DPoP binding (token_type=DPoP)
+      // but the token is opaque: cnf.jkt can't be read off the wire (it may
+      // still hold, verifiable only via introspection) → documented harness gap
+      // → SKIP. A non-DPoP token_type is a plain binding failure below, opaque
```
tnorimat, line 392: Theoretically, an access token is opaque so that some real authorization server issues an reference type access token.
Therefore, in this case, it might be good to do token introspection and verify its response by following https://www.rfc-editor.org/rfc/rfc9449.html#section-6.2 .

Regarding that, I think there are the following options. Either one is fine.:
 1. the PR does not handle it. write the comment in the issue showing that non-JWT access token is out of scope.
 2. The PR does handle it.

src/scenarios/authorization-server/dpop.ts
```diff
@@ -0,0 +1,727 @@
...
+      // or not, so it does not reach here.
```
tnorimat, line 393: It appears that this pull request does not cover some part of the corresponding issue https://github.com/modelcontextprotocol/conformance/issues/370 .

According to the "In Scope" in the issue, it states that "Binding refresh tokens to the same key for public clients".
And according to the "Checks to Cover - Positive" in the issue, it states that "For a public client, an issued refresh token is bound to the same key."

However, the PR does not test this point.

Regarding that, I think there are the following options. Either one is fine.:
 1. the PR does not handle it. write the comment in the issue stating that a public client is out of scope so that it does not check if a refresh token is bound to the same key.
 2. The PR does handle it.

src/seps/sep-1932.yaml
```diff
@@ -32,6 +30,8 @@ requirements:
   - check: sep-1932-server-audience-validation
     text: 'MCP servers MUST continue to validate that access tokens were specifically issued for them, even when DPoP is used'
 
+  - text: 'When `dpop_bound_access_tokens` is set to `true`, the authorization server MUST reject token requests from the client that do not include a valid DPoP proof'
```
tnorimat, line 34: It appears that this pull request does not cover some part of the corresponding issue #370 .

According to the "Check to Cover - Client registration enforcement", it states that "With dpop_bound_access_tokens=true, the authorization server rejects a
token request that does not include a DPoP proof.".

However, the PR does not test it (excluded).

I think it might be better to update the issue showing that (intentionally excluded).

src/scenarios/authorization-server/dpop.ts
```diff
@@ -0,0 +1,727 @@
...
+    const redirectUri = `${REDIRECT_URI_ORIGIN}:${options.port}${REDIRECT_URI_PATH}`;
+
+    const params = new URLSearchParams({
+      response_type: 'code',
+      client_id: options.clientId!,
+      state,
+      redirect_uri: redirectUri,
+      code_challenge: codeChallenge,
+      code_challenge_method: 'S256'
+    });
```
tnorimat, line 518: Recently, by the merged PR https://github.com/modelcontextprotocol/conformance/pull/466 , `resource` parameter is supported in authorization-code-grant.ts:171-172 .

I think it is good for dpop.ts to support it to be align with authorization-code-grant.ts.

tnorimat (changes requested):
@PieterKas Hello, I reviewed the PR and added some comment. I would appreciate it if you could check them.

src/seps/sep-1932.yaml
PieterKas: Agreed. I will add that.

src/scenarios/authorization-server/dpop.ts
PieterKas: Agreed. I will add this.

src/scenarios/authorization-server/dpop.ts
PieterKas: You're right it's listed In Scope on #370 and this PR doesn't cover it. 

Based on my understanding the shared test AS (createAuthServer) doesn't support refresh tokens today. It advertises refresh_token in grant_types_supported, but never issues a refresh_token and has no grant_type=refresh_token handler. 

Without that we can't build the conformant-vs-misbehaving pair needed to validate a refresh-binding check (an AS that re-binds vs one that doesn't), so the check can't be implemented and validated under the suite's 'prove it passes and fails' rule.

I'll leave refresh-token binding out of scope for this PR, record it as an excluded: row in sep-1932.yaml with this rationale, and note the descope on #370. Once the test AS gains refresh support, it can be added in the future.

src/seps/sep-1932.yaml
PieterKas: Agreed. I will update the issue to show it is intentionally excluded.

src/scenarios/authorization-server/dpop.ts
PieterKas: Going with out-of-scope for now. 

A DPoP binding on an opaque (reference) token isn't observable on the wire (the cnf.jkt lives inside the token), so verifying it requires token introspection (RFC 9449 §6.2), which is optional in OAuth. In addition we couldn't validate such a check in the suite since the shared test AS (createAuthServer) has no introspection endpoint, so there's no way to build the conformant-vs-misbehaving pair the "prove it passes and fails" rule needs. 

No other AS scenario introspects or inspects token structure, so the scenario reports the binding check as SKIPPED for opaque tokens (documented harness gap) rather than mis-scoring it. This is consistent with the rest of the suite. I'll note this out-of-scope on #370. This can be revisited if the suite gains introspection support.

src/seps/sep-1932.yaml
tnorimat: I confirmed the added comment.

src/scenarios/authorization-server/dpop.ts
tnorimat: I confirmed the added comment. Thank you.

src/scenarios/authorization-server/dpop.ts
tnorimat: I confirmed the added codes. Thank you.

src/seps/sep-1932.yaml
tnorimat: I confirmed the updated issue (#370). Thank you.

src/scenarios/authorization-server/dpop.ts
tnorimat: I confirmed the updated issue (#370). Thank you.

tnorimat (approved):
LGTM

tnorimat:
@pcarleton 
I confirmed that the PR fixed the point I had commented.
I ran the tests of the PR against keycloak supporting DPoP and confirmed that the tests work well (for both public and confidential client).
There seems to be room to consolidate and reduce duplicate code in the existing tests for AS. I would like to handle it as a follow-up PR.

I would appreciate it if you could check the PR. If it is OK, I would like to merge the PR. Many thanks.

PR #420, opened 2026-07-27
feat(cli): --extensions/--exclude-extensions scenario selection

pcarleton:
## Problem

Extension scenarios (DPoP, WIF, client-credentials, EMA, tasks) are optional by definition, but `--suite all` selects them unconditionally. SDKs that don't implement an extension end up baselining its scenarios in `expected-failures.yml` — a file whose stated semantics ("burns down per milestone", "MUST be removed as they're fixed") misrepresent optional surface as conformance debt. Both typescript-sdk and python-sdk currently carry DPoP/WIF and tasks entries for exactly this reason.

## Change

Explicit, flexible runner config for which extensions run:

- `--extensions <ids>` — keep only these extensions' scenarios in the selected suite (`none` deselects them all). Asking for an extension that contributes no scenarios to the suite is an **error**, so a wrong suite or a `--spec-version` that already filtered it out can't silently pass as tested.
- `--exclude-extensions <ids>` — run everything but these (`none` deselects nothing; excluding an absent extension is a satisfied no-op). Mutually exclusive with `--extensions` (commander `conflicts`).

Semantics:

- IDs comma-separated, full or short form (`auth/dpop`, `tasks`, `io.modelcontextprotocol/tasks`).
- Spec-timeline scenarios always pass through; an explicit `--scenario` runs unfiltered (a note is printed when combined with the flags).
- Deselected scenarios are logged, so CI output shows exactly what was opted out — no silent exclusions.
- Empty values error (`--extensions "$UNSET_VAR"` must not silently change which scenarios run); a selection that deselects every scenario in a non-empty suite fails loudly.
- `list` accepts both flags as a dry-run preview of a selection; `sdk` passes them through to the child client/server invocations (including empty values, so the child's guard fires, and with the same immediate conflict check).
- `authorization` is untouched — no extension-tagged AS scenarios exist, so the flags would be a guaranteed error there.

## Verification

- typescript-sdk v2: `client --suite all --exclude-extensions auth/dpop,auth/wif` → **439 passed, 0 failed with no expected-failures file** (today this requires baselining 3 scenarios).
- python-sdk: `server --suite all --extensions none` → **137 passed, 0 failed with no baseline** (today: 9 baselined tasks scenarios).
- ~20 adversarial CLI probes (empty values, `none` on both flags, conflicts, unknown IDs, include-with-no-match, spec-version interactions, suite emptied by selection, `--scenario` combinations).
- 10 new unit tests in `src/scenarios/extension-filter.test.ts`; full suite 459/459; typecheck + lint clean.

Follow-up once released: both SDKs can swap their DPoP/WIF/tasks baseline entries for `--exclude-extensions` legs (the stale-entry enforcement forces the cleanup in the same pin-bump PR).

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,7 @@ npx @modelcontextprotocol/conformance client --command "<client-command>" --scen
 - `--suite` - Run a suite of tests in parallel: `all`, `core`, `extensions`, `backcompat`, `auth`, `metadata`, `draft` (scenarios targeting the in-progress draft spec), or `sep-835`
 - `--spec-version <version>` - Filter scenarios by spec version (e.g., `2025-11-25`, `2026-07-28`; `draft` is accepted as an alias for the current draft identifier). The draft version selects the latest dated release plus any draft-only scenarios. When omitted, the version is inferred from the scenario's spec applicability (draft-only scenarios run at the draft version, everything else at the latest dated release); an explicitly requested version outside a scenario's applicability window skips the scenario (exit 0) unless `--force` is passed
 - `--force` - Run a scenario even if it is not applicable at the requested `--spec-version`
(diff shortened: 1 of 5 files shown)


pcarleton:
Closing as part of a maintainer backlog sweep: superseded in practice by --requirements (#447), please reopen or comment if this is still active and I'll take another look.

PR #432, opened 2026-08-01
feat: add swift-sdk to the built-in SDK matrix

shoemoney:
Register `swift-sdk` in the built-in SDK matrix using its existing `mcp-everything-client` and `mcp-everything-server` SwiftPM products. Build both fixtures, use `.build/debug` paths and start the server with `--port 3000` to match the runner URL. Change the unknown-SDK test fixture to `kotlin-sdk`.

The September 17 rebase onto `7169291ec0b6` preserves both the upstream Ruby registration test and this PR's Swift test. The Swift configuration is unchanged from the original patch.

Current validation: `npm run check`, `npm run build` and all 623 tests pass.

Live Swift fixture conformance was not rerun for this mechanical rebase. The original submission reported 216 core client passes and 67 server passes with two server failures; it also reported 22 failures in the wider client suite that were absent from the SDK baseline. Those are historical results, not current pass counts. No expected failures were added here to hide them. The related fixture fix remains modelcontextprotocol/swift-sdk#269.

AI assistance: this maintenance update, conflict resolution and verification were performed with Codex. The commands above were executed locally; no human review of this update is claimed.

### DIFF
--- a/src/sdk-runner/known-sdks.ts
+++ b/src/sdk-runner/known-sdks.ts
@@ -127,6 +127,24 @@ export const KNOWN_SDKS: Record<string, SdkConfig> = {
       }
     }
   },
+  // Fixtures live in Sources/MCPConformance/{Client,Server} — the
+  // mcp-everything-client and mcp-everything-server products, mirroring
+  // scripts/run-conformance.sh. SwiftPM writes binaries to a per-triple
+  // directory but also maintains a `.build/debug` symlink pointing at it, so
+  // the paths below stay portable across host architectures. The server
+  // defaults to port 3001; `--port` pins it to the 3000 convention used above.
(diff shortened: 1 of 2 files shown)


kahunamoore:
@shoemoney I started working on this before I saw your PR. I'm doing this as part of a larger PR for the swift-sdk to bring it up to the latest MCP standard. I'm happy to compare notes, help out with your version, combine the best of both, whatever makes sense to you.

The one thing I know is different, and unrelated to swift-sdk, is that I'm adding a new conformance script that tests every known sdk against each other in all combinations of --mode and --requirements to see what works, what kinda works and what is broken. I'm done with the former (adding swift-sdk to the test matrix) but still working on the script.

Issue #440, opened 2026-08-04
input-required-result: harness sends the literal string `"undefined"` as an inputResponses key when round 1 names no input requests


marcus-kepler-92:
Following up on #439 item 3 — I went to look at that one guard and found a
round-2 consequence that issue doesn't cover.

`InputRequiredResult.inputRequests` is optional (`required: ["resultType"]` in the
draft schema), and `sep-2322.yaml` states the rule as "at least one of
`inputRequests` **or** `requestState`". So a server answering

```json
{ "resultType": "input_required", "requestState": "s-1", "inputRequests": {} }
```

is conformant, and the round-1 checks are right to pass it. But five round-2 sites
then take the first key of that map without checking it exists:

```ts
const inputKey = Object.keys(r1Result.inputRequests!)[0];   // undefined
// ...
inputResponses: { [inputKey]: mockElicitResponse({ ok: true }) }
```

A computed key stringifies, so the harness puts the literal string `"undefined"` on
the wire.

## Repro

Against a probe server returning the result above for every MRTR tool and
`prompts/get`, with `requestState` rotating so round 2 is reachable:

```
node dist/index.js server --url http://localhost:PORT/mcp \
  --scenario input-required-result-request-state --spec-version 2026-07-28
```

The probe logged the `inputResponses` keys it received across four scenarios:

```
tools/call  test_input_required_result_request_state   -> ["undefined"]
tools/call  test_input_required_result_multi_round     -> ["undefined"]
tools/call  test_input_required_result_multi_round     -> ["undefined"]
prompts/get test_input_required_result_prompt          -> ["undefined"]
tools/call  test_input_required_result_tampered_state  -> ["undefined"]
```

Sites: [`:539`], [`:865`], [`:908`], [`:1095`], [`:1337`]. The two equivalents at
[`:269`] and [`:405`] are unreachable here — their round-1 arms use
`Object.keys(...)[0]` behind a `!key` guard that already failed the scenario.

## Why it matters

The round-2 check then fails, and the report attributes that to the server:

```
[sep-2322-request-state-incomplete] SUCCESS Server returns InputRequiredResult with both inputRequests and requestState
[sep-2322-request-state-complete  ] FAILURE Server validates echoed requestState and returns complete result
[wire-schema-valid                ] SUCCESS Every JSON-RPC message the implementation sent is valid ...
```

`wire-schema-valid` passes because `inputResponses` keys are unconstrained, so
nothing else catches it either.

## Suggested direction

[...]

Issue #439, opened 2026-08-03
input-required-result: sampling/roots scenarios vs SEP-2577, and capability-check accepting no inputRequests


paulbalandan:
Referee `0.2.0-alpha.10`, server mode, `--spec-version 2026-07-28 --suite all`.

Three scenarios in the SEP-2322 set that a server following SEP-2577 cannot pass, plus two separate observations about `input-required-result-capability-check`, which is the fourth. The first item may well be a deliberate call about what the suite is scoring, so that part is a question rather than a bug report. The last one looks like a plain defect.

### 1. Three scenarios require the server to emit a deprecated request

`input-required-result-basic-sampling`, `input-required-result-basic-list-roots` and `input-required-result-multiple-input-requests` pass only if the server under test puts a `sampling/createMessage` or `roots/list` entry in `inputRequests`.

In `schema/draft/schema.ts`, `CreateMessageRequest`, `CreateMessageResult`, `SamplingMessage`, `ListRootsRequest`, `ListRootsResult`, `Root` and the `ClientCapabilities.sampling` / `.roots` slots all carry `@deprecated Deprecated as of protocol version 2026-07-28 (SEP-2577)`. `ElicitRequest` is the one member of the `InputRequest` union left unmarked. SEP-2577 says new implementations "should not add support for them", and the changelog's Deprecated section repeats it.

A server that takes that guidance is then required by the suite to fail. The capabilities side compounds it: `ClientCapabilities.sampling` and `.roots` carry the same deprecation, so a client written to this revision declares neither, and the ability these three scenarios test has no conformant counterparty to exercise it against.

This may be intentional. An argument for the current behaviour: the suite scores the spec surface as written, deprecated members included, and a deprecated feature is still a valid feature for the twelve months SEP-2596 guarantees. If that is the intent then the failures are correct and we will keep them baselined on our side.

If it is not, the suite already has the shape this would want. The SEP-2575 client scenario reads `_meta['io.modelcontextprotocol/clientCapabilities']` and records `SKIPPED` for an absent capability, and `json-schema-2020-12` gates on spec version the same way. Either would let these three skip rather than fail for a server that omits the deprecated half of the union.

### 2. `input-required-result-capability-check` rejects `-32021`

The scenario sends `_meta['io.modelcontextprotocol/clientCapabilities'] = {sampling: {}}` and expects `inputRequests` naming sampling only.

[...]

PR #442, opened 2026-08-04
fix(input-required-result): report the capability check as untestable when nothing is requested

marcus-kepler-92:
Narrows #439 item 3. The round-2 consequence of the same case is #440; this PR
does not touch it.

## The bug

`input-required-result-capability-check` guarded its elicitation scan with:

```ts
} else if (isInputRequiredResult(result) && result.inputRequests) {
```

With `inputRequests` absent, that arm is skipped and the following
`isCompleteResult` arm is false by construction for an `input_required` result, so
nothing is recorded and the check passes. `inputRequests: {}` passes by a different
path — the arm *is* taken (`{}` is truthy), the loop just has nothing to scan.

Either way the check reports SUCCESS without having exercised the requirement it
scores.

## Why `notTestable()` rather than a plain assertion

`inputRequests` is optional — a result carrying only `requestState` still satisfies
"at least one of `inputRequests` or `requestState`" — so a server that names none
has not violated `sep-2322-respect-client-capabilities`
(_"Servers MUST NOT send an inputRequests that the client has not declared support
for in its capabilities"_). It has not exercised it either.

That is the #248 / #372 case, so the check now reports through `notTestable()` with
`details.untestable`, per AGENTS.md and the existing call sites in `stateless.ts`.
Status stays FAILURE — the severity follows the underlying MUST NOT — but the
message says the requirement could not be verified rather than claiming a violation
the server did not commit.

## Does this fail a compliant implementation?

No. The scenario's description is an explicit fixture contract: implement
`test_input_required_result_capabilities`, read
`_meta['io.modelcontextprotocol/clientCapabilities']`, and return inputRequests for
the declared capabilities. The probe declares `sampling: {}`, so a server following
that contract returns a `sampling/createMessage` request and still scores SUCCESS
(verified below).

What now goes red is a server that implements the tool but returns no input
request — i.e. the fixture prerequisite is missing, which is exactly what #248
says should be red, with the expected-failures baseline as the escape hatch. It is
still a behaviour change for anyone currently green on that basis.

## Fixture

Adds the case to `sep-2322-mrtr-broken-server`. The fixture returns a *conformant*
envelope — `requestState` present, no input requests — rather than a bare
`{"resultType":"input_required"}`, so it isolates this check instead of also
tripping `sep-2322-request-state-incomplete`.

## Verified

Against `--spec-version 2026-07-28`:

```
everything-server              Passed: 2/2, 0 failed, 0 warnings

sep-2322-mrtr-broken-server    FAILURE  RespectClientCapabilities
                               Not testable: server returned no inputRequests,
                               so the capability restriction was never exercised
```

The negative test pins the slug, the `Not testable:` prefix and
`details.untestable` rather than just asserting a failure count.

`npm run lint`, `npm run typecheck` and the full `vitest run` (43 files / 505
tests) pass.

### DIFF
--- a/examples/servers/typescript/sep-2322-mrtr-broken-server.ts
+++ b/examples/servers/typescript/sep-2322-mrtr-broken-server.ts
@@ -8,6 +8,11 @@
  * 2. Returns InputRequiredResult on `tools/list` (unsupported method)
  * 3. Accepts tampered requestState without integrity verification
  *
+ * Plus one case that is not a violation at all:
+ *
+ * 4. Answers with a conformant input_required result that names no input
+ *    request, leaving the capability check with nothing to verify.
+ *
  * The conformance scenarios should emit FAILURE against this server.
  */
 
@@ -54,6 +59,11 @@ handlers['tools/list'] = () => ({
       name: 'test_input_required_result_tampered_state',
       description: 'Test tool for tampered state',
       inputSchema: { type: 'object' as const, properties: {} }
+    },
+    {
+      name: 'test_input_required_result_capabilities',
+      description: 'Test tool for client capability handling',
+      inputSchema: { type: 'object' as const, properties: {} }
     }
   ]
 });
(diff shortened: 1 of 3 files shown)


pcarleton:
thank you!

pcarleton (approved):

PR #335, opened 2026-06-09
feat: add client-side JSON Schema 2020-12 preservation scenario

pugsatoshi:
This PR adds a new JSON-Schema-2020-12-preservation client conformance scenario that closes Issue #101. The existing `json-schema-2020-12` scenario tests **server-side** keyword emission; this new scenario tests the **client-side** preservation path that SEP-1613 and TypeScript SDK PR #1135 actually fixed (`.catchall()` so the parser retains the 2020-12 vocabulary).

The scenario sets up a mock MCP server (via the version-aware `MockServer` abstraction from #321) that advertises:
- `json_schema_2020_12_tool` — focal tool whose `inputSchema` contains the full SEP-1613 + SEP-2106 vocabulary
- `json_schema_echo` — permissive echo tool

The client under test calls `tools/list`, then round-trips the observed `inputSchema` back via `tools/call json_schema_echo` with `{ schema: <observed> }`.
The scenario diffs the echoed schema against the canonical fixture and emits eight per-keyword checks. SEP-2106 vocabulary checks are soft-gated via `sep2106KeywordCheckStatus` so they only fail on `DRAFT-2026-v1` runs.

## Motivation and Context

Closes #101.

Issue #101 noted that the existing `json-schema-2020-12` scenario covers only the server's wire output, while the actual SEP-1613 / TS SDK PR #1135 fix was on the client side.
This PR fills that gap with a dedicated client-side scenario that mirrors the server-side scenario, one check at a time.

Scope is intentionally pinned to `introducedIn: '2025-11-25'`.
The 2026-07-28-RC changelog ([draft changelog](https://modelcontextprotocol.io/specification/draft/changelog)) does not change the preservation requirements for SEP-1613 / SEP-2106, so the wire shape under test remains unchanged.
The broader 2026-07-28-RC changes (SEP-2575 stateless, SEP-2549 `CacheableResult`, SEP-2243 `Mcp-Method` MUST) are deliberately out of scope for this PR; once 2026-07-28 is finalized and lands in `DATED_SPEC_VERSIONS`, the same scenario can be re-targeted in a follow-up.

The focal-schema fixture (a 2020-12 `inputSchema` with `$schema`, `$defs.address.$anchor`, `allOf`/`anyOf`, `if`/`then`/`else`, `additionalProperties: false`) was previously embedded only in the server-side scenario's `description` markdown block.
[...]

PR #443, opened 2026-08-05
Freeze conformance requirements per spec revision, run each at its wire

felixweinberger:
Adds `requirements/<revision>.yaml`, a frozen list of the scenarios a spec revision
requires, and a `--requirements` flag that runs exactly that set **at that
revision's wire version**. Ships two sets, 2025-11-25 and 2026-07-28; `tier-check`
takes both at once and Tier 1 means every listed revision passes.

## Motivation and Context

Today there is no way to answer "which scenarios must my implementation pass to
conform to the spec released on 2026-07-28". The suite accumulates scenarios
continuously, so the set an implementation is measured against keeps growing
after the revision has shipped.

typescript-sdk is the live case. `json-schema-2020-12-preservation` merged on
2026-07-31 (#335), three days after the 2026-07-28 spec shipped and four days
after `0.2.0-alpha.10` was published. typescript-sdk pins alpha.10, so it has
never been able to run that scenario. On the same commit it passes 100% measured
against alpha.10 and less than 100% measured against `main`.

That matters because SEP-1730 puts Tier 1 at exactly 100% and relegates Tier 1 to
Tier 2 when any conformance test fails for four weeks. A scenario added after a
release starts a demotion clock against an implementation that did nothing wrong
and changed nothing.

`--spec-version` does not solve this. It filters on which revision a scenario
targets, not on when the scenario became a requirement, so a post-release
addition tagged `[2025-11-25, 2026-07-28]` is indistinguishable from one that
existed at ship.

## What this adds

```bash
# what does conforming to 2026-07-28 require?
npx @modelcontextprotocol/conformance list --requirements 2026-07-28

# run exactly that
npx @modelcontextprotocol/conformance server --url http://localhost:3000/mcp --requirements 2026-07-28
```

- `--requirements <revision>` on the `client`, `server`, `list` and `tier-check`
  commands. It replaces `--suite` and `--spec-version`, and passing it with
  either is an error rather than a silent precedence rule.
- A requirement set covers the two roles the specification defines: an MCP server
  acting as an OAuth resource server, and an MCP client acting as an OAuth
  client. It covers no authorization-server scenarios, because the spec puts
  authorization server implementation
  [beyond its own scope](https://modelcontextprotocol.io/specification/2026-07-28/basic/authorization)
  ("The implementation details of the authorization server are beyond the scope
  of this specification"), so those scenarios serve people deploying an
  authorization server rather than implementations of MCP itself. A set naming an
  `authorization` section is rejected with that reason, since the scenarios carry
  a dated tag for every revision and so read as core spec at a glance.
- Passing nothing keeps today's behaviour and runs everything, so the suite stays
  the place completeness lives.
- Scenarios a revision runs without scoring go under `not_scored`, each carrying
  the leg that runs it and why it does not count:

  | Reason | Meaning |
  | --- | --- |
  | `extension` | Optional by definition. SEP-1730: "Experimental features and protocol extensions (such as Tasks and MCP Apps) are not required for any tier." |
  | `added-after-release` | The scenario did not exist when the revision shipped, so no implementation could have been passing it. |

  Both still run and both appear in the report with their reason, so a failing
  extension stays visible without moving a pass rate. Promoting an entry into the
  required lists stays a deliberate, reviewable change, which is how the suite
  grows without retroactively failing anyone.
- `tier-check` reports `requirements_revision` and scores both legs against the
  set, replacing the spec-version matrix with a plain required/passed summary.

The `2026-07-28` set is generated from `0.2.0-alpha.10`, published the day before
that revision shipped, minus the scenarios pending in that release (SEP-1730
scores "applicable required tests" only). Older revisions are deliberately not
manifested: reconstructing them means archaeology on version-bump commits for
much less value, and with no flag nothing changes.

A requirement set is this project's contract, which is why it lives here and why
`--requirements` takes a revision rather than a path. It is the opposite of an
[expected-failures](README.md#expected-failures) baseline, which lives in an
implementation's own repository and records what that implementation knows it
fails. A baselined failure is still a failure against a requirement set, so the
two are independent inputs and both can be passed in the same run.

## How Has This Been Tested?

Driven end to end against the typescript-sdk everything server at
`main` (`cc4b41617`):

- `tier-check --requirements 2026-07-28` reports Server 37/37 and Client 32/32,
  both 100%, alongside `Not scored (17 run, 14 failing)`, and conformance drops
  off the Tier 1 blocker list, leaving only `triage` and `p0_resolution`. Without
  the flag the same commit scores Client: Core 4/5, failing on the post-release
  scenario.
- Against a deliberately broken fixture server the same command scores 1/37 and
  `server_conformance` does appear in the blockers, with the 10 extension
  failures still held apart. The split suppresses the right failures and only
  those.
- `server --requirements 2026-07-28` runs 37 scenarios, 120 checks, 0 failed.
- `list --requirements 2026-07-28` prints the 69-scenario set plus what is run
  without being scored.
- Conflicting and malformed input: combined with `--spec-version` or `--suite`,
  an unknown revision, a path-shaped revision, and an empty value each exit 1
  with a message naming the problem.
- Unit tests assert every name in every requirement set resolves to a scenario
  this build provides, so a rename or removal cannot silently shrink what
  conformance means.

## Breaking Changes

None. The flag is opt-in and the default path is unchanged.

## Types of changes

## Checklist

## Additional context

Two things deliberately left out, happy to add either:

- **No CI check that a shipped revision's lists never change.** The freeze is
  currently a comment in the file plus review. A test pinning each shipped file's
  hash would enforce it, at the cost of a deliberate two-step when a set does
  need amending.
- **No change to the `2026-07-28` draft classification.** `DRAFT_PROTOCOL_VERSION`
  is still `2026-07-28` even though the spec repo has published
  `schema/2026-07-28`. Promoting it shifts the `active` suite for every SDK, so it
  belongs in its own change. A requirement set does not depend on it either way.

### DIFF
--- a/.claude/skills/mcp-sdk-tier-audit/README.md
+++ b/.claude/skills/mcp-sdk-tier-audit/README.md
@@ -95,10 +95,12 @@ The skill lives in `.claude/skills/` in this repo, so if you open [Claude Code](
 3. Run the skill:
 
 ```
-/mcp-sdk-tier-audit <local-sdk-path> <conformance-server-url> [client-cmd]
+/mcp-sdk-tier-audit <local-sdk-path> <conformance-server-url> [client-cmd] [--requirements <revision>]
 ```
 
 Pass the client command as the third argument to include client conformance testing. If omitted, client conformance is skipped and noted as a gap in the report.
+
+**Pass `--requirements` with every revision the SDK claims**, comma-separated. Each revision's scenarios run at that revision's own wire version, and all of them must pass for Tier 1: the dated revisions through `2025-11-25` use the stateful initialize handshake while `2026-07-28` is stateless, so a scenario belonging to both has to work on both and one run does not cover the other. It also means a scenario added to the suite after a revision shipped cannot fail an SDK that had no opportunity to adopt it. Without the flag, scoring uses the suite as it stands today, which is not a tier claim. See [Conformance Requirements](../../../README.md#conformance-requirements), and run `conformance list --requirements 2025-11-25,2026-07-28` to see both sets.
 
 **TypeScript SDK example:**
 
@@ -107,7 +109,7 @@ Pass the client command as the third argument to include client conformance test
 cd ~/src/mcp/typescript-sdk && npm run test:conformance:server:run
 
 # Terminal 2: run the audit (from the conformance repo)
-/mcp-sdk-tier-audit ~/src/mcp/typescript-sdk http://localhost:3000/mcp "npx tsx ~/src/mcp/typescript-sdk/test/conformance/src/everythingClient.ts"
+/mcp-sdk-tier-audit <sdk-path> http://localhost:3000/mcp "node --import tsx <sdk-path>/test/conformance/src/everythingClient.ts" --requirements 2025-11-25,2026-07-28
 ```
 
 **Python SDK example:**
--- a/.claude/skills/mcp-sdk-tier-audit/SKILL.md
+++ b/.claude/skills/mcp-sdk-tier-audit/SKILL.md
@@ -5,7 +5,7 @@ description: >-
   Produces tier classification (1/2/3) with evidence table, gap list, and
   remediation guide. Works for any official MCP SDK (TypeScript, Python, Go,
   C#, Java, Kotlin, PHP, Swift, Rust, Ruby).
-argument-hint: '<local-path> <conformance-server-url> [client-cmd] [--branch <branch>]'
+argument-hint: '<local-path> <conformance-server-url> [client-cmd] [--requirements <revision>] [--branch <branch>]'
 ---
 
 # MCP SDK Tier Audit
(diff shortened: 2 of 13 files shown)


felixweinberger:
Closing this in favour of a version that builds on `src/sdk-runner/known-sdks.ts`.

Running it against python-sdk, rust-sdk, go-sdk and csharp-sdk turned up a design problem. `tier-check` takes a single `--conformance-server-url` for every revision, which is only correct for typescript-sdk, whose everything server routes each request and serves both wire eras from one endpoint. go-sdk needs a different server command per revision (its transport only supports >= 2026-07-28 when constructed stateless) and csharp-sdk serves the two lifecycles at different URLs, so both scored ~50% when the truth is 100%.

`known-sdks.ts` and `config.ts` already model exactly this, per SDK and per spec version, down to a comment noting that an override may change just the command (go-sdk) or just the url (csharp-sdk). This PR reinvented a narrower thing beside it. Redoing it so the requirement sets drive the SDK through that existing config.

PR #444, opened 2026-08-05
fix(sep-2243): reject x-mcp-header on number-typed params

### LINKED ISSUE #344: http-custom-headers scenario puts x-mcp-header on `type:"number"` params (forbidden by SEP-2243)
## Summary

The client scenario `http-custom-headers` (`src/scenarios/client/http-custom-headers.ts`) declares two tool parameters with an `x-mcp-header` annotation on a JSON Schema `type: "number"` property, and then asserts the client mirrors them into `Mcp-Param-*` headers:

```ts
priority:  { type: 'number', /* ... */ 'x-mcp-header': 'Priority' },   // ~line 296
float_val: { type: 'number', /* ... */ 'x-mcp-header': 'FloatVal' },   // ~line 322
// ...
this.checkParamHeader(req, 'Priority', args.priority, 'number');       // ~line 453
this.checkParamHeader(req, 'FloatVal', args.float_val, 'number');      // ~line 480
```

This violates SEP-2243, which the draft spec now states unambiguously (clarified in modelcontextprotocol/modelcontextprotocol#2863):

> The `x-mcp-header` extension ... **MUST** only be applied to parameters with primitive types (integer, string, boolean). Parameters with type `number` are **not permitted**.
> — `docs/specification/draft/server/tools.mdx`

## Why this matters

A spec-conformant client **must reject (exclude) tools whose `x-mcp-header` annotations are invalid** — this is exactly what the sibling scenario `http-invalid-tool-headers` verifies (it includes an "x-mcp-header on a primitive-only constraint" negative case). Because `test_custom_headers` carries `x-mcp-header` on `type:"number"` params, a conformant client excludes the whole tool from its listing, so the subsequent `tools/call` sends **no** `Mcp-Param-*` headers. Every positive "client mirrors `Mcp-Param-X`" check then fails, while the negative checks pass. The scenario therefore cannot be passed by a spec-conformant client.

This affects `main` and the latest published `@modelcontextprotocol/conformance@0.2.0-alpha.4`.

## Suggested fix

- Change `priority` and `float_val` to `type: "integer"` (the `priority` value is already integer-valued, and the check comment even says "integer number"), **or** drop the `x-mcp-header` annotation from the `number`-typed param, so the positive scenario only exercises spec-permitted primitive types (integer/string/boolean).
- Optionally add a `type:"number"` row to `http-invalid-tool-headers` asserting clients reject `x-mcp-header` on `number`, and refresh any `sep-2243.yaml` text that still implies `number` is allowed.

## Context

Filed from the C# SDK side while validating SEP-2243 conformance for modelcontextprotocol/csharp-sdk#1610. The C# client already emits `Mcp-Param-*` headers correctly for permitted primitive types and intentionally rejects `x-mcp-header` on `number` per the spec, so this one scenario stays red purely because of the harness declaration. Tracked on our side in modelcontextprotocol/csharp-sdk#1655.

halter73:
Closes #344

## Why this is still open after #371

SEP-2243 permits `x-mcp-header` only on **integer, string and boolean** parameters. The released `2026-07-28` spec is explicit:

> `x-mcp-header` **MUST** only be applied to parameters with primitive types (integer, string, boolean). Parameters with type `number` are not permitted.
> — [`server/tools.mdx`](https://github.com/modelcontextprotocol/modelcontextprotocol/blob/main/docs/specification/2026-07-28/server/tools.mdx)

and the value-encoding table in `basic/transports/streamable-http.mdx` lists only `string` / `integer` / `boolean`.

#371 fixed the **positive** fixture (`priority` is now `integer`, `float_val` is served unannotated and asserted *not* mirrored) but left the two items #344 listed as optional, which is why the issue stayed open. This finishes them.

## Changes

**1. Negative case for the `number` rule.** `HttpInvalidToolHeadersScenario` served invalid tools for every other `x-mcp-header` constraint — empty / object / array / null / duplicate / space / colon / non-ascii / control-char — but had no number-typed one. `sep-2243-x-mcp-header-primitive-only` was only ever exercised via object/array/null, so nothing verified the rule the issue was actually about. Adds an `invalid_number_header` tool with `score: { type: 'number', 'x-mcp-header': 'Score' }`, mapped to the existing requirement.

No check IDs added or removed — `INVALID_TOOL_DECLARED_CHECK_IDS` is unchanged.

**2. Requirement text corrected.** `sep-2243.yaml` still described the permitted set as "(number, string, or boolean)" and said "number values MUST be converted to their decimal string representation" — both contradict the released spec. `traceability.json` embeds this text verbatim, so its two `text` fields are updated in place; check IDs and `status` are untouched.

## Verification

- `npm run typecheck`, `npm run lint`, `npm run build` clean; `npm test` 505 passed / 43 files.
- **Mutation-tested**: removing the `INVALID_TOOL_CONSTRAINT_IDS` mapping makes the new test fail (`expected [ 'SUCCESS', 'SUCCESS', 'SUCCESS' ] to include 'FAILURE'`), so it isn't vacuous.
- Ran both scenarios against typescript-sdk@main: positive `http-custom-headers` still 18/18.

## The new check finds a real SDK bug

`http-invalid-tool-headers` against typescript-sdk@main goes 11/12 — `sep-2243-x-mcp-header-primitive-only` fails on `invalid_number_header`. The cause is in the SDK, not the harness:

```ts
// packages/core-internal/src/shared/mcpParamHeaders.ts
const PERMITTED_X_MCP_HEADER_TYPES = new Set(['string', 'integer', 'boolean', 'number']);
```

`number` was allow-listed there *specifically* to tolerate this fixture bug, with a comment noting the discrepancy is "tracked upstream." With #371 having fixed the fixture, that workaround can be dropped. Worth a follow-up on the SDK side.

This won't red-light CI: `ci.yml` runs only `npm ci` / `check` / `build` / `test`. The typescript-sdk run lives in `traceability.yml`, which is manual/scheduled, uses `|| true`, and per AGENTS.md is not a PR gate.

## Out of scope

#354 (positive scenario's `ttlMs: 0` vs. SEP-2549 stale-schema) is a separate defect owned by #358. The positive scenario's TTL is untouched here, and #358's scope note confirms it doesn't modify the invalid-tool scenario — so these shouldn't collide beyond a trivial rebase.

### DIFF
--- a/src/scenarios/client/http-custom-headers.test.ts
+++ b/src/scenarios/client/http-custom-headers.test.ts
@@ -30,6 +30,18 @@ async function post(
   });
 }
 
+async function postJson(serverUrl: string, body: object): Promise<any> {
+  const res = await fetch(serverUrl, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+      Accept: 'application/json, text/event-stream'
+    },
+    body: JSON.stringify(body)
+  });
+  return res.json();
+}
+
 function idsOf(checks: { id: string }[]): Set<string> {
   return new Set(checks.map((c) => c.id));
 }
@@ -193,6 +205,44 @@ describe('HttpInvalidToolHeadersScenario (SEP-2243) check IDs', () => {
       // The other constraints were not violated.
       expect(
         statusesFor(checks, 'sep-2243-x-mcp-header-charset')
+      ).not.toContain('FAILURE');
+    } finally {
+      await scenario.stop();
+    }
+  });
+
+  it('FAILs primitive-only when the client calls the number-typed tool', async () => {
+    const scenario = new HttpInvalidToolHeadersScenario();
+    const { serverUrl } = await scenario.start(testScenarioContext());
+    try {
+      const listed = await postJson(serverUrl, {
+        jsonrpc: '2.0',
+        id: 1,
+        method: 'tools/list'
+      });
+      // SEP-2243 permits x-mcp-header only on integer/string/boolean, so the
+      // number-typed tool must be served for the client to reject it.
+      const numberTool = listed.result.tools.find(
+        (t: { name: string }) => t.name === 'invalid_number_header'
+      );
+      expect(numberTool?.inputSchema.properties.score).toEqual({
+        type: 'number',
+        'x-mcp-header': 'Score'
+      });
+
+      await post(serverUrl, {
+        jsonrpc: '2.0',
+        id: 2,
+        method: 'tools/call',
+        params: { name: 'invalid_number_header', arguments: { score: 1.5 } }
+      });
+      const checks = scenario.getChecks();
+      expect(
+        statusesFor(checks, 'sep-2243-x-mcp-header-primitive-only')
+      ).toContain('FAILURE');
+      // The other constraints were not violated.
+      expect(
+        statusesFor(checks, 'sep-2243-x-mcp-header-not-empty')
       ).not.toContain('FAILURE');
     } finally {
       await scenario.stop();
--- a/src/scenarios/client/http-custom-headers.ts
+++ b/src/scenarios/client/http-custom-headers.ts
@@ -68,6 +68,7 @@ const INVALID_TOOL_CONSTRAINT_IDS: Record<string, string> = {
   invalid_object_header: 'sep-2243-x-mcp-header-primitive-only',
   invalid_array_header: 'sep-2243-x-mcp-header-primitive-only',
   invalid_null_header: 'sep-2243-x-mcp-header-primitive-only',
+  invalid_number_header: 'sep-2243-x-mcp-header-primitive-only',
   invalid_duplicate_same_case: 'sep-2243-x-mcp-header-unique',
   invalid_duplicate_diff_case: 'sep-2243-x-mcp-header-unique',
   invalid_space_in_name: 'sep-2243-x-mcp-header-charset',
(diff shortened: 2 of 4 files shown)


JosephDoUrden (commented):
Ran the whole thing locally. 505/505 on the PR head, which matches the body. Then merged current main in (post #447) to be sure, still a clean merge, 525/525. I also deleted the invalid_number_header mapping line to check the new test is load-bearing, it fails with exactly the assertion error quoted in the body, so it is.

Both yaml text rewrites check out quote-level against the released 2026-07-28 spec, and the hand-edited traceability block matches a fresh computeTraceability recompute byte for byte. The ts-sdk discrepancy claim is real too, PERMITTED_X_MCP_HEADER_TYPES on main still includes 'number'.

One follow-up worth a note, not blocking and not caused here: the manifest has some pre-existing staleness. There is no 1932 block at all, and two sep-2243 server rows still say -32001 where both the yaml and the released spec say -32020. A regeneration against a reference run would clear both.

Looks merge-ready to me.

pcarleton:
Thanks @halter73, and @JosephDoUrden for the independent verification. Merging: the 2026-07-28 text is explicit that `number` is not a permitted `x-mcp-header` type and that clients MUST exclude such tools, and python/go/csharp/rust already do.

Note for later for the typescript-sdk: `PERMITTED_X_MCP_HEADER_TYPES` in `packages/core-internal/src/shared/mcpParamHeaders.ts` still includes `'number'`, the comment there says it was added only to pass the pre-#371 fixture, so it can simply be dropped now; until then `http-invalid-tool-headers` will report `sep-2243-x-mcp-header-primitive-only` as a failure once the SDK bumps its conformance pin past 0.2.0-alpha.11. I'll open the SDK-side PR. (cc @felixweinberger ) 

https://github.com/modelcontextprotocol/typescript-sdk/pull/2762

pcarleton (approved):

PR #449, opened 2026-08-09
feat: add logging conformance scenarios (capability, invalid level, threshold filtering)

### LINKED ISSUE #435: Add conformance scenario: logging/setLevel and log notifications
The logging utility (logging/setLevel and log message notifications) has no conformance scenario. The spec defines:

- Server MUST accept logging/setLevel with a valid severity level
- Server SHOULD only emit log notifications at or above the configured level
- Valid levels: debug, info, notice, warning, error, critical, alert, emergency

Proposed checks:
- Setting level to "error" stops debug/info/warning notifications
- Setting level to "debug" allows all notifications through
- Invalid level values are rejected appropriately

I'll submit a PR for this.

elang2:
## Summary

Adds three new conformance scenarios for the `logging/setLevel` endpoint (issue #435):

1. **`logging-capability-advertisement`** — verifies the server advertises `capabilities.logging` in its initialize response
2. **`logging-invalid-level`** — verifies the server rejects unrecognized severity levels (e.g. "banana") with `-32602` (Invalid Params)
3. **`logging-threshold-filtering`** — verifies that `notifications/message` respects the configured severity threshold (suppresses below, allows above)

Scenarios 2 and 3 are registered in `pendingClientScenariosList` because the everything-server does not implement invalid-level rejection or threshold-based filtering. They can be run targeted against a compliant fixture server via `npm start -- server --scenario logging-invalid-level --url <fixture>`.

Scenario 1 is in the active list and passes against the everything-server.

## Spec references

- https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging

## Test plan

- [x] `npm run build` passes
- [x] `logging-capability-advertisement` passes against everything-server
- [x] `logging-invalid-level` correctly skipped (pending) in default suite
- [x] `logging-threshold-filtering` correctly skipped (pending) in default suite

Closes #435

### DIFF
--- a/src/scenarios/index.ts
+++ b/src/scenarios/index.ts
@@ -28,6 +28,12 @@ import {
 } from './server/utils';
 
 import {
+  LoggingCapabilityScenario,
+  LoggingInvalidLevelScenario,
+  LoggingThresholdFilteringScenario
+} from './server/logging';
+
+import {
   ToolsListScenario,
   ToolsCallSimpleTextScenario,
   ToolsCallImageScenario,
@@ -123,6 +129,12 @@ import { JsonSchema2020_12PreservationScenario } from './client/json-schema-2020
 
 // Pending client scenarios (not yet fully tested/implemented)
 const pendingClientScenariosList: ClientScenario[] = [
+  // Logging scenarios that require server-side filtering and invalid-level
+  // rejection — the everything-server does not implement these behaviors.
+  // Run targeted: `npm start -- server --scenario logging-invalid-level --url <fixture>`
+  new LoggingInvalidLevelScenario(),
+  new LoggingThresholdFilteringScenario(),
+
   // JSON Schema 2020-12 (SEP-1613)
   // This test is pending until the SDK includes PR #1135 which preserves
   // $schema, $defs, and additionalProperties fields in tool schemas.
@@ -163,6 +175,9 @@ const allClientScenariosList: ClientScenario[] = [
 
   // Utilities scenarios
   new LoggingSetLevelScenario(),
+  new LoggingCapabilityScenario(),
+  new LoggingInvalidLevelScenario(),
+  new LoggingThresholdFilteringScenario(),
   new PingScenario(),
   new CompletionCompleteScenario(),
 
--- a/src/scenarios/server/logging.ts
+++ b/src/scenarios/server/logging.ts
@@ -81,20 +81,60 @@ export class LoggingCapabilityScenario implements ClientScenario {
           details: { logging: caps.logging }
         });
       } else {
-        checks.push({
-          id: 'logging-capability-advertised',
-          name: 'LoggingCapabilityAdvertised',
-          description:
-            'Server advertises logging capability in initialize response',
-          status: 'SKIPPED',
-          timestamp: new Date().toISOString(),
-          errorMessage:
-            'Server did not advertise capabilities.logging. ' +
-            'This is compliant if the server does not emit log notifications. ' +
-            'The spec requires the capability only for servers that emit notifications/message.',
-          specReferences: SPEC_REFS,
-          details: { capabilities: caps }
-        });
+        // Capability absent — probe whether the server actually emits notifications.
+        // If it does, that's a MUST violation; if not, SKIPPED is correct.
+        await conn.close();
+
+        let observedNotifications = 0;
+        try {
+          const probe = await connectToServer(ctx.serverUrl, {}, ctx.specVersion);
+          const collector = new NotificationCollector(probe.client);
+
+          await probe.client.callTool({
+            name: 'test_tool_with_logging',
+            arguments: {}
+          });
+          await new Promise((resolve) => setTimeout(resolve, 200));
+
+          observedNotifications = collector.getLoggingNotifications().length;
+          await probe.close();
+        } catch {
+          // Probe failed (tool missing, etc.) — treat as no evidence
+        }
+
+        if (observedNotifications > 0) {
+          checks.push({
+            id: 'logging-capability-advertised',
+            name: 'LoggingCapabilityAdvertised',
+            description:
+              'Server advertises logging capability in initialize response',
+            status: 'WARNING',
+            timestamp: new Date().toISOString(),
+            errorMessage:
+              `Server emitted ${observedNotifications} log notification(s) without advertising ` +
+              'capabilities.logging. The spec requires servers that emit ' +
+              'notifications/message to advertise the logging capability.',
+            specReferences: SPEC_REFS,
+            details: { capabilities: caps, observedNotifications }
+          });
+        } else {
+          checks.push({
+            id: 'logging-capability-advertised',
+            name: 'LoggingCapabilityAdvertised',
+            description:
+              'Server advertises logging capability in initialize response',
+            status: 'SKIPPED',
+            timestamp: new Date().toISOString(),
+            errorMessage:
+              'Server did not advertise capabilities.logging and no log notifications ' +
+              'were observed. This is compliant — the spec requires the capability ' +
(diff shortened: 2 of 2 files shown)


elang2:
@joy7758 would appreciate your review on this one when you get a chance.

joy7758:
Thanks for the follow-up. I reviewed current head `79acb3987aec6ca0acf32264abbfeba250c0779f`. I am not proposing a competing implementation.

I found four scope/evidence points worth resolving before these checks are treated as conformance results:

1. **Threshold filtering does not currently have RFC-keyword backing.** [`logging.ts:L256-L262`](https://github.com/modelcontextprotocol/conformance/blob/79acb3987aec6ca0acf32264abbfeba250c0779f/src/scenarios/server/logging.ts#L256-L262) labels suppression as `MUST NOT` and emission at `debug` as `SHOULD`. The cited [2025-11-25 logging page](https://modelcontextprotocol.io/specification/2025-11-25/server/utilities/logging) says clients `MAY` send `logging/setLevel`, but I cannot find either receiver requirement there. Unless there is another normative sentence to cite, these should be treated as a non-scoring interoperability observation or an upstream spec gap, not enforced as protocol conformance.

2. **Capability advertisement is conditional.** [`logging.ts:L71-L95`](https://github.com/modelcontextprotocol/conformance/blob/79acb3987aec6ca0acf32264abbfeba250c0779f/src/scenarios/server/logging.ts#L71-L95) warns whenever `capabilities.logging` is absent. The spec requires that capability only for a server that emits log notifications. A server that does not implement logging can therefore be compliant. This check needs either evidence that the server emits/supports logging or a non-scoring outcome when the feature is not claimed.

3. **Zero notifications should not silently become a neutral result after the fixture contract is invoked.** [`logging.ts:L293-L305`](https://github.com/modelcontextprotocol/conformance/blob/79acb3987aec6ca0acf32264abbfeba250c0779f/src/scenarios/server/logging.ts#L293-L305) returns `SKIPPED`. If `test_tool_with_logging` is a controlled prerequisite that promises at least an error-level sentinel, zero notifications means the prerequisite/evidence path failed, so the threshold was not tested. I would use the suite's explicit inconclusive/prerequisite-failure convention rather than `SKIPPED`; if the server is allowed to emit nothing, this behavior is simply unobservable and should not be scored.

4. **The invalid-value probe still needs wire evidence.** Registering it as pending keeps it out of the default suite, but does not by itself show that a targeted run avoids `wire-schema-harness-error`. Please include the targeted `checks.json`/wire-schema output for the intentionally invalid request.

The repository [contribution guide](https://github.com/modelcontextprotocol/conformance/blob/main/CONTRIBUTING.md) also asks for a passing real-SDK run and a specific failing case. The PR currently documents the in-repo everything-server path, so adding the pinned real-SDK output plus a negative fixture/result would complete the evidence loop.

The `-32602` handling itself is correctly scoped as a `SHOULD`. No protocol change is being suggested here.

elang2:
Thanks @joy7758, all four points are valid. I'll push a revision that:

- Downgrades threshold filtering to non-scoring (no normative backing in the spec)
- Makes the capability check conditional on evidence of logging support
- Uses `untestableCheck()` for the zero-notification case
- Adds wire evidence for the invalid-level probe

elang2:
All four points from @joy7758 addressed and pushed. Threshold filtering is non-scoring now, capability check probes for logging support first, zero-notification uses untestableCheck(), and wire evidence added for invalid level.

Maintainer review welcome when someone has time.

joy7758:
Thanks for the revision. I re-reviewed current head `7bd53511b83731093c9e356e27fed685c4bb6c7f`. The zero-notification part is now addressed: both threshold branches use `untestableCheck()` ([error threshold](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/scenarios/server/logging.ts#L338-L350), [debug threshold](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/scenarios/server/logging.ts#L421-L433)). The remaining points are not fully closed yet:

1. **The invalid-level scenario still creates a harness wire-schema failure.** It sends `level: "banana"` through `conn.request()` ([logging.ts:L201-L209](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/scenarios/server/logging.ts#L201-L209)); the stateful SDK transport validates every outbound harness message before sending ([sdk-client.ts:L91-L95](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/connection/sdk-client.ts#L91-L95)), and the runner appends those wire checks ([server.ts:L147-L153](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/runner/server.ts#L147-L153)). Because `banana` is outside `LoggingLevel`, this is reported as `wire-schema-harness-error` `FAILURE` ([wire-schema.ts:L434-L448](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/validation/wire-schema.ts#L434-L448)). Please route this intentionally invalid request through an explicit scoped invalid-wire path (while still validating the response), and attach the targeted `checks.json` showing no harness error.

2. **The capability result still understates a normative violation and can turn a failed probe into a compliant skip.** If notifications are observed without `capabilities.logging`, the code returns `WARNING`, although the cited spec says such servers `MUST` declare the capability ([logging.ts:L109-L123](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/scenarios/server/logging.ts#L109-L123)). Server summaries only mark `FAILURE` as failed, so that run remains green ([server.ts:L181-L186](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/runner/server.ts#L181-L186)). Also, any probe exception is swallowed and then described as “no notifications observed”/compliant `SKIPPED` ([logging.ts:L103-L138](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/scenarios/server/logging.ts#L103-L138)), despite the repository’s missing-prerequisite policy ([untestable.ts:L4-L17](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/scenarios/untestable.ts#L4-L17)). Please distinguish a clean zero-notification probe from a probe failure, and make the observed MUST violation a `FAILURE`.

3. **Threshold filtering is described as non-scoring but still scores positive results.** The positive branches emit `SUCCESS` ([logging.ts:L373-L388](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/scenarios/server/logging.ts#L373-L388), [L434-L450](https://github.com/modelcontextprotocol/conformance/blob/7bd53511b83731093c9e356e27fed685c4bb6c7f/src/scenarios/server/logging.ts#L434-L450)), and `SUCCESS` is included in the score denominator. Also, `LoggingThresholdAllowsAll` succeeds on any non-empty notification list, so an error-only result passes even though the fixture contract says at least debug and error. Please use a genuinely non-scoring status consistently and either verify the claimed level set or narrow the claim.

Finally, the PR currently has no checks reported, and the body/thread still does not include the requested targeted wire output or the real-SDK positive plus specific failing result required by [CONTRIBUTING.md:L67-L70](https://github.com/modelcontextprotocol/conformance/blob/7169291ec0b68eb370fddcd9947313ab0d5e4156/CONTRIBUTING.md#L67-L70). Once those are attached and the three code paths above are corrected, I’m happy to re-check.

PR #454, opened 2026-08-13
fix(runner): bound server scenarios with a per-scenario timeout

### LINKED ISSUE #427: Server runner has no per-scenario timeout; a wedged server stalls the whole suite
## Problem

`src/runner/server.ts` awaits `scenario.run(ctx)` with no bound. The client runner already has a 30s per-scenario timeout (`src/runner/client.ts`), but on the server side any scenario that blocks on a wedged server-under-test (a server that accepts the TCP connection but never responds to an initialize POST, a tools call, or an SSE read) stalls the entire suite run indefinitely.

#316 bounded one hang class (the session-termination DELETEs in teardown, via `AbortSignal.timeout`), but that is call-site depth: every raw `fetch` in every scenario would need the same treatment to cover the rest.

## Proposal

Bound `scenario.run` at the runner layer, mirroring the client runner's per-scenario timeout (probably with the existing `--timeout` option applying to server mode too). Per-call bounds like #316's then become defense-in-depth rather than the only protection.

## Repro

Point `server --scenario server-initialize` (pre-#316) at a mock that accepts connections but never responds: the run never completes.

aviseth:
Closes #427.

The client runner has bounded each scenario since it was written. The server runner never did, so a server that accepts the TCP connection and then stops answering stalled the whole run instead of failing its own scenario.

`--timeout` now works in server mode, same 30s default as client mode. When it fires, the scenario gets a `scenario-timeout` FAILURE naming the bound and the suite carries on.

The one decision I'd like a second opinion on: the timed-out `scenario.run` promise is left pending rather than cancelled. A read blocked on a socket has nothing to cancel through, so there is nothing to await — I swallow its rejection so a late failure can't land on whatever scenario is running by then. Threading an `AbortSignal` through `RunContext` would let scenarios actually tear down, but that touches every scenario and seemed like a separate change. Say the word and I'll do it that way instead.

Tested by pointing `server-initialize` at a server that completes the handshake and never writes a byte. With the runner change reverted the test hangs until vitest kills it at 30s. Full suite, typecheck and lint pass.

#316's per-call bounds still earn their keep — they fail faster and attribute the failure better. They're just no longer the only thing between a wedged server and an unbounded run.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -555,11 +555,13 @@ program
     '--requirements <revision>',
     'Run exactly the scenarios a spec revision requires, frozen at its release (e.g. 2026-07-28). Replaces --suite and --spec-version'
   )
+  .option('--timeout <ms>', 'Per-scenario timeout in milliseconds', '30000')
   .option('--verbose', 'Show verbose output (JSON instead of pretty print)')
   .action(async (options, cmd) => {
     try {
       // Validate options with Zod
       const validated = ServerOptionsSchema.parse(options);
+      const timeout = parseInt(options.timeout, 10);
 
       const verbose = options.verbose ?? false;
       const outputDir = options.outputDir;
@@ -588,7 +590,8 @@ program
           validated.scenario,
           outputDir,
           specVersionFilter,
-          options.force ?? false
+          options.force ?? false,
+          timeout
         );
 
         // Inapplicable scenario/spec-version combination (already logged by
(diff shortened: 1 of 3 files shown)


aviseth:
@pcarleton against #427, which you filed. No CI has run here — the workflow runs are sitting in `action_required`.

One open question from the description, if you have a view: I left the timed-out `scenario.run` pending rather than threading an `AbortSignal` through `RunContext`. The latter gives real teardown but touches every scenario. Happy to switch.

pcarleton (approved):
Thanks, leaving the timed-out run pending is good enough for now; we can worry about threading an AbortSignal through RunContext if we find we need that later. One nit for a follow-up: validate --timeout as a positive integer, parseInt lets 30s through as 30 ms and NaN makes every scenario time out immediately.

PR #455, opened 2026-08-13
fix(lifecycle): bound the session-id probe and release its response body

### LINKED ISSUE #428: lifecycle.ts session-id probe: raw initialize fetch is unbounded and never consumes the response body
## Problem

The session-id visible-ASCII check in `src/scenarios/server/lifecycle.ts` sends a raw `fetch` POST (initialize) with no timeout signal, and the response body is never read or cancelled.

Two failure modes:
- A server that accepts the connection but never responds hangs the scenario before teardown ever runs, so the DELETE bound added in #316 never comes into play.
- A server that answers with an open SSE stream (`Accept` includes `text/event-stream`) leaves `response.body` unconsumed, leaking the connection.

## Proposal

Add `signal: AbortSignal.timeout(...)` to the probe fetch and cancel the body after reading the header. ~5 lines.

Related: #427 (runner-level per-scenario timeout) is the structural fix for the hang class; this one is still worth doing for fast failure and clean check attribution.

aviseth:
Closes #428.

The visible-ASCII session-id check sends its own initialize POST, outside the SDK transport, because the transport doesn't expose the `MCP-Session-Id` header. That fetch had no abort signal and nothing ever read or cancelled the body.

So a server that accepts and never answers hung the scenario before teardown, which meant the DELETE bound from #316 never got a chance to run. And since the probe sends `Accept: application/json, text/event-stream`, a server answering with an open stream left the body dangling for the rest of the run.

5s to match `SESSION_TERMINATE_TIMEOUT_MS`, plus `response.body?.cancel()` once the header has been read.

Branched off main rather than stacking on the #427 branch, since you noted this is worth having on its own even after the runner-level bound lands. Two tests, both fail against the unmodified probe. Full suite passes.

### DIFF
--- a/src/scenarios/server/lifecycle.test.ts
+++ b/src/scenarios/server/lifecycle.test.ts
@@ -140,4 +140,48 @@ describe('ServerInitializeScenario', () => {
     );
     expect(deleteCalls).toHaveLength(0);
   });
+
+  it('bounds the session-id probe so a server that never answers cannot hang it', async () => {
+    fetchMock.mockResolvedValue(new Response(null));
+
+    await new ServerInitializeScenario().run(testContext(serverUrl));
+
+    const probe = fetchMock.mock.calls.find(
+      ([, init]) => (init as RequestInit | undefined)?.method === 'POST'
+    );
+    expect(probe?.[1]?.signal).toBeInstanceOf(AbortSignal);
+  });
+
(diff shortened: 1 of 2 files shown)


aviseth:
@pcarleton against #428. Same as #454 — the workflow runs need approving before any CI runs.

Independent of #454, so it doesn't need to wait on it.

pcarleton (approved):

PR #466, opened 2026-08-21
feat(authorization): send the RFC 8707 `resource` parameter

### LINKED ISSUE #465: `authorization-code-grant` never sends the RFC 8707 `resource` parameter
In the `authorization` scenarios the runner is the MCP client. The spec:

> The `resource` parameter:
>
> 1. **MUST** be included in both authorization requests and token requests.

and, further down:

> MCP clients **MUST** send this parameter regardless of whether authorization
> servers support it.

`src/scenarios/authorization-server/authorization-code-grant.ts` sends it in
neither. There's no `--resource` flag and no `resource` field on
`AuthorizationServerOptionsSchema`, so `--file` can't supply it either.

The suite asserts this in the other direction — `token-endpoint-auth.ts` emits
`resource-parameter-in-authorization` and `resource-parameter-in-token` (#118).
So the harness fails a requirement it enforces elsewhere.

An authorization server that mishandles `resource` can't fail these scenarios:
the parameter never arrives.

### Why a flag rather than discovery

#110 suggested getting `resource` via discovery. That works in the `client`
direction, which has protected-resource metadata. This direction has none:
`--url` is the authorization server issuer, and the only document fetched is
RFC 8414 server metadata, which never names the MCP server. So it has to come
from the operator.

### Not proposing a new AS-side assertion

RFC 8707 §2 makes AS audience restriction a SHOULD, and MCP puts no `resource`
obligation on authorization servers at all. This isn't "the AS must honor it" —
it's the harness sending what a client sends. The PR adds no check IDs.

### Open question: what should a run without `--resource` do?

If clients MUST always send it, why is sending it optional? Three options:

1. **Silent** — nothing changes for existing users, but the MUST goes
   unexercised and the report doesn't say so.
2. **Fail** — matches the MUST → FAILURE rule in AGENTS.md, but every existing
   `authorization` run fails until operators pass the flag.
3. **Untestable at WARNING** (`untestableCheck`, per #248) — visible rather than
   silently green, but softer than the keyword rule implies.

The PR does (1). Happy to switch — your call.

### Related

- Requirement origin: modelcontextprotocol/modelcontextprotocol#734
- Client-side precedent: #118
- modelcontextprotocol/modelcontextprotocol#1614 is an open request to make
  `resource` OPTIONAL. If that lands this becomes conditional; today's text is a
  MUST.

PR: #466

tylerklose:
Fixes #465.

The runner is the MCP client in the `authorization` direction. The spec requires
clients to send `resource` in both the authorization request and the token
request. It sent neither.

### Changes

- `src/index.ts` — `--resource <uri>` on the `authorization` command.
- `src/schemas.ts` — `resource` on `AuthorizationServerOptionsSchema`, so
  `--file` supplies it too.
- `authorization-code-grant.ts` — sends `resource` in the authorization request
  and the token request when it's set.

The value must be an absolute URI with no fragment — the spec's own two invalid
examples are `mcp.example.com` (no scheme) and `https://mcp.example.com#fragment`.

Two conditionals, one per request. It's only sent when the flag or file field is
present, so existing runs are unchanged.

### No new check IDs

Deliberate. Here the runner *is* the client, so a check for "the runner sent
`resource`" is the harness grading its own homework — it can never fail. URI
validation happens in the schema, before scenarios run, so a check for that can't
fail either.

None is needed. The existing grant checks now reach a path they couldn't before:
an AS that mishandles `resource` fails them, where before it was never asked.

### Tests

Six cases added (`npm test`: 530 passing).

Scenario:

- `resource` appears in both the authorization URL and the token body when set
- it appears in neither when unset — guards against it silently becoming
  unconditional
- a mock authorization server that rejects `resource` makes the existing grant
  check fail, and passes when the parameter is absent

Schema:

- accepts an absolute URI
- rejects a fragment
- rejects a relative URI

Revert the two conditionals and the first and third fail.

### Validated against a real authorization server

Run against hitch-rails, my own Rails MCP server and its own authorization
server. It requires `resource` at both endpoints, and rejects a
token exchange whose `resource` doesn't match the one recorded at authorization
time. The public-client grant passes end to end — it couldn't have before this
change, because the parameter never arrived.

The confidential-client path wasn't exercised: `selectTokenAuthMethod` returns
`none` whenever an AS advertises `none`, even with a client secret supplied, so
the runner shows up as a public client against a client registered for
`client_secret_basic`. Unrelated to this change; worth raising separately.

### Open question

See the issue: a run without `--resource` stays silent about the requirement.
Whether it should fail instead, or report untestable at WARNING per #248, is your
call.

### Checklist

- `npm run build` passes
- `npm test` passes (530)
- eslint + prettier clean
- No scenario added or removed; no check IDs changed, so no traceability drift
- Run against a real authorization server (above)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -764,6 +764,10 @@ program
     'OAuth client secret (omit for public/PKCE-only clients)'
   )
   .option(
+    '--resource <uri>',
+    'Canonical URI of the MCP server the token is for (RFC 8707 resource parameter)'
+  )
+  .option(
     '-p, --port <port>',
     'Port for the local OAuth callback server; register http://127.0.0.1:<port>/callback as a redirect URI',
     (value) => Number(value),
--- a/src/scenarios/authorization-server/authorization-code-grant.test.ts
+++ b/src/scenarios/authorization-server/authorization-code-grant.test.ts
@@ -18,6 +18,8 @@ const SERVER_URL = 'https://example.com';
 const AUTHORIZATION_ENDPOINT = `${SERVER_URL}/auth`;
 const TOKEN_ENDPOINT = `${SERVER_URL}/token`;
 
+const RESOURCE = 'https://mcp.example.com/mcp';
+
 const OPTIONS = {
   url: SERVER_URL,
   clientId: 'client',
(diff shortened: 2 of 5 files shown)


pcarleton (approved):
LGTM, thanks. small, opt-in, and well tested. One nit if you feel like it: add a resource example to examples/authorization-server-settings.example.json. 

@tnorimat this is in your area; can you give it a review, and merge when you're happy with it?

tnorimat (approved):
LGTM

tnorimat:
@pcarleton Hello, I reviewed, approved and merged the PR.

tnorimat:
@tylerklose Thank you!

PR #468, opened 2026-08-26
fix(validation): accept extension result envelopes

### LINKED ISSUE #424: wire-schema validation rejects SEP-2663 task envelopes - no resultType "task" branch
The wire-schema validator introduced in #399 discriminates `resultType === 'input_required'` before falling back to the request method's result definition, but has no branch for `resultType === 'task'`. A well-formed SEP-2663 task envelope answering `tools/call` is therefore validated against `CallToolResult` and fails with `must have required property 'content'`.

The envelope shape cannot be present in the core spec schema, since tasks moved to an extension as of SEP-2663. So any SDK that implements the tasks extension and passes wire validation on everything else still fails these checks. In our runs (mcpkit) all 8 tasks scenarios fail exactly one check each (`wire-schema-valid`) for this reason but every other check in those scenarios passes.

Sample violation from `tasks-lifecycle`.  here is the response that the validator rejected:

```json
{
 "origin": "implementation",
 "specVersion": "2026-07-28",
 "context": "response to 'tools/call'",
 "errors": [
  "CallToolResult: must have required property 'content' (result of 'tools/call')"
 ],
 "message": {
  "jsonrpc": "2.0",
  "id": 2,
  "result": {
   "_meta": {
    "io.modelcontextprotocol/serverInfo": {
     "name": "tasks-v2-demo",
     "version": "0.1.0"
    }
   },
   "createdAt": "2026-07-29T20:36:13Z",
   "lastUpdatedAt": "2026-07-29T20:36:13Z",
   "pollIntervalMs": 1000,
   "resultType": "task",
   "status": "working",
   "taskId": "task-42451aab21bbd828f4db0c0d",
   "ttlMs": 300000
  }
 }
}
```

The result carries `resultType: "task"` and the SEP-2663 envelope fields.  There is nothing for `CallToolResult.content` to match because it is not a CallToolResult.

Possible fixes, in rough order of preference:

1. Treat an unrecognized `resultType` value as an extension result and skip typed validation for it (fall back to the generic `JSONRPCResultResponse` envelope check), since `resultType` is exactly the discriminator the final revision added for this purpose.
2. Vendor the tasks extension's envelope schema alongside the core schema and add a `task` branch, if extension-aware validation is wanted.

Happy to contribute a PR for either direction if maintainers have a preference.

jstar0:
## Summary

- Treat open resultType discriminators such as SEP-2663 task as extension results.
- Validate extension results against the generic JSON-RPC result envelope until a versioned extension schema is available.
- Add regression coverage for a task result returned from tools/call.

## Verification

- npm run build
- npm test (44 test files, 525 tests passed)
- npm run check (typecheck, ESLint, and Prettier)
- Pre-push hook passed (format check and full test suite)

Fixes #424

### DIFF
--- a/src/validation/wire-schema.test.ts
+++ b/src/validation/wire-schema.test.ts
@@ -117,6 +117,24 @@ describe('wireSchemaErrors', () => {
     ).toEqual([]);
   });
 
+  it('accepts extension result types through the generic result envelope', () => {
+    expect(
+      wireSchemaErrors(
+        DRAFT_PROTOCOL_VERSION,
+        {
+          jsonrpc: '2.0',
+          id: 3,
+          result: {
+            resultType: 'task',
+            taskId: 'task-1',
+            status: 'working'
+          }
+        },
+        'tools/call'
+      )
+    ).toEqual([]);
+  });
+
   it('accepts a JSON-RPC batch under 2025-03-26 and reports per-element errors', () => {
     expect(
       wireSchemaErrors('2025-03-26', [
--- a/src/validation/wire-schema.test.ts
+++ b/src/validation/wire-schema.test.ts
@@ -135,6 +135,31 @@ describe('wireSchemaErrors', () => {
     ).toEqual([]);
   });
 
+  it('names an unrecognised resultType when the result then fails its typed definition', () => {
+    const errors = wireSchemaErrors(
+      DRAFT_PROTOCOL_VERSION,
+      { jsonrpc: '2.0', id: 3, result: { resultType: 'complet' } },
+      'tools/call'
+    );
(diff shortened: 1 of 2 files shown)


pcarleton (approved):
Thanks @jstar0. I pushed one small commit on top (49bea9e): only known extension resultTypes (today just "task") take the envelope-only path; any other unrecognised value is still validated as the method's result, with the value named in the failure.

@panyam @LucaButBoring FYI. Follow-ups I'd like your read on: whether "task" is the only value to allow-list for now, and making resultType a proper discriminator in the core schema (we likely need new a spec issue). #473's typed CreateTaskResult validation is something we can consider next.

PR #469, opened 2026-08-26
fix: bound raw session ID initialize probe

### LINKED ISSUE #428: lifecycle.ts session-id probe: raw initialize fetch is unbounded and never consumes the response body

jstar0:
Fixes #428.

The raw initialize probe used to inspect `MCP-Session-Id` had no timeout and never released its response body. A server that accepted the connection without responding could stall the scenario, while a long-lived SSE response could keep the connection open after the header was inspected.

This change:

- bounds the probe with a 5-second `AbortSignal.timeout`;
- cancels the response body after reading the session header, including long-lived SSE responses;
- preserves the existing check IDs, statuses, and best-effort session termination behavior.

Verification:

- `npm test`
- `npm run build`
- `npm run typecheck`
- `npm run lint`

### DIFF
--- a/src/scenarios/server/lifecycle.test.ts
+++ b/src/scenarios/server/lifecycle.test.ts
@@ -27,10 +27,14 @@ describe('ServerInitializeScenario', () => {
 
   afterEach(() => {
     vi.unstubAllGlobals();
+    vi.restoreAllMocks();
   });
 
   it('returns INFO when the server does not provide an MCP-Session-Id header', async () => {
-    fetchMock.mockResolvedValue(new Response(null));
+    const cancelMock = vi.fn();
+    fetchMock.mockResolvedValue(
+      new Response(new ReadableStream({ cancel: cancelMock }))
+    );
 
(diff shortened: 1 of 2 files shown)


pcarleton:
Closing as part of a maintainer backlog sweep: duplicate of #455

PR #479, opened 2026-08-31
Docs: optional TOA verify after conformance

dulrajnr:
## Summary

Docs-only. Optional step to run `toa-verify` when a `toa.json` attestation is present, after MCP protocol conformance.

- README section under GitHub Action
- `SDK_INTEGRATION.md` section for SDK maintainers
- `examples/toa-after-conformance.yml`

Does not add a conformance scenario. TOA is adjacent delivery evidence (`toa/0.1`), not wire-protocol checks. No AgentStatus account is required to verify.

## Test plan

- [ ] README / SDK_INTEGRATION examples are valid YAML
- [ ] Copy is clear that TOA is optional and not a protocol scenario
- [ ] Example workflow path is copy-paste only

Made with [Cursor](https://cursor.com)

### DIFF
--- a/README.md
+++ b/README.md
@@ -305,6 +305,26 @@ steps:
 | `verbose`           | No          | Show verbose output (default: false)            |
 | `node-version`      | No          | Node.js version (default: 20)                   |
 
+### Optional: Tool Outcome Attestation (TOA) after conformance
+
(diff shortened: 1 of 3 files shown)


pcarleton:
Closing as part of a maintainer backlog sweep: if you would like to propose an extension & have conformance tests for it, we could consider that.

PR #481, opened 2026-08-31
feat(server): add SEP-2350 scope challenge scenario

### LINKED ISSUE #480: Add server conformance for SEP-2350 scope challenges
## Goal

Add one language-neutral server conformance scenario for request-time OAuth `insufficient_scope` behavior from SEP-2350. The scenario should exercise every invocable MCP server primitive: `tools/call`, `resources/read` for both a static URI and a template-expanded URI, and `prompts/get`.

## Portable fixture contract

Reuse the existing cross-SDK fixtures so implementations fail on observable wire behavior rather than on a missing test-only primitive:

| Operation | Fixture | Required scopes in one challenge |
| --- | --- | --- |
| Tool | `tools/call` / `test_simple_text` | `mcp:conformance:tools:call`, `mcp:conformance:tools:test_simple_text` |
| Static resource | `resources/read` / `test://static-text` | `mcp:conformance:resources:read`, `mcp:conformance:resources:static` |
| Template resource | `resources/read` / `test://template/123/data` | `mcp:conformance:resources:read`, `mcp:conformance:resources:template:123` |
| Prompt | `prompts/get` / `test_simple_prompt` | `mcp:conformance:prompts:get`, `mcp:conformance:prompts:test_simple_prompt` |

Servers should treat `mcp-conformance-scope-low` as a valid opaque token with only `mcp:conformance:baseline`, and `mcp-conformance-scope-full` as a valid opaque token with all required scopes. Unauthenticated behavior remains unchanged for existing scenarios.

For each operation, the low token should produce HTTP 403 and a Bearer `WWW-Authenticate` challenge containing `error="insufficient_scope"`, the fixture protected-resource metadata URL, and both required scopes in one challenge. Retrying the same operation with the full token should produce the normal successful MCP result.

## Rationale

`src/seps/sep-2350.yaml` currently excludes the server single-challenge requirement because the challenge was the only source of truth for required scopes. The fixed fixture contract supplies independent ground truth, so `sep-2350-server-single-challenge` can become an observable check.

This deliberately does not test general token validation or require a mock authorization server; it isolates the SDK/server request-time challenge seam. It also avoids the TypeScript-specific client/auth imports and guessed `admin-action` fixture in the older, unmerged #106 approach.

## Proof

Run the official scenario through the existing SDK runner against:

1. Unmodified `modelcontextprotocol/typescript-sdk@main`: build succeeds, but low-token operations return HTTP 200 and the scenario exits non-zero.
2. `SamMorrowDrums/typescript-sdk@scope-challenge-server-sdk` (modelcontextprotocol/typescript-sdk#1624): all challenge and upgraded-retry checks pass.

## Related

- SEP-2350: modelcontextprotocol/modelcontextprotocol#2350
- Existing client coverage: #281 / #282
- Existing server-row exclusion: #302
- Older overlapping proposal: #106
- TypeScript SDK implementation: modelcontextprotocol/typescript-sdk#1624

SamMorrowDrums:
## Summary

- add a language-neutral server conformance scenario for request-time OAuth `insufficient_scope` challenges
- exercise `tools/call`, static and templated `resources/read`, and `prompts/get` with a fixed cross-SDK fixture contract
- assert HTTP 403, ****** parameters, the complete required scope set in one challenge, and successful upgraded-token retries
- restore `sep-2350-server-single-challenge` traceability and document the portable SUT contract
- make normative server-side SHOULD warnings gate targeted conformance runs consistently while leaving optional/inconclusive diagnostics informational

The scenario uses existing primitive fixtures and opaque test tokens, so the negative result is observable wire behavior rather than a missing TypeScript API or test-only primitive. The core scenario does not import or depend on the TypeScript SDK.

## A/B proof

Both legs used this conformance branch's official `sdk ... --mode server --scenario sep-2350-server-scope-challenge --spec-version 2026-07-28` path.

| TypeScript SDK ref | Result |
| --- | --- |
| unmodified `modelcontextprotocol/typescript-sdk@main` (`dcc01028`) | `SUCCESS=5`, `WARNING=12`, exit `1`; four warnings each for HTTP 403, `WWW-Authenticate`, and `sep-2350-server-single-challenge`; all four full-token retries succeeded |
| `SamMorrowDrums/typescript-sdk@scope-challenge-server-sdk` (`54e8deda`) | `SUCCESS=17`, no warnings or failures, exit `0` |

The feature fixture adapter is part of [modelcontextprotocol/typescript-sdk#1624](https://github.com/modelcontextprotocol/typescript-sdk/pull/1624) and remains independent from this branch. The latest green run includes the maintainer-requested reuse of bearer-auth metadata configuration through verified `AuthInfo`.

## Validation

- `npm test` — 45 files, 541 tests passed
- `npm run lint`
- `npm run typecheck`
- `npm run build`
- targeted compliant and deliberately broken scope-challenge fixture tests
- TypeScript SDK feature head `54e8deda` — 17/17 official scenario checks passed

## References

- Closes #480
- [SEP-2350](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2350)
- [TypeScript SDK implementation](https://github.com/modelcontextprotocol/typescript-sdk/pull/1624)
- Existing client-side coverage: #282
- Prior server-row traceability decision: #302

### DIFF
--- a/src/requirements.ts
+++ b/src/requirements.ts
@@ -3,6 +3,7 @@ import { dirname, join } from 'path';
 import { fileURLToPath } from 'url';
 import { parse as parseYaml } from 'yaml';
 import { ALL_SPEC_VERSIONS } from './scenarios';
+import type { ConformanceCheck } from './types';
 
 /**
  * A frozen requirement set for one specification revision: the scenarios an
@@ -249,4 +250,43 @@ export function filterScenariosByRequirements(
     );
   }
   return wanted.filter((name) => available.has(name));
+}
+
+function isBlockingCheck(check: ConformanceCheck): boolean {
+  return check.status === 'FAILURE' || check.status === 'WARNING';
+}
+
+/**
+ * Exit status under a frozen requirement set. Scored failures and warnings
+ * block conformance; not-scored scenarios are reported without changing it.
+ */
+export function requirementsExitCode(
+  requirements: RequirementSet,
+  leg: Leg,
+  results: { scenario: string; checks: ConformanceCheck[] }[]
+): number {
+  const scored = new Set(scoredScenarios(requirements, leg));
+  const unscored = results.filter((result) => !scored.has(result.scenario));
+  if (unscored.length > 0) {
+    const blocking = unscored.filter((result) =>
+      result.checks.some(isBlockingCheck)
+    );
+    console.log(
+      `\nNot scored for ${requirements.revision}: ${unscored.length} scenario(s) run, ${blocking.length} non-green. These do not affect conformance.`
+    );
+    for (const result of unscored) {
+      const why = notScoredScenarios(requirements, leg).find(
+        (entry) => entry.scenario === result.scenario
+      );
+      const isBlocking = result.checks.some(isBlockingCheck);
+      console.log(
+        `  ${isBlocking ? '\u2717' : '\u2713'} ${result.scenario} (${why?.reason ?? 'not scored'})`
+      );
+    }
+  }
+
+  const scoredBlocked = results
+    .filter((result) => scored.has(result.scenario))
+    .some((result) => result.checks.some(isBlockingCheck));
+  return scoredBlocked ? 1 : 0;
 }
--- a/src/scenarios/server/scope-challenge.ts
+++ b/src/scenarios/server/scope-challenge.ts
@@ -1,0 +1,433 @@
+import {
+  type ClientScenario,
+  type ConformanceCheck,
+  DRAFT_PROTOCOL_VERSION
+} from '../../types';
+import {
+  buildStandardHeaders,
+  sendStatelessRequest,
+  withRequestMeta,
+  type RunContext,
+  type StatelessResponse
+} from '../../connection';
+
+export const SCOPE_CHALLENGE_LOW_TOKEN = 'mcp-conformance-scope-low';
+export const SCOPE_CHALLENGE_FULL_TOKEN = 'mcp-conformance-scope-full';
+
+interface ScopeChallengeFixture {
+  key: string;
+  label: string;
+  method: string;
+  params: Record<string, unknown>;
+  requiredScopes: readonly [string, string];
+  validateResult(result: Record<string, unknown>): string | undefined;
+}
+
+function firstArrayItem(value: unknown): Record<string, unknown> | undefined {
+  if (!Array.isArray(value)) return undefined;
+  const first = value[0];
+  return typeof first === 'object' && first !== null
+    ? (first as Record<string, unknown>)
+    : undefined;
+}
+
+export const SCOPE_CHALLENGE_FIXTURES: readonly ScopeChallengeFixture[] = [
+  {
+    key: 'tool',
+    label: 'Tool',
+    method: 'tools/call',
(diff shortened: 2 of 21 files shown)


src/requirements.ts
```diff
@@ -250,3 +251,42 @@ export function filterScenariosByRequirements(
   }
   return wanted.filter((name) => available.has(name));
 }
+
+function isBlockingCheck(check: ConformanceCheck): boolean {
+  return check.status === 'FAILURE' || check.status === 'WARNING';
```
panyam, line 256: I can see the ratioanle for this.    INFO already exists for diagnostic so so this is a good improvement

panyam: One thing though - If other tests/scenarios that are using WARNING we would see all sdks testing against those would start blocking?

src/scenarios/server/scope-challenge.ts
```diff
@@ -0,0 +1,473 @@
...
+
+function challengeChecks(
+  fixture: ScopeChallengeFixture,
+  response: ChallengeResponse | undefined,
+  requestError: unknown,
+  serverUrl: string
+): ConformanceCheck[] {
+  const parsed = parseBearerChallenge(response?.wwwAuthenticate ?? null);
+  const scopes = parsed?.params.scope?.split(/\s+/).filter(Boolean) ?? [];
+  const metadataError = resourceMetadataUrlError(
```
panyam, line 318: If parsed is null - ie because of no WWW-Authenticate, or one that doesn't parse properly - then scopes is [] and hasAllScopes is false, so the check reports an incomplete challenge even though no challenge was issued. I hit this with resources/read in mcpkit.   My server returned 200, and all three challenge checks reported, though only the 403 one was actionable.

I think this is could use a fix because this one mapped to the SEP requirement, so the way it is written it is hard to tell "violated the single-challenge SHOULD" from "doesn't implement scope challenges". The latter is already covered by server-scope-challenge-http-403.  I think "SKIPPED when parsed is null" would better separate them?

Also would this compound with our WARNINGS change?  A server with no support would produce a bunch of blocking rows for one fact (I think I saw 11 or 12 on my side before becomign conformant)?

panyam (approved):

src/requirements.ts
SamMorrowDrums: Audited the affected WARNING producers. Normative SHOULD checks still block; acceptable diagnostics now use INFO, including optional/inconclusive session cases, the non-normative polling fixture notice, and acceptable late retry timing. Tests cover WARNING -> exit 1 and INFO -> exit 0 for requirements and targeted server runs.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

src/scenarios/server/scope-challenge.ts
SamMorrowDrums: Fixed: `sep-2350-server-single-challenge` is now SKIPPED when there is no parseable Bearer challenge. The HTTP/challenge prerequisite checks retain the actionable warnings. Added missing, malformed, and no-challenge regression cases.

> [!NOTE]
> Auto-replied by the [GitHub Copilot app](https://github.com/github/app).

PR #483, opened 2026-09-02
fix(client/auth): bind negative-check verdicts to an observed prerequisite

### LINKED ISSUE #467: Prevent false greens when a negative check rejects for the wrong reason
## Problem

A negative conformance check can report `SUCCESS` when the implementation returns the expected rejection for an unrelated reason before the requirement under test is exercised.

Example: a vector is intended to verify that a declaration absent from a validly signed manifest is rejected. A broken verifier rejects the same input at an earlier version guard. A check that compares only the final verdict sees `reject == reject` and stays green even though manifest membership was never evaluated.

This failure mode was independently reported in the live signed-declarations discussion at modelcontextprotocol/modelcontextprotocol#3140. The thread also identified the security boundary: causal detail should be observable to the harness/auditor, not required on the MCP wire.

## Minimal reproduction

Protocol-neutral fixture: https://gist.github.com/Silentpartnercoding/2c5ddb251d4b60f337b86b0452409950

It runs one vector against a correct verifier and a deliberately short-circuiting mutant:

| Harness | Correct verifier | Short-circuit mutant |
| --- | --- | --- |
| Verdict only | PASS | PASS — false green |
| Reason bound | PASS | FAIL — mutant killed |

The fixture binds four machine-readable observations:

```json
{
  "verdict": "reject",
  "property_under_test": "manifest_membership_binding",
  "property_reached": true,
  "stop_reason": "DECLARATION_NOT_IN_MANIFEST"
}
```

The mutant instead reports `property_reached: false` and `stop_reason: "UNSUPPORTED_VERSION"`. Its final verdict is still `reject`.

Local result on Node v24.18.1: 3/3 tests pass, including the required false-green and mutant-kill assertions.

## Relationship to current conformance behavior

This appears adjacent to, but narrower than, #248:

- `untestableCheck()` already makes a missing prerequisite red and records `details: { untestable: true, reason }`.
- `AGENTS.md` already requires deliberately broken implementations for new scenarios.
- Some helpers deliberately construct otherwise-valid inputs so a negative probe isolates one failure dimension.

Those are strong safeguards. The remaining gap is a reusable contract ensuring that an expected negative verdict came from the check's target property rather than an earlier guard. `ConformanceCheck.details` can already carry the observation, so this may require a convention/helper and negative meta-test rather than a new result status or runner.

## Proposed direction for discussion

For negative checks where multiple rejection paths can produce the same wire result:

1. Name the target with the existing stable check ID or an equivalent `property_under_test` identifier.
2. Emit `SUCCESS` only after the harness has exercised and observed the intended property.
3. If an earlier guard or missing prerequisite prevents that, emit the same check ID as `FAILURE`/`WARNING` with machine-readable `details` identifying that the property was not reached and why.
4. Add a deliberately broken implementation that returns the expected wire verdict through the wrong guard; the vitest must prove the check rejects that mutant.

Questions for maintainers:

- Should this be an explicit negative-check convention in `AGENTS.md`, a small helper over `ConformanceCheck.details`, or simply applied case-by-case?
- Is the existing `untestable` detail shape the right vocabulary to extend, or should causal reachability remain separate from missing prerequisites?
- Once SEP-3140 has a sponsor and a normative spec diff, would its signed-manifest membership check be a useful pilot?

This issue does not propose an MCP wire field, a new CLI entry point, or implementation of SEP-3140 before its requirements are ready.

AI assistance disclosure: OpenAI Codex assisted with the reproduction, tests, analysis, and drafting this issue. I reviewed the fixture, evidence, and draft and remain responsible for the claims and publication.

OfficialAbhinavSingh:
Fixes #467.

## What

Negative checks in the client auth suite bound SUCCESS to a verdict alone, so "the client declined" and "the client never got far enough to decide" were the same observation. `auth/resource-mismatch` is the clearest case: its verdict is `!authorizationRequestMade`, and a client that throws before making a single HTTP request satisfies it.

Before, against a client that performs no discovery request at all:

```
SUCCESS  resource-mismatch-rejected
"Client correctly rejected authorization when PRM resource does not match server URL"
```

After:

```
FAILURE  resource-mismatch-rejected
Error: Not testable: client never requested the Protected Resource Metadata
document, so it never read the resource value it was required to validate
```

`auth/metadata-issuer-mismatch` already gated its verdict on the metadata fetch; the other five checks did not. This makes the policy uniform and routes the unreached case through `notTestable()` (#248) instead of SUCCESS, since the requirement was not violated — it was never exercised.

## Changes

- `resource-mismatch.ts` — verdict requires `prm-pathbased-requested`; absent it, untestable.
- `issuer-parameter.ts` — the four RFC 9207 rejection checks require `authorization-server-metadata` and `authReached`. SEP-2468 conditions each of them on the issuer recorded "from the selected authorization server validated metadata document", so a client that never fetched it cannot have made the comparison. Policy lives in one `issRejectionCheck` function: the duplication is what let five of six sites drift.
- `issuer-parameter.ts` — `metadata-issuer-mismatch` already detected the unreached case but reported it as a plain FAILURE; it now reports untestable.
- All six record `propertyReached` and `stopReason` in `details`, per the machine-readable shape in the issue.
- `examples/clients/typescript/auth-test-inert.ts` — inert client, plus a vitest case asserting the untestable report.

## What this does not close

A client that receives the redirect and then aborts before the token request for an unrelated reason remains indistinguishable from one that rejected on `iss`. Separating those needs a signal from inside the client that a black-box harness does not have. This closes the "never reached the requirement at all" class.

## Validation

- `npm test` — 525 tests passed, 44 files
- `npm run typecheck`, `npx eslint src/ examples/`, `npx prettier --check .` — all clean
- `npm run build`
- Real SDK through the CLI runner: all six affected scenarios PASS against `examples/clients/typescript/everything-client.ts` (built on `@modelcontextprotocol/sdk`), and the prerequisite checks fire for it — `prm-pathbased-requested` and `authorization-server-metadata` both SUCCESS.
- Same runner, inert client: `❌ OVERALL: FAILED`, exit 1. The same client scored SUCCESS before this change.

Written with Claude Code, against this specific issue rather than a generic bug-hunt prompt.

### DIFF
--- a/examples/clients/typescript/auth-test-inert.ts
+++ b/examples/clients/typescript/auth-test-inert.ts
@@ -1,0 +1,22 @@
+#!/usr/bin/env node
+
+import { runAsCli } from './helpers/cliRunner';
+
+/**
+ * Broken client that gives up before performing any discovery request.
+ *
+ * BUG: it never fetches Protected Resource Metadata, so it never reads — let
+ * alone validates — the `resource` value the scenario mismatches on purpose.
+ *
+ * It exists to pin issue #467. `auth/resource-mismatch` decides its verdict
+ * from `!authorizationRequestMade` alone, and a client that does nothing at
+ * all satisfies that verdict, so the check scores SUCCESS for a client that
+ * cannot possibly have performed the validation under test.
+ */
+export async function runClient(_serverUrl: string): Promise<void> {
+  throw new Error(
+    'inert client: aborted before any discovery request (no PRM fetch, no authorization request)'
+  );
+}
+
+runAsCli(runClient, import.meta.url, 'auth-test-inert <server-url>');
--- a/src/scenarios/client/auth/index.test.ts
+++ b/src/scenarios/client/auth/index.test.ts
@@ -26,6 +26,7 @@ import { runClient as noPkceClient } from '../../../../examples/clients/typescri
 import { runClient as reuseCredsClient } from '../../../../examples/clients/typescript/auth-test-reuse-credentials';
 import { runClient as noAppTypeClient } from '../../../../examples/clients/typescript/auth-test-no-application-type';
 import { runClient as noIssValidationClient } from '../../../../examples/clients/typescript/auth-test';
+import { runClient as inertClient } from '../../../../examples/clients/typescript/auth-test-inert';
 import { runClient as issNormalizeClient } from '../../../../examples/clients/typescript/auth-test-iss-normalize';
 import { runClient as echoScopeClient } from '../../../../examples/clients/typescript/auth-test-echo-scope';
 import { runClient as dpopBearerClient } from '../../../../examples/clients/typescript/auth-test-dpop-bearer';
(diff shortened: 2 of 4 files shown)


pcarleton (approved):
Thanks! Routing the never-reached case through untestable() instead of recording a SUCCESS is a good fix for #467, and consolidating the iss checks into one function is a nice cleanup. I verified locally against main with the full suite.

PR #488, opened 2026-09-03
feat(client/auth): check the resource parameter matches the PRM-published identifier

KKonstantinov:
Add a client-auth check that the RFC 8707 `resource` parameter is sent **exactly as published** in protected resource metadata, and have the metadata-discovery scenarios emit the resource checks so the existing root-PRM scenario (`auth/metadata-var2`) covers the pathless case.

## Motivation and Context

Both reference SDKs re-serialized a pathless PRM `resource` such as `https://example.com` through a URL parser and sent `https://example.com/`. Exact-match authorization servers reject that (`AADSTS9010010` on Microsoft Entra ID). See typescript-sdk#1968 (fixed on 2.x by #2581, 1.x backport #1972) and python-sdk#2578 / #2883 (fixed on 2.x by #2925).

The referee could not catch it. The RFC 8707 checks from #118 cover presence, a fragment-free URI, and consistency between the two requests, but never compare the sent value with what the PRM served. The token-endpoint-auth scenarios also serve a path-qualified resource, where normalization is a no-op, and `metadata-var2`, the only bare-origin scenario, did not observe `resource` at all.

The spec requires the `resource` parameter to be the canonical URI, defined as the RFC 9728 `resource` value; RFC 9728 §3.3 requires that value to be identical.

**Changes:**
- **`resource-parameter-matches-prm`** — sent `resource` on both requests must equal the PRM-published string. `FAILURE` on the MUST above; skipped when the PRM was never fetched or no request carried `resource`, like the sibling checks.
- **`helpers/resourceParameterChecks.ts`** — the four existing checks move there verbatim (same IDs and wording) next to the new one.
- **`onPrmRequest` hook on `createServer`** — the mock server reports the `resource` it served.
- **Metadata scenarios emit the five checks** — `metadata-var2` is where the new one bites; no new scenario, per "fewer scenarios, more checks".
- **Negative example** — `auth-test-resource-slash.ts` returns a slash-terminated URL from `validateResourceURL`, with a vitest case asserting the check fails.
- **Self-expiring baseline** — the everything-client runs SDK 1.29, which still has the bug, so `index.test.ts` gets a `knownExampleClientFailures` entry for `metadata-var2`. It is asserted to keep failing, so it must be removed once the example's SDK includes #1972.

## How Has This Been Tested?

- Negative test written first; failed with `Expected failure check resource-parameter-matches-prm not found`, passes now.
- `npm test`: 525 passing. `typecheck`, `lint`, `build` clean.
- Built referee against the TypeScript SDK's `everythingClient` on `auth/metadata-var2`:

  | SDK build | Result |
  |---|---|
  | v2 `main` with #2581, spec 2025-11-25 | 19/19 |
  | v2 `main` with #2581, spec 2026-07-28 | 18/18 |
  | v2 `main` before #2581 | 18/19, only `resource-parameter-matches-prm` fails |

## Breaking Changes

None to the runner or CLI. The three token-endpoint-auth scenarios gain one check and the four metadata scenarios gain five; SDKs reconcile expected-failures when they bump the pin. TypeScript and Python 2.x `main` pass; TypeScript 1.x fails on `metadata-var2` until #1972 ships.

## Types of changes
- [ ] Documentation update

## Checklist

## Additional context

- **Severity.** `FAILURE` rests on the MUST to use the canonical URI. If the spec's trailing-slash note (a SHOULD) is read as governing, flip to `WARNING`; CI treats both as failing.
- **Traceability.** The new ID is not tied to a SEP yaml, so it shows as `untracked` until the manifest workflow runs.
- **Out of scope.** The TypeScript server serves `resource: resourceServerUrl.href` in its OAuth metadata middleware, so a pathless resource server URL is published with a trailing slash (the Python twin was #2883). Server conformance has no PRM scenario yet; separate issue.

### DIFF
--- a/examples/clients/typescript/auth-test-resource-slash.ts
+++ b/examples/clients/typescript/auth-test-resource-slash.ts
@@ -1,0 +1,73 @@
+#!/usr/bin/env node
+
+import { Client } from '@modelcontextprotocol/sdk/client/index.js';
+import { StreamableHTTPClientTransport } from '@modelcontextprotocol/sdk/client/streamableHttp.js';
+import { withOAuthRetryWithProvider } from './helpers/withOAuthRetry';
+import { ConformanceOAuthProvider } from './helpers/ConformanceOAuthProvider';
+import { runAsCli } from './helpers/cliRunner';
+import { logger } from './helpers/logger';
+
+/**
+ * Broken client that re-serializes the protected resource metadata's
+ * `resource` through a URL parser before sending it as the RFC 8707
+ * `resource` parameter.
+ *
+ * BUG: `new URL('https://example.com').href` is `https://example.com/`, so a
+ * pathless resource identifier gains a trailing slash and no longer matches
+ * the value the server published. Authorization servers that compare the
+ * indicator exactly (Microsoft Entra ID, AADSTS9010010) reject the request.
+ * This is the shape of typescript-sdk#1968 and python-sdk#2578.
+ */
+class SlashAppendingResourceProvider extends ConformanceOAuthProvider {
+  async validateResourceURL(
+    _serverUrl: string | URL,
+    resource?: string
+  ): Promise<URL | undefined> {
+    if (!resource) {
+      return undefined;
+    }
+    const url = new URL(resource);
+    // BUG: always emit the normalized form with a trailing slash, whatever
+    // the server published.
+    if (!url.pathname.endsWith('/')) {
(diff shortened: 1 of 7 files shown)


pcarleton (approved):
Thanks! I've also opened a backport for python-sdk here: https://github.com/modelcontextprotocol/python-sdk/pull/3462

Issue #445, opened 2026-08-05
SEP-2243: safe-integer-range MUST is untested, and SDKs disagree on what it requires


halter73:
## The requirement

Released `2026-07-28` `server/tools.mdx`, in the `x-mcp-header` constraint list:

> Integer values **MUST** be within the safe range for integers represented using IEEE754 double-precision floating point numbers (−2<sup>53</sup>+1 to 2<sup>53</sup>−1)

Nothing in the harness covers it: no row in `src/seps/sep-2243.yaml` (the `primitive-only` text stops at "not permitted"), no check ID emitted, and `grep -iE 'safe range|MAX_SAFE_INTEGER|9007199254740991|2\*\*53'` over `src/` returns zero matches. An SDK that mirrors `9007199254740993` into a header passes the suite today.

## It binds at call time, on argument values

Settled by the discussion on [PR #2772](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2772), which introduced the sentence (@mikekistler, merged 2026-05-29):

> As discussed in the Transports WG, I think we aligned on something like: only integer arguments from -2^53 to 2^53 (inclusive) can be mirrored to a header. **Any other numeric values produce an error.**
> — @pja-ant

and the WG meeting notes @mikekistler [pasted into the review](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2772#discussion_r3321131874):

> **Consensus:** Custom numeric headers are strictly restricted to integers. **Clients must reject tool inputs containing non-integer numbers in custom headers.**

So this is a runtime constraint on arguments, not a schema-shape constraint.

## The actual gap: SDKs disagree on the consequence

The spec states the constraint but never says what the client does when an argument violates it. Two SDKs read it two ways:

| SDK | Behavior on an out-of-range integer |
|---|---|
| **csharp-sdk** | throws `McpException` — the call fails ([`McpHeaderExtractor.cs#L116-L121`](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.Core/Client/McpHeaderExtractor.cs)) |
| **typescript-sdk** | `mcpParamPrimitiveToString` returns `undefined`, and `buildMcpParamHeaders` does `continue` — the header is **silently omitted** and the call proceeds ([`mcpParamHeaders.ts#L211, L311`](https://github.com/modelcontextprotocol/typescript-sdk/blob/main/packages/core-internal/src/shared/mcpParamHeaders.ts)) |

[...]

PR #499, opened 2026-09-09
feat(client): validate safe integer range for x-mcp-header parameters (#445)

Yudis-bit:
Related to #445.

An out-of-range integer in the ordinary `test_custom_headers` call can make a client reject that entire call before sending it, incorrectly failing several unrelated header requirements. This change gives the integer-range probe its own tool and places its `toolCalls` entry last, after the ordinary encoding and null-handling calls.

## Changes

- Declare `sep-2243-x-mcp-header-integer-safe-range` in `sep-2243.yaml` and the scenario's emitted check IDs.
- Advertise `test_custom_headers_unsafe_integer` with an annotated integer argument and request the exactly representable, out-of-safe-range value `9007199254740992` (`2^53`). The ordinary tool no longer contains this argument.
- Fail the range check if the dedicated call sends `Mcp-Param-UnsafeInteger`; retain the proposed wire-level success result when the requested argument arrives without that header.
- If the probe never arrives, emit a single range-check failure with `details.untestable: true` and a diagnostic naming the dedicated tool and possible local rejection. Absence alone cannot distinguish rejection from a skipped call. A missing or changed probe argument also cannot produce a success result.
- Cover isolation, tool ordering/schema, mirrored and omitted headers, altered/missing arguments, and idempotent backfill in regression tests.
- Leave the generated traceability manifest to the scheduled workflow; remove the hand-edited coverage entry from this PR.

## Specification decision still needed

The [discussion on #445](https://github.com/modelcontextprotocol/conformance/issues/445#issuecomment-5448966883) raises both definition-time versus call-time enforcement and error versus omission. This implementation exercises the proposed call-time, wire-level range constraint; it does **not** settle those questions or establish that silent omission is the normative recovery behavior.

A locally rejecting client still gets an **untestable failure for the range check**, following the harness's missing-prerequisite policy. Its ordinary header checks remain independently testable. Maintainer guidance is still needed before treating this proposal as the final conformance policy. The rejection validation below is a local fixture, not a C# SDK run.

## Validation

- `npm run check` — typecheck, ESLint, and Prettier pass.
- `npm run build` — passes.
- `npm test` — **45 test files, 595 tests pass**, including 15 custom-header tests.
- Real TypeScript SDK at `b65426158ed9f29aea8ef3dc09ca22d7d9d6f970`, using its unmodified `test/conformance/src/everythingClient.ts`, with the SDK packages built from source:

  ```sh
  node dist/index.js client \
    --command "node --import tsx .sdk-under-test/typescript-sdk/test/conformance/src/everythingClient.ts" \
    --scenario http-custom-headers --spec-version 2026-07-28 \
    -o results/pr-499-typescript
  ```

  **19/19 checks pass; client and runner exit 0.** The client used Node **22.14.0** on Windows. Node 24.19.0 produced the same 19 passing wire checks but crashed during client shutdown with a libuv assertion, so it is not counted as a passing end-to-end run.

- Negative CLI validation with the same SDK and a local fetch wrapper that injects `Mcp-Param-UnsafeInteger: 9007199254740992` only for the dedicated probe: **18/19 pass; only the integer-range check fails**, runner exit 1.
- Local-rejection CLI validation with the same SDK and a fetch wrapper that throws before transmitting the dedicated probe: **18/19 pass; only the integer-range check fails**, with `untestable: true` and the dedicated-tool diagnostic; client and runner exit 1.

AI assistance disclosure: Initial implementation used Antigravity; review fixes and validation used Codex.

### DIFF
--- a/src/scenarios/client/http-custom-headers.test.ts
+++ b/src/scenarios/client/http-custom-headers.test.ts
@@ -86,6 +86,7 @@ describe('HttpCustomHeadersScenario (SEP-2243) check IDs', () => {
             arguments: {
               region: 'us-west1',
               priority: 42,
+              unsafe_integer_val: 9007199254740992,
               non_ascii_val: nonAscii,
               query: 'SELECT 1'
             }
@@ -188,6 +189,7 @@ describe('HttpCustomHeadersScenario (SEP-2243) check IDs', () => {
             arguments: {
               region: 'us-west1',
               priority: 42,
+              unsafe_integer_val: 9007199254740992,
               non_ascii_val: nonAscii,
               query: 'SELECT 1'
             }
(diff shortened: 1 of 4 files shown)


Danny-Devs:
This will fail csharp-sdk on five checks, and the failure message will name the wrong cause.

The oversized integer goes into the existing `test_custom_headers` call. csharp-sdk throws while building headers ([`McpHeaderExtractor.cs#L118`](https://github.com/modelcontextprotocol/csharp-sdk/blob/main/src/ModelContextProtocol.Core/Client/McpHeaderExtractor.cs#L118)), so the call is never sent. With no call arriving, the backfill at [`http-custom-headers.ts#L238`](https://github.com/modelcontextprotocol/conformance/blob/cf3e28a63bd0bd1f7d41b5ecf3f12eb47cb06ced/src/scenarios/client/http-custom-headers.ts#L238) marks every declared ID it never saw as FAILURE.

Five of the six declared IDs are emitted only from the `test_custom_headers` branch: `supports-custom-headers`, `mirrors-designated-params`, `encode-values`, `base64-unsafe`, and the new `integer-safe-range`. All five go red reporting that the client never called the tool. The real cause is one argument out of seventeen. The sixth, `omit-null`, comes from the separate `test_custom_headers_null` call and may survive.

Give the probe its own tool and its own `toolCalls` entry, placed last. A client that refuses the value then fails one check instead of five.

@JosephDoUrden asked for this issue on August 28 and built the same check, then held it pending a maintainer steer ([comment](https://github.com/modelcontextprotocol/conformance/issues/445#issuecomment-5448966883)). That comment has the four-SDK survey and the case for whether the constraint binds at definition time or at call time. The dedicated probe tool above is their design.

CONTRIBUTING asks for a run against at least one real SDK before opening a PR. Validation lists typecheck, lint, build and the unit suite, but no SDK run. The unit tests build the request and its headers by hand, so they cannot catch a client that never sends the call.

The bigger question is for the maintainers. Scoring omission as SUCCESS answers the open question in that comment, and makes csharp-sdk non-conforming by merge rather than by decision.

Yudis-bit:
@Danny-Devs Thanks for catching the coupled failures and pointing me to @JosephDoUrden's earlier design and SDK survey. I've pushed the isolation fix in 7482d377a7babfea783cf46cd7f92fa291271e06.

- Moved the unsafe integer into `test_custom_headers_unsafe_integer`, with its own `toolCalls` entry placed last. The ordinary encoding and null-handling calls run first.
- A missing probe now produces only the range-check failure, marked `details.untestable: true`. Its diagnostic names the dedicated tool and explicitly acknowledges possible local rejection. A missing or changed probe argument cannot pass the check either.
- Added regression coverage and ran the real TypeScript SDK at `b65426158ed9f29aea8ef3dc09ca22d7d9d6f970`: **19/19 checks pass**, with client and runner exit 0 using Node 22.14.0 on Windows. The PR description includes the command and the Node 24 shutdown limitation.
- Also ran that SDK through local fetch fixtures: injecting the unsafe header gives **18/19 passing**, with only the range check failing; throwing before transmitting the probe gives **18/19 passing**, with only the range check marked untestable. The latter simulates local rejection; it is not a C# SDK run.

All **595 tests across 45 files**, typecheck, lint, build, and pre-push checks pass.

The specification concern remains open. This revision retains the proposed wire-level success result for omission, and a rejecting client still receives an untestable failure. Isolation fixes the misleading collateral failures; it does not resolve definition-time versus call-time enforcement or error versus omission. I've made that limitation explicit in the code and PR description and removed the description's claim to close #445. Maintainer guidance is still needed before merging this as conformance policy.

PR #510, opened 2026-09-21
fix(everything-server): classify by the wire's markers, not by the presence of _meta

### LINKED ISSUE #506: Everything-server rejects valid _meta on stateful initialize
## Problem

The TypeScript everything-server misclassifies a stateful `initialize` request as stateless whenever `params._meta` is present. It rejects a valid 2025-11-25 request with HTTP 400 and `-32020 Missing MCP-Protocol-Version header`.

`InitializeRequestParams` extends `RequestParams` in the 2025-11-25 schema, so `_meta` is allowed on `initialize`. MCP Python SDK 2.x currently emits `_meta: {}` there, which makes the conformance fixture unusable as an upstream server for that client.

## Reproduction

Start the fixture:

```sh
npx tsx examples/servers/typescript/everything-server.ts
```

Then send:

```sh
curl -i -X POST http://127.0.0.1:3000/mcp \
  -H 'Accept: application/json, text/event-stream' \
  -H 'Content-Type: application/json' \
  --data '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2025-11-25","capabilities":{},"clientInfo":{"name":"probe","version":"1"},"_meta":{}}}'
```

Observed:

```json
{"jsonrpc":"2.0","id":1,"error":{"code":-32020,"message":"Missing MCP-Protocol-Version header"}}
```

The same request succeeds when `_meta` is omitted.

## Suggested fix

Route `initialize` through the stateful transport when `params.protocolVersion` names a session-era protocol, independently of whether optional request metadata is present. Add an integration regression test that sends `_meta: {}` to the everything-server and asserts a successful initialize response.

po-et:
Fixes #506.

## What's wrong

The fixture decides which wire a request is on by asking whether `params._meta` is present:

```ts
const isLegacySessionEraRequest =
  meta === undefined && reqVersion !== undefined && LEGACY_SESSION_PROTOCOL_VERSIONS.includes(reqVersion);
if (!sessionId && (reqVersion || meta) && !isLegacySessionEraRequest) { … }
```

But `_meta` is part of the base request shape in *every* revision — progress tokens, trace context, anything a client attaches — so a session-era request that carries it is routed to the 2026-07-28 path and rejected for metadata it never owed. Measured on `7169291`:

```
initialize 2025-11-25 + "_meta": {}                      -> 400  -32020 Missing MCP-Protocol-Version header
initialize 2025-11-25 + "_meta": {} + era header         -> 400  -32602 Invalid params: missing _meta or required fields
initialize 2025-11-25 + "_meta": {"progressToken": "p1"} -> 400  -32602 (same line)
initialize 2025-11-25, no _meta                          -> 200
```

The second and third are the same root cause as the reported one, reached through the other branch. MCP Python SDK 2.x sends `_meta: {}` on `initialize`, which is what makes this fixture unusable as its upstream server.

## What this changes

Two markers identify the request-meta wire, and neither is "there is a `_meta` key":

- a `MCP-Protocol-Version` header naming a post-session revision, and
- the per-request metadata envelope, whose own marker is `io.modelcontextprotocol/protocolVersion` inside `_meta`.

Classification uses those. Session-era traffic keeps the session path whether or not it carries `_meta`.

## What did NOT change

The rejections the envelope exists to trigger still fire — a request-meta-era request that omits the header still answers `-32020`, and an envelope missing its protocol version still answers `-32602`:

```
server/discover + full envelope, no header               -> 400  -32020
server/discover + header, envelope without protocolVersion -> 400  -32602
server/discover + header + full envelope                 -> 200
```

Nothing in the suite depended on the old rule: every scenario probe builds its request through `buildStandardHeaders`, which always sets `MCP-Protocol-Version` (default `DRAFT_PROTOCOL_VERSION`), so the `_meta`-integrity probes in `stateless.ts` reach the modern path by header and keep failing as they should.

## Verification

- `all-scenarios.test.ts`: 55/55 before and after.
- `npm test`: 627/627. `npm run check`: clean.
- New vitest case (`everything-server-request-classification.test.ts`) pins both halves — the two session-era requests that were rejected, and the three request-meta-era outcomes that must not move.
- Mutation-checked: reverting to presence-of-`_meta` fails 2 of the 5; dropping the envelope signal fails the `-32020` case; treating a session-era header as modern fails the progress-token case.


po-et:
Closing in favour of #507 — @lucarlig opened it three days before this, as the reporter, and it fixes the same thing. It routes a session-era `initialize` to the session path by its `params.protocolVersion`, which also covers the -32602 variant (session-era header plus `_meta`) I added here. I missed #507 when I checked the issue; sorry for the noise.

